PR #29, opened 2021-05-06
Add some other people

dngray:


jonaharagon (dismissed):
Websites need to be accurate (or simply removed)

dngray:
> Websites need to be accurate (or simply removed)

oops that was my mistake.

freddy-m:
@privacyguides/team we ought to get on with this, update it with photos, sites and all.

dngray:
> Websites need to be accurate (or simply removed)

This is true. I need to point mine to polarbear.army and finish setting that domain up.

Issue #55, opened 2021-09-13
Recommended browser extensions for android no longer available 


JeeVee11:
## Description
The two recommended browser extensions for android, ETag Stoppa and Cookie AutoDelete. Show  as no longer available for Firefox android on the add-ons page, and they can no longer be installed. 

Etag Stoppa was also last updated in December 2018, so it’s probably abandoned.

**URL of affected page:**
https://www.privacyguides.org/browsers/

PR #60, opened 2021-09-13
Remove Etag Stoppa

freddy-m:
## Description

Resolves: #55

### DIFF
--- a/_includes/legacy/sections/browser-addons.html
+++ b/_includes/legacy/sections/browser-addons.html
@@ -134,16 +134,6 @@
 </div>
 
 {% include legacy/cardv2.html
-  title="ETag Stoppa"
-  image="/assets/img/legacy_svg/3rd-party/etag_stoppa.svg"
-  description="<strong>ETag Stoppa</strong> Prevents Firefox from storing entity tags by removing ETag response headers unconditionally and without exceptions."
-  website="https://addons.mozilla.org/firefox/addon/etag-stoppa/"
-  forum=""
-  github="https://github.com/claustromaniac/etag-stoppa"
-  firefox="https://addons.mozilla.org/firefox/addon/etag-stoppa/"
-%}
-
-{% include legacy/cardv2.html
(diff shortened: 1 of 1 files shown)


jonaharagon:
The mentioned issue says neither are available on Android, can we double-check that? Maybe we should be removing the Android add-ons section entirely.

freddy-m:
> The mentioned issue says neither are available on Android, can we (or did you) double-check that? Maybe we should be removing the Android add-ons section entirely.

Yeah, I've always thought it a bit odd. Can't see any references to Android on their sites, but didin't thoroughly check. Either way, Etag can definitely be removed. And Cookie Auto Delete hasn't been updated since January... 

I'm in favour of removing the secion altogether @privacyguides/team

dngray:
> I'm in favour of removing the secion altogether @privacyguides/team

Agreed. I would like to bring forward https://github.com/privacytools/privacytools.io/pull/2081 and https://github.com/privacytools/privacytools.io/issues/1430 as this section has been neglected.

jonaharagon (changes requested):
Waiting for these changes https://github.com/privacyguides/privacyguides.org/pull/60#issuecomment-918902834

youdontneedtoknow22:
> I'm in favour of removing the secion altogether 

I would love keeping it there and only adding uBO to the list. The web without an adblocker is unusable these days.

NSABackdoors:
Sorry for bumping in but I checked that Etag is not available on Android.

dngray:
> Sorry for bumping in but I checked that Etag is not available on Android.

indeed, that's correct. It is my aim to encorporate https://github.com/privacytools/privacytools.io/pull/2081 into it.

We'll focus more on what settings to enable in the UI, that didn't exist when that page was written.

We'll be moving away from also specifying `about:config` settings, and referring people to arkenfox/user.js on desktop as you can't set those on android/googleplay version anyway.

The plan for that page, is to specify some extensions. Remove ones that aren't relevant/available. On the mobile space, really the only one that *is* relevant is ublock origin.

PR #79, opened 2021-09-16
Update android-addons to mention Work Profile apps

Victor239:
## Description

I noticed that Work Profile apps aren't on the website yet despite them being commonly used by people seeking greater control over their Android privacy. These apps are especially useful when using apps that require the Files permission to work, as it keeps them from accessing the rest of your personal files. Likewise with Contacts, Calendar, etc. No affiliation with any of these projects, have simply tried them all and they successfully do what they set out to.

### DIFF
--- a/_includes/legacy/sections/android-addons.html
+++ b/_includes/legacy/sections/android-addons.html
@@ -28,6 +28,6 @@
 
 <ul>
   <li><a href="/providers/dns#dns-android-clients">Our DNS client recommendations</a>, which have information on enabling encrypted DNS on Android.</li>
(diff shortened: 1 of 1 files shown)


jonaharagon:
Very good idea, thanks for the PR!

## Added Recommendations

This PR adds:

- **DueProcess** (Not discussed, [FOSS, GitHub](https://github.com/brnhrd/DueProcess)), described as "additionally includ[ing] plausibly deniable encryption to hide these apps on-demand, on a per-app basis or by triggering a so-called lockdown-mode." (as opposed to the other two options)
- **Shelter** (Not discussed, [FOSS, GitHub](https://github.com/PeterCxy/Shelter))
- **Island** (Not discussed, [FOSS, GitHub](https://github.com/oasisfeng/island))

I think these should probably be evaluated beforehand. I've heard of and used Shelter before myself, and would be happy to recommend it on the site.

dngray:
I'm not sure that Due Process is actually maintained anymore, last commit or any sign of activity was October 2020.

PR #97, opened 2021-09-18
Remove "encrypted" header for cloud services

weiw11:
## Description

Having "Encrypted" doesn't make any sense as E2EE is still unstable on Nextcloud, especially when it may cause data loss if the user did not backup their files and enabling it without knowing.


dngray:
This entire page is gone as https://github.com/privacyguides/privacyguides.org/commit/7394473f324d6cd9d9bf42d0a30f286a41ac530f#diff-70e8090343cd898aa0898cad5b3da1008b7037806f30ae33869b8f3890b90806

PR #133, opened 2021-09-22
Remove darkening of $warning in dark mode

elitejake:
## Description

Fixes unreadable low-contrast font color of `$warning` in dark mode.

### Before
image
image
<hr>

### After
image
image

### DIFF
--- a/_includes/legacy/cardv2.html
+++ b/_includes/legacy/cardv2.html
@@ -94,11 +94,11 @@
           <a
             href="{{include.tor}}"
             rel="noopener"
-            class="btn btn-tor mt-1 me-1"
+            class="btn btn-info mt-1 me-1"
             data-bs-toggle="tooltip"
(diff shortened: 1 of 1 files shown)


elitejake:
Also, in [`card.html`](https://github.com/privacyguides/privacyguides.org/blob/main/_includes/legacy/card.html) and [`panel.html`](https://github.com/privacyguides/privacyguides.org/blob/main/_includes/legacy/panel.html),
```html
{% if include.color == "warning" %}
<div class="card-header text-dark bg-{{include.color}}">
{% else %}
<div class="card-header text-white bg-{{include.color}}">
{% endif %}
```
doesn't make much sense because `text-dark` makes text much harder to read on `bg-warning`.
For example:
image
is harder to read than:
image

Any thoughts?

dngray:
I'm inclined to think the second one looks better.

PR #134, opened 2021-09-22
Fix Tor link buttons

elitejake:
## Description

In [`cardv2.html`](https://github.com/privacyguides/privacyguides.org/blob/main/_includes/legacy/cardv2.html), there is a class `btn-tor`, which doesn't exist in Bootstrap. It causes the Tor links to appear without a Bootstrap button. It looks like this:
image

I changed `btn-tor` to `btn-info`, which is fixed in [`recommendation-card.html`](https://github.com/privacyguides/privacyguides.org/blob/main/_includes/recommendation-card.html).

### DIFF
--- a/_includes/legacy/cardv2.html
+++ b/_includes/legacy/cardv2.html
@@ -94,11 +94,11 @@
           <a
             href="{{include.tor}}"
             rel="noopener"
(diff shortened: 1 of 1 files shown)


freddy-m (approved):
LGTM

PR #155, opened 2021-10-02
Add deprecation warning in HTTPS Everywhere

### LINKED ISSUE #145: Removal of HTTPS Everywhere
## Description

EFF has announced that HTTPS Everywhere browser extension will be soon depreciated by the end of 2021 as modern web browsers already have native support to forcefully encrypt `http` sites to `https`.

"With these simple settings available, EFF is preparing to deprecate the HTTPS Everywhere web extension as we look to new frontiers of secure protocols like SSL/TLS. After the end of this year, the extension will be in “maintenance mode.” for 2022. "
~ EFF
* Source : https://www.eff.org/deeplinks/2021/09/https-actually-everywhere  

This makes no sense in keeping the extension further enlisted in privacyguides.org .

## Suggestion 

Normal users generally tend not to update themselves on a regular basis. In this case referring a **soon to be depriciated** extension to such a consumer base is not paticularly a good option. It is best for a privacy guide to stay as updated as a browser so that its users at any point can get the correct information they need and not an older alternative solution to the problem.

PrivacyGuides should rather encourage users to enable the native settings (available on both firefox and chromium and all its derivatives) manually as they offer the same functionality.

**URL of affected page:**
https://privacyguides.org/browsers/

elitejake:
## Description

Added a warning label in HTTPS Everywhere. Hovering over it will bring up a tooltip stating that `EFF will deprecate this extension in 2022.`

I think this partially resolves #145.

### DIFF
--- a/_includes/legacy/sections/browser-addons.html
+++ b/_includes/legacy/sections/browser-addons.html
@@ -21,6 +21,7 @@
   title="HTTPS Everywhere: Secure Connections"
   image="/assets/img/legacy_svg/3rd-party/https_everywhere.svg"
   description="<strong>HTTPS Everywhere</strong> enables encryption of your connections to many major websites, making your browsing more secure. It is a collaboration between The Tor Project and the Electronic Frontier Foundation."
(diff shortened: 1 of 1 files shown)


Mikaela:
> Deprecation? How about we just completely remove it?

In my opinion it would be too early for that; https://github.com/privacyguides/privacyguides.org/issues/145#issuecomment-930937863.

PR #185, opened 2021-10-10
Include Proton Drive for Cloud Storage

splch:
## Description

Proposes adding Proton Drive as possible privacy-respecting cloud storage. Though the feature is still in beta, it already supports encrypted files as well as sharing and previewing in the browser.

All this edit does is add a small bullet-point mentioning ProtonMail's foray into encrypted file storage.


dngray:
Looks good, though instead of using `Beta |` I would use a badge. Ie:

```html
<a href="https://protonmail.com/blog/proton-drive-early-access/"
    class="text-decoration-none link-warning"
    data-bs-toggle="tooltip" title=""
    data-bs-original-title="">
    <i class="fad fa-question-circle">
    </i> Beta
</a>
```

dngray (approved):

dngray:
Hmm, I didn't mean to close this. I think it got closed when I deleted the branch, as I was trying to figure out why it wouldn't merge, forgetting we deleted the page, because we only had nexcloud on it.

PR #225, opened 2021-10-31
Add monero note (ivpn)

fredster33:
## Description
Updates the IVPN section with a note that they also accept Monero.

No COI, not currently/ever using IVPN.


Mikaela (approved):
https://www.ivpn.net/pricing/ confirms

PR #247, opened 2021-11-07
Invidio.us has been replaced by Invidious.io... a year ago

TheFrenchGhosty:
The invidio.us instance has been dead for more than a year now, we moved to a new domain almost 1 year ago, in January... yet you never updated the link.

You are still sending thousands of users to a domain we don't own (we just control it) and that have been kept alive - as a redirect - because a ton of people - including you - still link to it because they never update their stuff.

Please, can you at least use what you recommend (because this shows that you obviously don't) or at the very least CHECK that the link aren't dead and/or haven't moved?!

Thank you.


TheFrenchGhosty:
And CI fail because you link to 4 different 404 (and a dozen of 403). Nice.

Mikaela (approved):
Thank you for the pull request and I apologise on behalf of the team that we haven't handled this situation as well as we could have. There has been a lot of busyness recently.

I opened #250 in order to hopefully manage cases like this better.

TheFrenchGhosty:
>There has been a lot of busyness recently.

Sure, recently. The problem is that privacytools io has been managed this way for years and I'm not expecting the new branding/management to change that. There is SO much issues with this ptio/pgorg that, yes, I'm just tired of it.

dngray:
This one had slipped by me, thanks @Mikaela

PR #254, opened 2021-11-08
Remove hosting providers

tommytran732:
Related Discussion: https://github.com/privacyguides/privacyguides.org/discussions/207

The hosting provider is subpar right now, it is unclear who the intended audience for this is. I don't think the suggestions are good either. 

It is best to remove this for now. I have some idea on which providers to recommend for which purpose, and I will make a PR to readd this section with better quality content later.


tommytran732:
The ran before I made my second commit :( 
Would be great if we could run it again

dngray:
Agreed, we don't even eat our own dog-food on this. If this was to be re-added I think it would be necessary that there is a criteria.

freddy-m:
@dngray IIRC we had this in the pipeline for PTio ([#1196](https://github.com/privacytools/privacytools.io/issues/1196)), though generally I think every section should have it's own criteria as per [#24](https://github.com/privacyguides/privacyguides.org/discussions/24).

dngray:
Well looking at that previous suggestion:

> Prioritize Products by privacy respecting nationality.

This is something we already do in general across the site.

> Cannot be based in USA or UK.

Those aren't the only two bad countries in the world. We're progressively phasing out mention of "eyes" countries because it is not the only intelligence gathering agreement. See: https://github.com/privacytools/privacytools.io/issues/1437

> Must be acessable via free software (i.e OpenVPN, WireGuard)

I can't think of any that aren't? That would be up to the owner, not the hosting service.

> Use Encryption

Up to individual sysop to make certificates, so in other words the client. I don't know of any hosting providers that don't allow sftp/ssh.

> Accept Cryptocurrency

Not inherently private, that depends on what cryptocurrency. You're gonna lose a lot of services if you say they must all take privacy coins like XMR or Zcash for example.

> No logging policy

Hosting providers will do this to mitigate DDoS attacks etc. I don't think you'll find any "no logs" hosting providers out there.

dngray (approved):

PR #257, opened 2021-11-08
Revamping the VPN recommendations

tommytran732:
This is a follow up for my https://github.com/privacyguides/privacyguides.org/pull/255 pull request. I feel like the FAQ section is borderline misinformation since it neglects to emphasize the benefits that a VPN provides and acts as if encrypted DNS + https is a viable alternative. It downplays the dangers of IP based tracking. Before we even worry about browser fingerprinting and the like, we must take care of IP base tracking first. Dynamic IP addresses is not enough - you will still show up as being from the same area, and usually still in the same IP block. I have also updated the information on the VPN providers too.

Also - we do not need to mention a VPN killswitch as a feature. It is already listed as a requirement.


tommytran732:
I messed up 1 single tag. I will fix it now.

dngray:
As for the other removals. I agree with those.

tommytran732:
> As for the other removals. I agree with those.

Um, is there something you do not agree with? It sounds like there is something but your message wasn't sent.

dngray:
> Um, is there something you do not agree with? It sounds like there is something but your message wasn't sent.

No, I read through it and I think it is now better worded, more clear.

dngray (approved):

dngray:
LGTM

PR #259, opened 2021-11-08
Remove canvas blocker

tommytran732:
This exention is quite hard to use, and really cannnot be used effectively. There are not enough people using it, and not everyone using it will block the same stuff. All it does is that it will make the user stand out more.

The user should be using a fingerprinting resistant browser and not relying on an obscure third party extension. It does nothing for privacy and potentially worsen security, since there are more entities to trust and extensions can weaken web isolation.


dngray:
Going to approve this one as it's really only useful if you've got RFP disabled for a site.

For those situations I just use a different browser profile.

TheFrenchGhosty:
@tommytran732 @dngray 

>This exention is quite hard to use

How is that even an argument? Also... Canvas Blocker when setup literally doesn't have to be touched.

>There are not enough people using it, and not everyone using it will block the same stuff. All it does is that it will make the user stand out more.

Anonymity and Privacy aren't the same thing, this extension makes user more private.

>The user should be using a fingerprinting resistant browser and not relying on an obscure third party extension

Ah, yes, the thing that is so bad, and so black and white that half of the websites will break when it's enabled.

>It does nothing for privacy and potentially worsen security, since there are more entities to trust and extensions can weaken web isolation.

It literally blocks canvas fingerprinting, with ENHANCE privacy... and the extension is FOSS... so you don't have to trust anyone.

tommytran732:
> @tommytran732 @dngray
> 
> > This exention is quite hard to use
> 
> How is that even an argument? Also... Canvas Blocker when setup literally doesn't have to be touched.
> 
> > There are not enough people using it, and not everyone using it will block the same stuff. All it does is that it will make the user stand out more.
> 
> Anonymity and Privacy aren't the same thing, this extension makes user more private.
> 
> > The user should be using a fingerprinting resistant browser and not relying on an obscure third party extension
> 
> Ah, yes, the thing that is so bad, and so black and white that half of the websites will break when it's enabled.
> 
> > It does nothing for privacy and potentially worsen security, since there are more entities to trust and extensions can weaken web isolation.
> 
> It literally blocks canvas fingerprinting, which ENHANCE privacy... and the extension is FOSS... so you don't have to trust anyone.

Have you ever heard of the magical setting called RFP?

dngray:
> Have you ever heard of the magical setting called RFP?

Is that [that thing](https://support.mozilla.org/en-US/kb/firefox-protection-against-fingerprinting) made by Mozilla?

RFP is not mentioned, because it only has [very few](https://github.com/arkenfox/user.js/wiki/4.1-Extensions) usecases.

We recommend everyone uses RFP, as this is vetted by Mozilla, Tor Project and experts who understand how fingerprinting actually works. As it's a part of the browser it has certain privilege that extensions will never have. Extensions can only do what the APIs allow.

PR #260, opened 2021-11-08
Remove Decentraleyes

tommytran732:
This extension does nothing to improve the user experience, and is making the user more identifiable by not loading contents from the CDNs. It adds another party to trust, and could potentially weakens site isolation. Moreover, there is no reason to assume CDNs are malicious and then take the enumeration of badness approach and load content locally. It doesn't work. It's privacy theater.


Mikaela (approved):
👍🏻  this was additionally a part of https://github.com/privacytools/privacytools.io/pull/2081 by @dngray

dngray (approved):

PR #261, opened 2021-11-08
Remove additional functionality

tommytran732:
Each browser a user add is another party to trust and another way site isolation is weakened. Thus, the recommendations for extensions should be minimal, if there are any recommendations to be made at all.

These 2 extensions are privacy frills. They are not very useful in the first place, and all what the users really get from them is additional attack surface and more fingerprintability. 

We should not be recommending useless extensions at all.


Mikaela (commented):
:-1: I don't think users are suddenly going to start reading privacy policies very carefully, so I think TosDR is important to mention.

Additionally Tor network constantly needs help and growing, so I wouldn't go removing such an easy way to help the project.

tommytran732:
I think it's privacy frills. We should be look for technical guarantees, not privacy policies. If a site says it does not store your IP for, say, over 7 days, do you believe it? I'd say no, use a VPN.

Likewise, if a site says it does not sell your personal data, would you still put your real name and personal information on it? No. You should be using a fake identity.

Having an extension  with access to your browsing data means that you have to have complete trust in the extension developer. And even if you do trust them, extensions increase the attack surface of a browser and weaken site isolation. You are sacrifice a lot for very little gain (if any at all) when you install something like this. It's nothing substantive. 

As for the Snowflake extension, there are other ways to help the Tor project. You can donate actual money, you can get a VM and run a relay node for a few dollars a month. You shouldn't be weakening your security unnecessarily only to feel a bit better that you help out the Network.

Mikaela:
We are a site for everyone interested in privacy, not only for advanced users. Not everyone is going to use fake details especially if they aren't told to do that anywhere. That also results to additional concern such as how to apply for a job when staying anonymous, how to connect to real people around you when using a wrong name and ensure they won't leak your real name which again goes out of scope of this issue.

Additionally I don't want to enforce mindset that makes people wonder whether they should get a VPN, food or pay rent, not everyone has money to be giving at nice things like Tor

tommytran732:
Even so, this is worsening security for people who are interested in privacy. We should be encouraging good opsec, keeping things simple and stupid. The things we recommend must have substantial benefits and no to very little drawbacks for the user's privacy and security. Is it really worth trading site isolation and fingerprinting resistence for a privacy policy summary? It's really, really not worth it.

A lot of VPNs and Tor are free. If the user doesn't have the money, it is okay. They don't have to pay for anything. They don't have to contribute to anything. They should not feel the obligation to worsen their security for the feeling of being part of something bigger. They should take care of themselves first.

1hiking:
I don't see the point of Snowflake considering they make up almost an insignificant percentage of the [Tor network](https://metrics.torproject.org/userstats-bridge-transport.html?start=2021-01-01&end=2021-11-10&transport=%21%3COR%3E&transport=obfs4&transport=meek&transport=scramblesuit&transport=snowflake&transport=%3C%3F%3F%3E&transport=%3COR%3E).

Just my 2 cents

dngray:
> Each browser a user add is another party to trust and another way site isolation is weakened. Thus, the recommendations for extensions should be minimal, if there are any recommendations to be made at all.

While that is the the case, I am inclined to believe that TosDr does provide some merit for people interested in usage policies of services they use.

While technical methods to increase privacy are obviously something that we are interested in, I think TosDr helps users navigate legal issues too, such as what a service can actually do.

When using these services it's not always possible to find technical means to safeguard your privacy either, simply contributing to the service or having an account with, will mean that the service can track/use the information you provide within their policy. I think it is important that some services have better privacy policies than others, and therefore I think this extension should stay.

> I don't see the point of Snowflake considering they make up almost an insignificant percentage of the [Tor network](https://metrics.torproject.org/userstats-bridge-transport.html?start=2021-01-01&end=2021-11-10&transport=%21%3COR%3E&transport=obfs4&transport=meek&transport=scramblesuit&transport=snowflake&transport=%3C%3F%3F%3E&transport=%3COR%3E).

In regard to Snowflake, this is an extension maintained by the Tor Project. I also think it's a rather neat way that people can *help* increase privacy for others, without making financial donations to various projects or contributing code, or time etc.

Rather than having a card for it, we might like to write that it is a way to help others. or something like that.

dngray:
> Each browser a user add is another party to trust and another way site isolation is weakened. Thus, the recommendations for extensions should be minimal, if there are any recommendations to be made at all.

While that is the the case, I am inclined to believe that TosDr does provide some merit for people interested in usage policies of services they use.

While technical methods to increase privacy are obviously something that we are interested in, I think TosDr helps users navigate legal issues too, such as what a service can actually do.

When using these services it's not always possible to find technical means to safeguard your privacy either, simply contributing to the service or having an account with, will mean that the service can track/use the information you provide within their policy. I think it is important that some services have better privacy policies than others, and therefore I think this extension should stay.

> I don't see the point of Snowflake considering they make up almost an insignificant percentage of the [Tor network](https://metrics.torproject.org/userstats-bridge-transport.html?start=2021-01-01&end=2021-11-10&transport=%21%3COR%3E&transport=obfs4&transport=meek&transport=scramblesuit&transport=snowflake&transport=%3C%3F%3F%3E&transport=%3COR%3E).

In regard to Snowflake, this is an extension maintained by the Tor Project. I also think it's a rather neat way that people can *help* increase privacy for others, without making financial donations to various projects or contributing code, or time etc.

Rather than having a card for it, we might like to write that it is a way to help others. or something like that.

tommytran732:
Closing in favor of https://github.com/privacyguides/privacyguides.org/pull/309

PR #262, opened 2021-11-08
Remove https everywhere

tommytran732:
Both Chromium and Firefox now has https everywhere built in, and this extension is no longer necessary. In fact, that is why it is being retired. We are approaching 2022, and there is no longer any reason to keep recommending it. The users should use the built in feature of their browser instead of a third party extension.


Mikaela (commented):
👎🏻  Enabling HTTPS Everywhere mode is difficult especially in Chromium and mobile browsers, so I disagree with removing it for now, while I agree that it will have to be done at a later point.

Please see also #145

tommytran732:
I'd rather remove it first, clean the site up from useless suggestions, then we can work on detailed guides on hoow to do this.

Mikaela:
I would prefer less technical users to install an extension that is going away naturally in a couple of years than them to be at risk of visiting HTTP sites due to the guide on enabling HTTPS only mode not existing and due to HTTPS Everywhere syncing unlike the HTTPS only mode (which is the case with at least Edge and Vivaldi).

In addition to #145, please see https://github.com/EFForg/https-everywhere/discussions/20138#discussioncomment-1439427.

dngray:
> 👎🏻 Enabling HTTPS Everywhere mode is difficult especially in Chromium and mobile browsers, so I disagree with removing it for now, while I agree that it will have to be done at a later point.

Mobile chromium browsers don't support extensions anyway, so not sure that this impacts that.

tommytran732:
> > 👎🏻 Enabling HTTPS Everywhere mode is difficult especially in Chromium and mobile browsers, so I disagree with removing it for now, while I agree that it will have to be done at a later point.
> 
> Mobile chromium browsers don't support extensions anyway, so not sure that this impacts that.
> 
> > The extension won't be taken down until late next year (winter, likely). Before then we want to utilize existing mechanisms to remove the extension from Firefox.
> 
> We could put a hold on this perhaps until then. Alternatively we could provide guidance on enabling HTTPS-only.

I think we should provide guidance on enabling it. I am just making PRs to clean up right now, after every unnecessary stuff is removed I will make PRs providing guidance.

Mikaela:
I will not give this PR my approval until the guidance exists as I don't approve leaving users without using https anywhere during the time between potential merging of this PR and merging of the guidance.

Additionally I am still concerned by the option not syncing between browser instances when logged in, at least on Vivaldi and Edge.

tommytran732:
The user shouldn't be using any of those 2 browsers in the first place. Proprietary + no fingerprinting resistance whatsoever 

If anything, Brave should be in the recommended list, and and Vivaldi should be anti recommendation

tommytran732:
In fact, it is quite literally impossible to give any  good recommendations if we recommends a ton of stuff and try to accommodate every obscure use cases out there.

dngray:
> > > 👎🏻 Enabling HTTPS Everywhere mode is difficult especially in Chromium and mobile browsers, so I disagree with removing it for now, while I agree that it will have to be done at a later point.
> > 
> > 
> > Mobile chromium browsers don't support extensions anyway, so not sure that this impacts that.
> > > The extension won't be taken down until late next year (winter, likely). Before then we want to utilize existing mechanisms to remove the extension from Firefox.
> > 
> > 
> > We could put a hold on this perhaps until then. Alternatively we could provide guidance on enabling HTTPS-only.
> 
> I think we should provide guidance on enabling it. I am just making PRs to clean up right now, after every unnecessary stuff is removed I will make PRs providing guidance.

I'll approve this on the basis that we provide guidance on enabling HTTPS-only mode.

PR #267, opened 2021-11-10
Remove duplicated cloud section

tommytran732:
This remove the duplicated cloud section and fixes a dead hyperlink


dngray:
Approved. This page only made more sense when we had more recommendations.

PR #274, opened 2021-11-10
Video frontend Revamp

tommytran732:
The video page is extremely wrong and broken right now:
- Being decentralized doesn't mean it's private, it's more for censorship resistance
- Your IP is still visible to the individous instance, and its not like they will even proxy you by default
- Broken link

This page is so much of a mess that it needs to be completely redone, and I don't think keeping it is a good idea considering it's a legacy page. We will make a new page recommending things like Newpipe/Freetube later using the new system.


dngray:
I do agree with re-working the page and recommending things like Newpipe/Freetube.

tommytran732:
Oh dear, I need to fix the PR

tommytran732:
@dngray 

This PR is ready for merge. It also fixes https://github.com/privacyguides/privacyguides.org/issues/334

It also fixes https://github.com/privacyguides/privacyguides.org/issues/334

PeerTube is not mentioned as I cannot find how you can have a private subscription list with it. It just looks like Youtube, but Federated. You still need to sign up for an account and what not. 

LBRY is a much better alternative as it stores the subscription list on your local LBRY wallet (as weird as it sounds, the subscription list part inside of the wallet is not on the blockchain - only the LBRY channel, videos, and funds are).

dngray (approved):

PinkDev1:
> This page is so much of a mess that it needs to be completely redone, and I don't think keeping it is a good idea considering it's a legacy page. *We will make a new page recommending things like Newpipe/Freetube later using the new system.*

This addition should've been on this same PR before it was merged into main. Pushing incomplete contributions into main, defeats the purpose of stable-branches. 

[Some people were confused by this](https://www.reddit.com/r/PrivacyGuides/comments/qy7ozx/recent_updates_to_privacyguidesorg/hlemee6/), so please try to avoid it in the future

dngray:
> [Some people were confused by this](https://www.reddit.com/r/PrivacyGuides/comments/qy7ozx/recent_updates_to_privacyguidesorg/hlemee6/), so please try to avoid it in the future

Regarding this post specifically. I'll post my [reply](https://github.com/privacyguides/privacyguides.org/pull/392#issuecomment-981073251).

PR #280, opened 2021-11-10
Fix some broken links

dngray:


dngray:
Does seem to be some false errors there regarding support pages that use zendesk. Maybe they block the checks somehow.

PR #283, opened 2021-11-10
Added a warning to Fruux: proprietary application

TheNoobWar:
## Description

Resolves: Added a warning badge to Fruux under the Worth Mentioning section as it is a proprietary application.

### DIFF
--- a/_includes/legacy/sections/calendar-contacts-sync.html
+++ b/_includes/legacy/sections/calendar-contacts-sync.html
@@ -58,7 +58,7 @@
     {% include badge.html
     color="warning"
     icon="fas fa-exclamation-triangle"
-    tooltip="The source programming code for Fruux is not available. Fruux is a proprietary application."
+    tooltip="The source code for Fruux is not available. Fruux is a proprietary application."
     %}
   </li>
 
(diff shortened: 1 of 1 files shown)


dngray:
I'm actually wondering if this should even be on PrivacyGuides at all.

From what I can tell it was always there and never open source. It is not encrypted at rest either.

_includes/legacy/sections/calendar-contacts-sync.html
```diff
@@ -55,6 +55,11 @@ <h3>Worth Mentioning</h3>
 <ul>
   <li>
     <a href="https://fruux.com/">fruux</a> - A unified contacts/calendaring system that works across platforms and devices.
+    {% include badge.html
+    color="warning"
+    icon="fas fa-exclamation-triangle"
+    tooltip="The source programming code for Fruux is not available. Fruux is a proprietary application."
```
Mikaela, line 61:
```suggestion
    tooltip="The source code for Fruux is not available. Fruux is a proprietary application."
```

The word "programming" in the middle here looks a bit strange for me

Mikaela (changes requested):

Mikaela (approved):
Approving for now unless a delisting PR happens.

freddy-m:
> I'm actually wondering if this should even be on PrivacyGuides at all.

@dngray I don't think it should be either.

Mikaela (approved):
I think it should possibly be in a different PR that says `Resolves: #283` 🤷🏻

freddy-m:
> I think it should possibly be in a different PR that says Resolves: #283 🤷🏻

Nevermind

PR #284, opened 2021-11-10
Moving Cloud storage

tommytran732:
Ah, it turned out to be quite a big PR since there are so many interconnected things that I wanna fix all at once
1. Moving cloud storage from software to providers
2. Removing Cryptpad from the cloud page (it's a productivity tool, not cloud storage, it can't hold anything other than pads)
3. Deleting Paste Services and move PrivateBin to Productivity tools
4. Added caveats to CryptPad, PrivateBin, and Etherpad. I didn't use the small read warning because I think eventually all of this should be written directly into the description to make it more clear to the user.
5. Removed ethercalc.net and disroot mention because etherpad is not private at all unless you self-host.
6. Added ProtonDrive
7. Remove the obnoxious alert banner and provide actual explanation


freddy-m:
Do we need to list PasteBins @privacyguides/team? They're not really a 'privacy tool'.

dngray:
> Do we need to list PasteBins @privacyguides/team? They're not really a 'privacy tool'.

My personal opinion is they are not. Even the ones which are "encrypted", you can't tell if the owner has modified them to not work in some minor way. It's part of the reason I still clbin.com and termbin.com, at least they don't require javascript.

dngray (approved):
LGTM, this one is ready for merge.

freddy-m (approved):
LGTM

Issue #290, opened 2021-11-11
Add explanation for verified boot


tommytran732:
This issue is a note to make proper explanation on why verified boot is important later on

zecamigo:
Aren't you guys confusing privacy with security?
I mean AFAIK, verified boot will protect against evil maid attacks. This is mostly a security issue, and a bit distant from typical privacy concerns along that spectrum. It has little to do with protecting oneself from surveillance, unless you're targeted. Maybe you should take into consideration privacy needs of different levels and also the fact that the remaining ROM's support a very limited range of devices. (cross commenting from related merge request).

dngray:
We're looking at a criteria for the for the mobile OS.

I think the general consensus is there is too much uncertainty over *random rom* for *random mobile device*.

Providing a criteria and requiring verified boot limits our recommendations to devices we know at least are able to meet a minimum security baseline.

PhysicsIsAwesome:
> Aren't you guys confusing privacy with security?

You cannot really separate the two. For privacy you need to secure stuff you want to keep secret from access through other parties. And by getting hacked you could lose all digital privacy. An example is, that a lot of people root their smartphone to get additional privacy stuff working, that needs deep integration into the system, but aren't aware, how much vulnerable this makes the device. To be able to make good decisions you need to know the pros and cons of the security and privacy aspects.

zecamigo:
>  An example is, that a lot of people root their smartphone to get additional privacy stuff working, that needs deep integration into the system, but aren't aware, how much vulnerable this makes the device.

Spot on. LineageOS stopped shipping SU from v.17. Root access via ADB is very limited and they're continuously hardening the OS security wise. Still don't understand the threat model leading to this decision,.

jonaharagon:
I think this is covered by #371 now, but add a note to that discussion if you feel it isn't.

zecamigo:
What's the advantage of a verified boot again? https://www.theverge.com/2021/12/4/22817758/broken-google-pixel-phone-privacy-leak

SkewedZeppelin:
Verified boot has nothing to do with the user not having a (strong) pin/password
set.

PR #294, opened 2021-11-11
Remove LineageOS

freddy-m:
## Description

Removed LineageOS - Calyx and Graphene are better alternatives.

### DIFF
--- a/_includes/legacy/sections/android-operating-systems.html
+++ b/_includes/legacy/sections/android-operating-systems.html
@@ -25,16 +25,6 @@
   github="https://github.com/GrapheneOS/"
 %}
 
-{% include legacy/cardv2.html
-  title="LineageOS (Official Builds)"
-  image="/assets/img/legacy_svg/3rd-party/lineageos.svg"
-  description='LineageOS is a custom ROM based on the Android Open Source Project with <a href="https://wiki.lineageos.org/devices/">official support for a wide variety of devices</a>. The project is the successor to the popular CyanogenMod project, focusing on system control and customization, device longevity, and privacy. For security reasons, we do not recommend third-party builds of LineageOS for unsupported devices.'
-  badges="info:AOSP"
-  website="https://www.lineageos.org/"
-  privacy-policy="https://www.lineageos.org/legal/"
-  github="https://github.com/LineageOS"
-%}
-
 <h4>Worth Mentioning</h4>
 
 <ul>


dngray:
This is good to do in preparation for https://github.com/privacyguides/privacyguides.org/issues/290 as we'll be requiring verified boot and encryption at rest from any new roms/devices

dngray (approved):

zecamigo:
Aren't you guys confusing privacy with security?

julianfairfax:
Exactly, this leaves all non-Pixel users thinking they have no options, which is most users.

pm4rcin:
I think it could be added as worth mentioning maybe with a warning or asterisk about "verified boot". In my opinion this should satisfy everyone. And also mention to use "Official" LineageOS Builds or trust the maintainer with "Unofficial".

freddy-m:
> Aren't you guys confusing privacy with security?

Not really. While we are a site focused on privacy, we wouldn't want to recommend something that was insecure when there are alternatives available. Hence, we recommend Calyx and Graphene which are both private and (especially in the case of Graphene) secure.

zecamigo:
> While we are a site focused on privacy, we wouldn't want to recommend something that was insecure when there are alternatives available.

I understand and agree both are interdependent. However, what is the threat model you are working with here? You would have a real high security profile to have to defend against an evil maid attack on your device. 

If you want to profile guidance for high risk investigative journalists, maybe you'll have to remove about half of the services you recommend, because for example you know now that many mail providers will give out user's IP's to the government.

Or maybe I'm missing something about the risks of not having a verified boot?

zecamigo:
Another example, you still recommend Signal even though it is known that they're hosted on PRISM-connected companies and that the secure enclaves have vulnerabilities for that exact scenario, server was closed sourced for over a year, and other red flags. However no word on XMPP. So you're hosted on MicroSoft server, recommended devices made by Google and other services with known issues, however your reluctant to recommend a community powered alternative because of a high security (basically affects you if you're targetted) issue.

pm4rcin:
> Another example, you still recommend Signal even though it is known that they're hosted on PRISM-connected companies and that the secure enclaves have vulnerabilities for that exact scenario, server was closed sourced for over a year, and other red flags.

Yes it's true that Signal is hosted there but it doesn't matter since the main idea of Signal is that the server could be compromised and even then your messages are still secure. You can even use Signal with Tor easily like Snowden does. There were some more red flags with Signal like the crypto but there are FLOSS forks which removes binary blobs and could remove the crypto thing if it'll come to an app.

Regarding the rest I think I agree. If your device gets stolen you should probably reinstall stock ROM or contact a service or something to securely extract the data you want and reset phone to factory defaults. But take this with a big grain of salt since I'm not an expert on phones security.

TheFrenchGhosty:
@freddy-m @dngray 

>Removed LineageOS - Calyx and Graphene are better alternatives.

Calyx and Graphene are available, on at most, 5 devices, LineageOS is available dozens of devices. Calyx and Graphene are better than LineageOS, sure, but they are in no way an alternative to it. LineageOS, when degoogled, is perfectly fine privacy wise.

tommytran732:
> @freddy-m @dngray
> 
> > Removed LineageOS - Calyx and Graphene are better alternatives.
> 
> Calyx and Graphene are available, on at most, 5 devices, LineageOS is available dozens of devices. Calyx and Graphene are better than LineageOS, sure, but they are in no way an alternative to it. LineageOS, when degoogled, is perfectly fine privacy wise.

DivestOS exists. Did you even read the site?

dngray:
We recommend optimal solutions. You should always buy hardware to meet software needs and not the other way around. It's worth noting not all lineageos builds are maintained to the same level of quality, and they don't really provide any "privacy" enhancing features when compared to some other random rom that doesn't include gapps.

The site is moving towards creating criteria for most/all sections. We're looking at quality over quantity approach which slowly tightens as products improve.

TheFrenchGhosty:
>You should always buy hardware to meet software needs and not the other way around.

That's now how the world works.

dngray:
> That's now how the world works.

Says who? Please stop spamming closed PRs thanks. I don't care what your agenda is.

> Then stop recommending 3/4 awful stuff.

We're slowly refining the pages as we create criteria/research for each one.

PR #295, opened 2021-11-11
Fix I2P buttons

elitejake:
## Description

Fix I2P buttons.


elitejake:
What should be the color of the button? Does I2P have a brand color?

dngray:
> What should be the color of the button? Does I2P have a brand color?

No it's supposed to be purple like the one above as it is Tor

dngray:
This one is an error, it shouldn't have an onion logo either.

dngray (changes requested):
We probably need a custom icon for this with like ⠶⠶ or something to represent the logo

elitejake:
They have a bunch of logos here: https://github.com/i2p/i2p.android.base/tree/master/art
@dngray

elitejake:
Here's the updated font file I created. How does it look?
[ptio-font.zip](https://github.com/privacyguides/privacyguides.org/files/7531881/ptio-font.zip)

elitejake:
Doesn't look good but that is their logo.

dngray:
> Here's the updated font file I created. How does it look?

It looks good I think. I like it better than the ⠶⠶ idea I had.

Issue #288, opened 2021-11-11
Update OpenWRT logo


1hiking:
## Description

The logo of OpenWRT has changed
Suggested logo to be replaced: https://en.wikipedia.org/wiki/OpenWrt#/media/File:OpenWrt_Logo.svg

**URL of affected page:**
https://privacyguides.org/operating-systems/#firmware

PR #302, opened 2021-11-12
Update OpenWrt logo

elitejake:
## Description

Resolves: #288


dngray:
We only have images that are 128x128 or 384x128.

In this case I think 128x128 makes more sense. Looking at their [usage guidelines](https://openwrt.org/_media/docs/guide-graphic-designer/openwrt-logo-usage-guidelines.pdf) I think this looks a little better.

I think for the dark version, the inverse would make sense

PR #305, opened 2021-11-13
Remove Turtl

freddy-m:
## Description

Removed Tutrl as they haven't updated their source code in over 8 months.


dngray:
Looking at the lead developers https://github.com/orgs/turtl/people doesn't look like there is much activity in any public repositories.

dngray (approved):

Issue #273, opened 2021-11-10
Provide guidance on enabling HTTPS-Only mode


dngray:
## Description

Removal of https://github.com/privacyguides/privacyguides.org/pull/262 I decided we should provide instruction on how to enable HTTPS-Only mode in both Chromium/Firefox.

BraINstinct0:
Leaving quick note on Chrome.
- Chrome
- * Starting from at Chrome 94, `chrome://settings/security` ->`Advanced` section->toggle `Always use secure connections` on. Unsure if it is on by default, or if it is since which version.
- * If not shown(i.e. on older versions), worth checking if `chrome://flags/#https-only-mode-setting` is set to `Disabled`. If so, switch it to `Enabled` or `Default` (which is equivalent). Not sure which version this flag was added.
- * Works on Windows, Linux, Chrome OS and Android.

(due to lack of patience) didn't write further; a quick search brought up a Link: **h**ttps://beebom.com/how-enable-https-only-mode-chrome-firefox-edge-safari/ .

Issue #245, opened 2021-11-05
Browser Fingerprinting: Influences


PhysicsIsAwesome:
**https://privacyguides.org/browsers/**

Please clear up a few sentences regarding browser fingerprinting, that could be easily missunderstood.

> You need your browser to look as common as everyone else.

This could be missunderstood as: Just use a browser with standard settings and you're fine. But there is more to it and I think this should be made clear:

- Not only your browser (meaning type, version, extensions, settings), but also parts of your 
- hardware, 
- OS,
- and libraries (like fonts, maths, drivers etc) involved. 

This is what makes browser fingerprinting such a difficult problem on desktop computers and by far not that much on top-selling smartphones or tablets. Worst case: Building your own highly uncommon desktop computer, with a 8k Monitor and your super special Arch Linux build, installing fonts libraries and a customized browser in a seldom way.

> Disabling JavaScript, using Linux, or even using the Tor Browser Bundle, will make your browser stick out from the masses.

- Yes it will let your browser stick out from the masses, but that is not the point to make. It is all about state-of-art detectable features and their real world entropy.
- It could be simply missunderstood, like: Don't use Tor browser. 
- And yes, disabling JavaScript will let you stick out, but it will also block a lot of fingerprinting vectors and thus might outweigh the sticking out.
- Agree on the Linux part, though.

dngray:
We're looking at stripping this section shortly and re-writing it.

Issue #243, opened 2021-11-05
Spoofing user-agent strings


PhysicsIsAwesome:
Hi,

please change this sentence on https://privacyguides.org/browsers/:

> You should have a spoofed user-agent string to match what the large userbase has.

Spoofing a user-agent gets easily detected by fingerprinting scripts and will let you stand out extremely. FF's privacy.resistFingerprinting spoofs the user-agent, but in this particular case it doesn't hurt, since all RFP users do it in the same way and using RFP can get detected anyway. In all other cases spoofing a user-agent string is a big no. It would be best to not only delete this sentence, but also make clear, that you strongly advice against it.

Best regards,
PhysicsIsAwesome

Issue #69, opened 2021-09-14
remove privacy.firstparty.isolate


tauin:
## Description
`privacy.firstparty.isolate` has been made (effectively) redundant with recent firefox updates, and should be removed
**URL of affected page:**
https://www.privacyguides.org/browsers/

rusty-snake:
https://github.com/privacytools/privacytools.io/pull/2081#issuecomment-784441119 and the following

dngray:
> [privacytools/privacytools.io#2081 (comment)](https://github.com/privacytools/privacytools.io/pull/2081#issuecomment-784441119) and the following

What I am looking at doing with that is removing the whole "about:config" section, and just showing what can be achieved through Mozilla's UI.

We'll of course still mention https://github.com/arkenfox/user.js

NSABackdoors:
Any thoughts on how you would expect this to happen?

What does the Strict mode in Firefox leave to be desired? Curious here.

PhysicsIsAwesome:
> What does the Strict mode in Firefox leave to be desired? Curious here.
 A lot of things. privacy.resistFingerprinting is not enabled, just to give an example.

PhysicsIsAwesome:
> `privacy.firstparty.isolate` has been made (effectively) redundant with recent firefox updates, and should be removed

I don't see it being redundant. Do you have a source or sth? [Tor browser](https://gitlab.torproject.org/tpo/applications/tor-browser/-/blob/7ed0992f877c447f0605a0e430eedd69aa1fe64d/browser/app/profile/000-tor-browser.js) and [arkenfox](https://github.com/arkenfox/user.js/issues/1051#issuecomment-940903984) still use FPI.

rusty-snake:
If you set ETP=strict, FPI is effectively redundant (for the average user).

Thorin-Oakenpants:
FPI overrides ETP Strict - so you should mention to make sure that is set to false - since you listed it for so long to so many users

dngray:
> FPI overrides ETP Strict - so you should mention to make sure that is set to false - since you listed it for so long to so many users

Would it not be worthwhile to just suggest users create a new profile?

If I'm being totally honest, we don't know what state a user's browser might be in if they've been fiddling with `about:config` settings. I'm not totally confident that flicking that single option will restore things.

Perhaps something general like.

> If you've fiddled with about:config settings in the past, we suggest you create a new browser profile.

Thorin-Oakenpants:
[...]

PR #309, opened 2021-11-14
Browser Revamp

tommytran732:
Discussion: https://github.com/privacyguides/privacyguides.org/issues/298#issuecomment-968229963

Closes: https://github.com/privacyguides/privacyguides.org/issues/273
Closes: https://github.com/privacyguides/privacyguides.org/issues/69
Closes: https://github.com/privacyguides/privacyguides.org/issues/243
Closes: https://github.com/privacyguides/privacyguides.org/issues/245
Closes: https://github.com/privacyguides/privacyguides.org/discussions/90


SkewedZeppelin:
Perhaps link the Android versions of Brave and Tor Browser?
Also this leaves out a Gecko based Android browser.

I also don't really like that Librewolf isn't source-built in any distros.
And that isn't even an option for Brave.
But that belongs in another issue.

tommytran732:
Er yeah I am basically done with my part, if anyone has anything else to add, feel free to make suggestions

SkewedZeppelin:
uBlock Origin recommendation was removed?
The "enumerating badness" argument is defeatist imo.
Even if you can't block all trackers or what not, you still drastically minimize execution of untrusted content/connections.

Librewolf is still mentioned in Firefox section.

dngray:
> uBlock Origin recommendation was removed?

It's not, things are being moved around.

Thorin-Oakenpants:
> And the Tor Browser excuse for not adding uBlock Origin boils down to "we don't want people thinking Tor is just Firefox with an adblocker". Even though adding uBlock Origin to Tor Browser would directly increase security, and reduce load on the network. I mean the TBB devs went out of their way to disable the Firefox default tracking protection list. Asinine.

That's not why TB doesn't have any content/tracker blocking
- they already have FPI so it's not a major concern in that regard
- ETP is/was a bit "complicated" (and has changed a lot since including defaults) - a lot of parts and concerns/issues
- DNT was an issue (comes with TP)
- uBO outbound connections and customizing of lists and fingerprinting of that (they have a higher threat model)
- uBO some default lists _were_ using insecure hosts
- ethics? politics? gk (at least) feels somewhat strongly about not interfering in ad income (e.g. some blogger who relies on that)... in other words, that this is not TB's problem - also see DNT (gk and almost everyone I know hated it with a passion)

That said, there have been tickets at Tor Project for some time about adding uBO/ETP - for the exact reasons you said: Reduce the attack surface. And reduce latency and load on the tor network - but the reason for any impetus behind that now, is that it is one step of many towards getting Tor into Firefox: disclosure: I spent a week with the Tor Project guys, including Roger Dingledine, Mike Perry, gk, sysrqb etc (the big guns) working on this (well, they were, I tagged along) and then at a meeting with them and eka that discussed all of this - steps to get Tor Window Mode into Firefox.

Not that Tor Project don't want to do that anyway (and there are other ways they have reduced latency etc) - but they are a small team, resources are scarce, and the timeline is somewhat dictated by other factors. They can't just throw an extension in or stop ripping out/blocking ETP - it's going to take time, analysis, testing.

Not looking to get into a discussion, just though I would share my insight

_data/software/browsers-resources/1_ublock_origin.yml
```diff
@@ -0,0 +1,12 @@
+title: uBlockOrigin
+type: Recommendation
+logo: /assets/img/browsers/ublock_origin.svg
+description: |
+  <strong>uBlockOrigin</strong> is a popular content blocker that could help you block ads, trackers, and fingerprinting scripts.
+
+  Consider enabling the <em>AdGuard URL Tracking Protection</em> <a href="https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists">filter list</a> and adding the <a href="https://raw.githubusercontent.com/DandelionSprout/adfilt/master/LegitimateURLShortener.txt">Actually Legitimate URL Shortener Tool</a> list. These filter lists use uBlock's <a href="https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#removeparam">removeparam</a> option to filter tracking parameters in addresses you might visit.
+
+  Please note that this is an <a href="https://www.ranum.com/security/computer_security/editorials/dumb/">enumerating badness</a> approach, and thus uBlockOrigin will not (and cannot) keep you safe from every single tracker out there.
```
elitejake, line 9: It's _uBlock Origin_ (separate words), not _uBlockOrigin_, right?

dngray: correct that's a typo, and also I want to work on that a bit more tomorrow.

The other cards also need proof reading/fixing. Only thing I think that is complete is the Firefox one at the moment.

freddy-m:
Might be worth mentioning FLOC under Chrome 'anti recommendation' (as per #19), if we're still doing those.

tommytran732:
> Might be worth mentioning FLOC under Chrome 'anti recommendation' (as per #19), if we're still doing those.

If you are talking about the Privacy Sandbox, then yeah, I think we really need to look deep into it

Interestingly enough however, it is enabled by default in Bromite

SkewedZeppelin:
Here is SVG for Mull:
[mull.svg.txt](https://github.com/privacyguides/privacyguides.org/files/7565780/mull.svg.txt)

dngray:
> Here is SVG for Mull:
> [mull.svg.txt](https://github.com/privacyguides/privacyguides.org/files/7565780/mull.svg.txt)

Thanks, just note, we're still thinking about this one, (I pushed that update late at night).

I'm kinda thinking for language reasons, it might be better to mention it as a footnote. We could possibly do the same with LW.

tommytran732:
> > Here is SVG for Mull:
> > [mull.svg.txt](https://github.com/privacyguides/privacyguides.org/files/7565780/mull.svg.txt)
> 
> Thanks, just note, we're still thinking about this one, (I pushed that update late at night).
> 
> I'm kinda thinking for language reasons, (and it not being on google play) it might be better to mention it as a footnote. We could possibly do the same with LW.

I am not in favor of recommending Fenix. If we are going to recommend a Firefox based browser on Android at all, I think Mull should be recommended directly instead of Fenix.

Likewise, if Arkenfox didn't exist (or if Mozilla drops user.js support), then I would drop Firefox as a recommendation and switch to Librewolf.

tommytran732:
Also, this gives more argument on Firefox on Android beyond just the increased attack surface of having 2 browser engines like I mentioned: https://grapheneos.org/usage#web-browsing

dngray:
@tommytran732 these are valid points. I think with the switch to Fenix, there is a bit of catching up on Android before Firefox there is at parity with desktop.

Less options is better so I'm not opposed to cutting it.

tommytran732:
@dngray I think we need a rebase here

tommytran732:
@freddy-m I think the content is ready for review. We just need to wait for dngray to fix one of the dependency to fix hyperlink checking I think

elitejake:
Could you change Brave's logo to SVG?

I made one [here](https://gist.github.com/elitejake/48bf3af777a60a3bd0ac4e4feeaa09f5).

tommytran732:
> Could you change Brave's logo to SVG?
> 
> I ran [this](https://gist.github.com/elitejake/48bf3af777a60a3bd0ac4e4feeaa09f5) one through scour.

Dpone

tommytran732:
@dngray @freddy-m 

This is ready for review and merge. The Matrix group helped me proof read and fix up typos with the last few commits :)

_data/software/browsers-desktop/2_brave.yml
```diff
@@ -0,0 +1,76 @@
...
+  Disable social media components\:
+  <ul>
+    <li><strong>Google:</strong> (≡ Settings → Social media blocking → Allow Google login buttons on third-party sites)</li>
+    <li><strong>Facebook:</strong> (≡ Settings → Social media blocking → Allow Facebook logins and embedded posts)</li>
+    <li><strong>Twitter:</strong> (≡ Settings → Social media blocking → Allow Twitter embedded tweets)</li>
+    <li><strong>LinkedIn:</strong> (≡ Settings → Social media blocking → Allow LinkedIn embedded posts) (off by default)</li>
+  </ul>
+
+  <h5><strong>Privacy and Security</strong></h5>
+  There are a few options in here you may want to disable.
```
rusty-snake, line 28:
```suggestion
  There are a few options in here you may want to change.
```

_data/software/browsers-desktop/2_brave.yml
```diff
@@ -0,0 +1,76 @@
...
+
+  <ul>
+    <li>Set <a href="https://support.brave.com/hc/en-us/articles/360017989132-How-do-I-change-my-Privacy-Settings-#webrtc">WebRTC IP Handling Policy</a>. We recommend either <em>Default Public Interface Only</em> or <em>Disable Non-Proxied UDP</em>. This is in (≡ Settings → Privacy and Security → WebRTC IP handling policy).</li>
+    <li>Disable (≡ Settings → Privacy and Security → Allow privacy-preserving product analytics (P3A))</li>
+    <li>Disable (≡ Settings → Privacy and Security → Automatically send daily usage ping to Brave)</li>
+    <li>Enable (≡ Settings → Privacy and Security → Security → Always use secure connections)</li>
+    <li>Enable DNS over HTTPS (DoH), if you're not on a VPN (≡ Settings → Privacy and Security → Security → Use secure DNS) with one of the choices or a <a href="/providers/dns">custom provider</a>.</li>
```
rusty-snake, line 35: Did I miss something or why is DoH (soft-)recommended for Brave but not for Firefox (both have DoH)?

_data/software/browsers-resources/1_ublock_origin.yml
```diff
@@ -0,0 +1,13 @@
+title: uBlock Origin
+type: Recommendation
+logo: /assets/img/browsers/ublock_origin.svg
+description: |
+  <strong>uBlock Origin</strong> is a popular content blocker that could help you block ads, trackers, and fingerprinting scripts.
+  
+  We suggest enabling the <em>AdGuard URL Tracking Protection</em> <a href="https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists">filter list</a> and adding the <a href="https://raw.githubusercontent.com/DandelionSprout/adfilt/master/LegitimateURLShortener.txt">ActuallyLegitimate URL Shortener Tool</a> list.
```
rusty-snake, line 7: What about "Block access to LAN"? I would recommend it to.

_data/software/browsers-desktop/2_brave.yml
dngray: Good point, we should recommend that for both.

_data/software/browsers-resources/1_ublock_origin.yml
dngray: Actually all of these are disabled by default:

- AdGuard Base 
- AdGuard Mobile Ads
- AdGuard Tracking Protection
- AdGuard URL Tracking Protection
- Block Outsider Intrusion into LAN
- Phishing URL Blocklist
- PUP Domains Blocklist


Personally I have them all enabled. That's not getting into the Annoyances/Multipurpose lists.

tommytran732: Why not just have everything enabled?

rusty-snake: Since ublock origin changes the default available list form time to time I would suggest to use something like "Enable all filterlists except for regional".

tommytran732: Lol I personally enabled everything, including regional lists (yeah, all of them)

dngray: >  would suggest to use something like "Enable all filterlists except for regional".

I do like that, keeps it brief.

tommytran732: > > would suggest to use something like "Enable all filterlists except for regional".
> 
> I do like that, keeps it brief.

Why except regional though?

dngray: well i clarified there, that if you speak the languages, you should select your regions.

More lists slows things down obviously.

tommytran732: > well i clarified there, that if you speak the languages, you should select your regions.
> 
> More lists slows things down obviously.

1. This is still very redundant. "AdGuard URL Tracking Protection" is already in the massive lists, I don't know why you would spell it out specifically. Might as well just delete that part.

2. What is the harm in enabling the regional ones even if you don't speak the language? Why not just enable all of them?

dngray:
Regarding Brave, going to just take that off this PR, it really shouldn't be with legacy content migration.

We'll put it in a branch for further discussion separately.

_data/software/browsers-resources/1_ublock_origin.yml
dngray: > This is still very redundant. "AdGuard URL Tracking Protection" is already in the massive lists, I don't know why you would spell it out specifically. Might as well just delete that part.

Wanted to emphasize you don't need CleanURLs or NeatURLs anymore because of this.

> What is the harm in enabling the regional ones even if you don't speak the language? Why not just enable all of them?

Every filter slows your browser a tiny bit I guess.

dngray (approved):

freddy-m (approved):
I take we're not doing 'anti-recommendations' anymore then lol

dngray:
> I take we're not doing 'anti-recommendations' anymore then lol

Basically some of the anti-recommendation reasons weren't terribly good. (Particularly the Brave one).

Thorin-Oakenpants:
> remove mention of addons

Still one there in the Tor card

dngray:
> > remove mention of addons
> 
> Still one there in the Tor card

Done... bleh the things I miss.

A-AhmedYasir:
What about browsers for iOS?
If you recommend sticking to Safari, why not make it clear? Maybe just a note under the "Mobile Recommendations"

dngray:
> What about browsers for iOS?

We're still discussing solutions there. It's worth noting all browsers on iOS have to use the Webkit engine.

At this point we're thinking Firefox at least offers ETP and some other features that might be worthwhile. I want to check some things with iOS users like @jonaharagon at some point regarding these recommendations.

PhysicsIsAwesome:
> https://deploy-preview-309--privacyguides.netlify.app/browsers/

The rewrite preview is a lot more reasonable than the old version. But the page seems a bit overloaded. Maybe it is a good idea to put the settings and notes in some kind of submenu instead?
Is there a reason, that JIT disabling is recommended on Bromite, but not on Firefox and is it important enough to be mentioned?
Firefox runs with Wayland by default on major Linux distributions with a Wayland-enabled DE (can be checked via about:support). So mentioning this flag could lead to more confusion, than help.

dngray:
> Is there a reason, that JIT disabling is recommended on Bromite, but not on Firefox and is it important enough to be mentioned?

Not so much it's recommended, rather it is an optional security enhancement that comes at the cost of performance.

Most browsers don't allow you to disable it. I think only Vanadium, Bromite and Hexavalent have options for that.

Firefox requires you to mess with about about:config these are [the options](https://github.com/arkenfox/user.js/blob/master/user.js#L1183).

We're still deciding whether or not we should mention it.

qua3k:
Browser exploits rely on being able to call `VirtualAlloc` or `VirtualProtect` to easily copy shellcode into memory; Chrome enables ACG and CET shadow stacks in the renderer when JIT compilation is disabled, preventing an attacker from generating their own code after hijacking the control flow as is the norm. It's going to cause significant pain for attackers and should be enabled by default if the browser easily allows site opt-out in settings such as Edge is now. There is no equivalent W ⊕ X policy on Linux but Hexavalent is researching modifying the seccomp policy to do the same.

dngray:
@qua3k I don't doubt there is risk. However I think enabling/disabling it has pros and cons. Enabling does grant better security, but it does seem that is a [fingerprintable](https://lists.torproject.org/pipermail/tbb-dev/2021-August/001211.html) metric. Obviously things that require WASM aren't going to work at all, and some things may work with less performance. So seems to me like security vs privacy/performance.

Most browsers don't allow for disabling it. Firefox doesn't have an exposed UI element so you'd have to set `about:config` options, meanwhile there is no option on iOS where everything is Webkit.

I'm thinking if my understanding is correct elaborating on that and putting it in the blog article. For the Bromite card we might then just say "consider disabling", (and cross link to the blog article).

csagan5:
> Interestingly enough however, it is enabled by default in Bromite

It has never been enabled/effective on Bromite; some UI remnants were present but the underlying code was deleted/disabled since the very first day FLoC was released upstream in Chromium.

PR #323, opened 2021-11-16
Remove LessPass and Spectre

freddy-m:
## Description

Removed LessPass and Spectre.

### DIFF
--- a/_includes/legacy/sections/password-managers.html
+++ b/_includes/legacy/sections/password-managers.html
@@ -49,31 +49,9 @@
   chrome="https://chrome.google.com/webstore/detail/keepassxc-browser/oboonakemofpalcgghocfoadofidjkkk"
 %}
 
-{%
-  include legacy/cardv2.html
-  title="LessPass - Browser"
-  image="/assets/img/legacy_svg/3rd-party/lesspass.svg"
-  description="<strong>LessPass</strong> is a free and open-source password manager that generates unique passwords for websites, email accounts, or anything else based on a master password and information you know. No sync needed. Uses PBKDF2 and SHA-256. It's advised to use the browser addons for more security."
-  website="https://lesspass.com/"
-  privacy-policy="https://addons.mozilla.org/en-US/firefox/addon/lesspass/privacy/"
-  github="https://github.com/lesspass/lesspass"
-  windows="https://pypi.org/project/lesspass/"
-  mac="https://pypi.org/project/lesspass/"
-  linux="https://pypi.org/project/lesspass/"
-  freebsd="https://pypi.org/project/lesspass/"
-  firefox="https://addons.mozilla.org/en-US/firefox/addon/lesspass/"
-  chrome="https://chrome.google.com/webstore/detail/lesspass/lcmbpoclaodbgkbjafnkbbinogcbnjih"
-  fdroid="https://f-droid.org/packages/com.lesspass.android"
-  googleplay="https://play.google.com/store/apps/details?id=com.lesspass.android"
-%}
-
 <h3>Worth Mentioning</h3>
 
 <ul>
-  <li>
-    <a href="https://spectre.app">Spectre App</a> - A password manager based on an ingenious password-generation algorithm that guarantees your passwords can never be lost. Its passwords aren't stored: they are generated on-demand from your name, the site, and your master password. No syncing, backups, or internet access needed.
-  </li>
-
   <li>
     <a href="https://psono.com/">Psono</a> - Free and open source password manager for teams with client side encryption and secure sharing of passwords, files, bookmarks, emails. All secrets are protected by a master password. Uses <a href="https://nacl.cr.yp.to/">NACL Crypto</a>, a combination of Curve25519, Salsa20 and Poly1305.
   </li>


tommytran732:
Just in case anyone is wondering why, here is the explanation:

If a malicious actor, let's say me, somehow gets your master password:
- If you use a normal password manager like Bitwarden with 2FA, I still cannot get my hands on your database to decrypt it
- If you use KeepassXC, I have to somehow access your computer to steal your password database first. And you can make the database extra safe by using password + HMAC to protect it, so even when I get your database and know the master password I still cannot unlock it
- With these 2 password managers, if I know your master password, all I need to do is to find your username/email address(which is trivia cuz usernames are public and email addresses are not confidential ), I can derive every single password you have and completely mess you up

tommytran732:
@freddy-m I see that you haven' deleted the icon, might as well just do it for house keeping lol

dngray:
This was merged in https://github.com/privacyguides/privacyguides.org/commit/22e62f905eab0c5e994e5fb43f7b80ee9d7521cf not sure why it didn't automatically close.

freddy-m:
> @freddy-m I see that you haven' deleted the icon, might as well just do it for house keeping lol

Oh yeah, there are probably quite a few old icons that need to go. Thanks @tommytran732!

freddy-m:
> This was merged in 22e62f9 not sure why it didn't automatically close.

Is that because it wasn't reviewed?

samuel-lucas6:
> 
> 
> Just in case anyone is wondering why, here is the explanation:
> 
> If a malicious actor, let's say me, somehow gets your master password:
> 
>     * If you use a normal password manager like Bitwarden with 2FA, I still cannot get my hands on your database to decrypt it
> 
>     * If you use KeepassXC, I have to somehow get access to your computer to steal your password database first. You can also make the database extra safe by using password + HMAC to protect it, so even when I get your database and know the master password I still cannot unlock it
> 
>     * With these 2 password managers, if I know your master password, all I need to do is to find your username/email address(which is trivia cuz usernames are public and email addresses are not confidential), and I can derive every single password you have and completely mess you up

I'm currently working on a command line alternative to Spectre as a fun project. Using a full name or username as a salt is bad practice. You really have to use a random salt, which means storing at least that somewhere. Then a keyfile can be used as a pepper, like a form of 2FA.

lhunath:
You might have reached me for comment prior to concluding this assessment, in order to validate whether your logic is complete and sound.

To reduce the comparison of these models to nothing more than that of their respective authentication passphrases is an injustice inflicted primarily upon your own audience.

@tommytran732 draws a false equivalency between the password used to sign into Bitwarden, the password used to unlock an encrypted KeePassXC database, and the secret component in Spectre's algorithm. Neither of these tokens are remotely equivalent to each other. Furthermore, the comparison ends there, ignoring any other components of the threat models of each of these solutions as somehow irrelevant.

I should also posit that the argument that begins with "If I <control the secret factors>", necessarily concludes with the solution being broken, but merely positing it is hardly proof of anything. "If I am a rogue developer at Bitwarden, all I need to do is X" doesn't prove anything.

If you are serious about providing a trustworthy security service to your audience, I suggest you evaluate these recommendations based on a complete threat model which invites community and expert participation.

@samuel-lucas6 FYI, https://gitlab.com/spectre.app/cli/ is a command-line implementation of the Spectre algorithm. Regarding your assessment that a full name as a salt is poor practice, I'd love to discuss what makes you feel this way, perhaps in an issue on the project, or you may reach me at support@spectre.app. A random salt defeats the entire driving principle which forms the foundation of a deterministic password generator. If keeping a state dependency is an acceptable trade-off to you, you really should be using an encrypted vault as your state instead of a random salt.

Open to questions if anyone has any.

tommytran732:
@lhunath 

I do not see how this is false equivalency. Under which specific circumstances/use case would it make more sense to use Spectre over Bitwarden and Keepass?

If we are talking about a rogue developer, both Bitwarden (especially the cloud hosted version) and Spectre can quite easily get access to the user's login (you can just use malicious JS to get the user's secret).

However, in the case of a leaked master password, both Bitwarden and Keepass can protect the user against a malicious third party (as I elaborated above). Bitwarden has 2FA for protection, and Keepass requires access to the device of some form to steal the database. In the case of Spectre, it would be the end of the story. The attacker can derive just about every secret a user has. Even lesspass has the variable password lengths and the "counter" for a bit of increased randomness.

If an attacker has compromised the user's device, then yes, none of these 3 password managers will save them.

At the end of the day, with what use case would it make sense to derive all of your secrets from a single master password?

tommytran732:
In fact, I have just found this article: https://tonyarcieri.com/4-fatal-flaws-in-deterministic-password-managers
"Exposure of the master password alone exposes all of your site passwords" is just one of the few problems explained there. 

Spectre is stateless. It can't just simply bump the counter like the stateful one (Lesspass in our case). Even if a user only the password of a single site leaked, it is not going to be very pleasant for them, because then they will need to either start remembering 2 passwords or go through the trouble of changing all of their current passwords to one derived from a new master password.

lhunath:
Hi @tommytran732,

The false equivalency I mentioned is in reference to the fact that a Bitwarden password, KeePassXC keyphrase and Spectre secret are not equivalent to each other in terms of their respective threat model. While they may appear similar in the sense that "each is information a user knows", within the Bitwarden threat model, the password plays a very different role than it does within the KeePassXC threat model, and entirely different yet from Spectre's threat model.

When anyone evaluates Spectre as a solution for them, there is a deliberate choice that the entire threat model revolves around the individual.
The trade-off here is that where the Spectre secret is the pillar upon which the threat model rests, what you get in exchange is liberation from reliance on external factors that are transient or beyond your control.
This is something none of the other solutions are capable of offering.

You would choose Spectre because the knowledge that you do not rely on hardware, which may be lost, or trust, which may be violated, is more important to you than the chore associated with assuming personal responsibility over safekeeping a strong Spectre secret.

samuel-lucas6:
> @samuel-lucas6 FYI, https://gitlab.com/spectre.app/cli/ is a command-line implementation of the Spectre algorithm. Regarding your assessment that a full name as a salt is poor practice, I'd love to discuss what makes you feel this way, perhaps in an issue on the project, or you may reach me at support@spectre.app. A random salt defeats the entire driving principle which forms the foundation of a deterministic password generator. If keeping a state dependency is an acceptable trade-off to you, you really should be using an encrypted vault as your state instead of a random salt.

Hi @lhunath, nice to hear from you here. I should clarify that I think your software is a very nice idea, it looks the part, and it's great that it's available on so many platforms. However, having a random salt does provide some additional protection. On the other hand, I completely understand what you mean about not having to store anything. I was aware of the CLI tool, but I was planning on changing a few of the algorithms. I've been debating the full name/username vs random salt aspect myself. I will drop you an email to avoid clogging up this issue. Thank you for the invitation to talk.

dngray:
> In fact, I have just found this article: https://tonyarcieri.com/4-fatal-flaws-in-deterministic-password-managers

I read that article, and there was another part there I found relevant:

  *But beyond that, a deterministic scheme cuts you out of the ability to store data which is not derived deterministically in a password vault.*

I think this would be a major one to a lot of users, secret question/answers or telephone PIN codes etc. The fact is institutions already give users a lot of things they need to remember besides a username and password.

> Even if a user only the password of a single site leaked, it is not going to be very pleasant for them, because then they will need to either start remembering 2 passwords or go through the trouble of changing all of their current passwords to one derived from a new master password.

I don't see that as practical, and we do see breaches from time to time. It's also worth noting they must use the same username/email and that can be a problem for privacy as it really makes using services like anonaddy/simplelogin impossible.

Additionally if a site's domain changes (re brands etc), you must change your password.

> The trade-off here is that where the Spectre secret is the pillar upon which the threat model rests, what you get in exchange is liberation from reliance on external factors that are transient or beyond your control.

As ideal as that sounds, I think users are always going to have data that can't be deterministically derived, so why not backup the encrypted secrets database with that data.

> To fairly compare these different solutions, one would need to balance the weight of safekeeping the Spectre secret against the weight of safekeeping a Bitwarden password while also trusting that Bitwarden's employees are well-behaved, and the company's internal security is never compromised.

It's worth noting that most commericial password manager offerings have to undergo compliance checks (eg SOC 2 Type II and SOC 3). If those compliance checks aren't to be trusted, then encrypting your secrets with a local manager (like keepassxc) makes sense.

I don't see a threat model in which:

- you can't rely on anyone's (company etc) services
- you need to store secrets for services that are not local

> More and more, people are unwilling to accept trusting corporations or that our devices won't be confiscated at the border, and more willing to become solely responsible for the security of their own identity.

A cloud provider like bitwarden solves that problem too. It has the added benefit of no evidence of it being on your device, (even if they look at browsing history), you could say it was "someone else who used your computer".

The existence of Spectre on your device tells your adversary exactly what they need to get out of you, in order to access everything.

lhunath:
I'm afraid commenting on the entire article would likely be well beyond the scope of this discussion, but I'm happy to provide clarity on whatever topics you like to raise here.

The main issue with this article is that it proposes a very black-and-white picture of deterministic password management and then purports this as a figurative straw man through which to assassinate the very concept. This approach shines through in many areas of the author's arguing, but I'm sure we'll get to concrete instances of how that leads them astray.

> But beyond that, a deterministic scheme cuts you out of the ability to store data which is not derived deterministically in a password vault.

A stateless deterministic scheme cannot store state, correct. A stateful deterministic scheme can store state, of course. If the argument is that a stateless scheme does not offer the ability to use the application as a store for persisting information on your sites, that is indeed well perceived: storing information is not its objective.

However, two counter-points: not being able to store information does not preclude the user's ability to use the solution for security questions/answers, phone/bank PIN numbers, etc. In fact, Spectre users are actively using these capabilities in Spectre today in a fully stateless manner.

Secondly, a solution such as Spectre need not be exclusively stateless. The function of its statelessness is to [1] enable recovery of the critical components in a total loss situation, and [2] expunge those critical components from the hardware device. These objectives can still be met with a hybrid approach. In fact, again, Spectre employs a hybrid model which allows users today to persist all sorts of important information with their site, such as for example login names, by encrypting them using the user's statelessly derived Spectre key and storing the result locally. This solution can also be used for storing custom passwords, with the caveat that this approach reverts the site to a typical vault-status, making it unrecoverable in the event of loss.

To pretend that these solutions need to be an all-or-nothing is an argument made in bad faith.

> Even if a user only the password of a single site leaked, it is not going to be very pleasant for them, because then they will need to either start remembering 2 passwords or go through the trouble of changing all of their current passwords to one derived from a new master password.

Rolling compromised passwords is a very important capability which is 100% supported by Spectre's stateless model. Each site is accompanied by a site counter, which can be incremented when the site's password needs to be expired. The counter is stored locally and in the event that it needs to be recovered (again, recovery is an exceptional situation), the process involves the user starting the counter at 1 and bumping it until they receive a password the site accepts, at which point the counter value has been recovered. It's really not that big of a deal, especially if you keep a record of your site metadata (which contains no secrets, so is easy to backup safely).

I can assure you of the thousands of people who use Master Password and Spectre, this is not an issue. I have had no support requests yet where people hit a wall on this in the product's 10 year lifetime.

> I think users are always going to have data that can't be deterministically derived, so why not backup the encrypted secrets database with that data.

I have data which I would like to store securely. A lot of it. It doesn't go in a password manager. It gets encrypted and distributed. My encryption key comes from Spectre. Spectre is fully compatible with the idea that you should have data which you want to keep safe.

As for "why not", the answer is simply: Spectre offers you a mechanism which is immune to data loss. If I should choose whether I want my digital identity to get encrypted in a file I need to make sure not to ever lose, or immune to any harm from external sources beyond my absolute control, I will happily choose the latter. I don't have that option for all of my data, but I definitely have it for my digital identity. And that is a lot.

> It's worth noting that most commericial password manager offerings have to undergo compliance checks (eg SOC 2 Type II and SOC 3).

This is a perfectly fair assessment to make. But not everybody will agree with that being adequate. Google was SOC compliant when PRISM was exposed. Amazon is SOC compliant and we just learned its employees have been spying on purchase histories of romantic interests and more. You'll forgive a user-base if it isn't comforted by this argument.

> I don't see a threat model in which:
> - you can't rely on anyone's (company etc) services
> - you need to store secrets for services that are not local

It doesn't need to be a black-and-white situation. The argument that you need to accept some amount of trust in services doesn't suffice to say that it is now acceptable to give them your entire digital identity and trust it will remain both permanently available and uncompromised. Multi-layered security and zero-trust are important concepts in security architecture.

> A cloud provider like bitwarden solves that problem too. It has the added benefit of no evidence of it being on your device, (even if they look at browsing history), you could say it was "someone else who used your computer".

Recognize that there is a growing user-base which does not want to swallow the dogma that everything that we own should be rented from and managed by a corporation, especially one which likely does not operate within your country's legal jurisdiction, and is not governed democratically.
People choose Spectre because they do not want to upload their identity to a data center, and that is a valid concern to have.

> The existence of Spectre on your device tells your adversary exactly what they need to get out of you, in order to access everything.

If your adversary is in a position to employ force onto your person to compel you to give them what they want, there is no amount of protection that will make you safe, and you should probably give them whatever they ask for.

samuel-lucas6:
> But beyond that, a deterministic scheme cuts you out of the ability to store data which is not derived deterministically in a password vault. 
>
> It's also worth noting they must use the same username/email and that can be a problem for privacy as it really makes using services like anonaddy/simplelogin impossible.
>
> As ideal as that sounds, I think users are always going to have data that can't be deterministically derived, so why not backup the encrypted secrets database with that data.

I agree that this is really the biggest problem. If the concern is storing passwords in the cloud, then you can just use KeePassXC and back up the file to external hard drives and memory sticks. Storing all of this information in one location is far more convenient. Most people wouldn't want to faff around with decrypting a file containing this other information. Even if you just stored emails/usernames in an unencrypted text file, it would still be slower to access.

> Additionally if a site's domain changes (re brands etc), you must change your password.

I suppose it depends. If you could remember the old domain, then you wouldn't necessarily have to, but it would be sensible in case you forgot. Either way, I don't think this is a major problem.

> A cloud provider like bitwarden solves that problem too. It has the added benefit of no evidence of it being on your device, (even if they look at browsing history), you could say it was "someone else who used your computer".
> 
> The existence of Spectre on your device tells your adversary exactly what they need to get out of you, in order to access everything.

You might be able to get away with just using the Spectre website. Furthermore, most people likely have Bitwarden installed, and if someone wants to force a password out of you, then there will be a way of doing it.

> Each site is accompanied by a site counter, which can be incremented when the site's password needs to be expired. The counter is stored locally and in the event that it needs to be recovered (again, recovery is an exceptional situation), the process involves the user starting the counter at 1 and bumping it until they receive a password the site accepts, at which point the counter value has been recovered. It's really not that big of a deal, especially if you keep a record of your site metadata (which contains no secrets, so is easy to backup safely).

This works but is less ideal than using a password manager. If you incremented the counter a lot, then recovery would be annoying.

>  People choose Spectre because they do not want to upload their identity to a data center, and that is a valid concern to have.

I agree, but KeePassXC also satisfies this criteria.

> Regarding your assessment that a full name as a salt is poor practice, I'd love to discuss what makes you feel this way

Lastly, regarding the salt matter, I will repeat what I said in my email since this discussion is ongoing. It's not a practical problem if the user chooses a strong Spectre secret. However, if it's very short or something in a password dictionary, then the user is probably in trouble. The memory-hard KDF slows attacks down but won't save you. If a random salt was stored offline, out of reach of an attacker prior to device compromise, then that would provide additional protection. Similarly, some method of 2FA would help.

The situation is similar to the case of file encryption since the salt is stored in the encrypted file. The difference is that the attacker needs to access the file to know the salt, whereas with Spectre, you might be able to grab someone's salt off social media, some company website, or an email address.

The fact that multiple people may use the same full name/username also helps aid precomputation attacks against weak passwords. Then if you need to change the Spectre secret, you should change the salt, but that's not possible if you always use your full name. 

In other words, a username based salt doesn't fulfil the typical unpredictable and unique salt criteria.

lhunath:
For those interested in further reading and some actual concrete numbers to back up the claims, see: https://chat.spectre.app/d/46-security-critism/5

PrivacyBay:
> For those interested in further reading and some actual concrete numbers to back up the claims, see: https://chat.spectre.app/d/46-security-critism/5

Thanks for sharing, Ihunath. Shocking that they've removed them anyway. But you've said it already...  "It's a common opinion, especially from amateur security enthusiasts."

dngray:
> For those interested in further reading and some actual concrete numbers to back up the claims, see: https://chat.spectre.app/d/46-security-critism/5

Which are hidden behind useless "Login to see the code"... 61 times.

> Shocking that they've removed them anyway

It was removed because the arguments provided by @lhunath were largely their personal opinion.

lhunath:
> Which are hidden behind useless "[Login to see the code](https://user-images.githubusercontent.com/48640805/203991887-ee4dece9-79c1-47f9-89ed-78b19d4544e7.png)"... 61 times.

My apologies, this was not my intention. I have corrected the board's configuration.

PR #330, opened 2021-11-16
Update the homepage  of Lemmy

elitejake:
## Description

I think [join-lemmy.org](join-lemmy.org) is the website of Lemmy. The [old one](dev.lemmy.ml)'s certificate is not valid and [their repositories ](https://github.com/LemmyNet) point to [join-lemmy.org](join-lemmy.org).


freddy-m (approved):
Good spot!

PR #337, opened 2021-11-18
Re-add privacy_policy functionality

elitejake:
## Description

Add optional `privacy-policy` to [recommendation layout](https://github.com/privacyguides/privacyguides.org/blob/main/_layouts/recommendation.html).

Haven't tested it but it should look exactly like `privacy-policy` in legacy [`cardv2.html`](https://github.com/privacyguides/privacyguides.org/blob/main/_includes/legacy/cardv2.html).

another

### DIFF
--- a/_includes/recommendation-card.html
+++ b/_includes/recommendation-card.html
@@ -18,6 +18,7 @@
           More Information</a>{% endif %}
         {% if item.website %}<a href="{{ item.website }}" rel="noopener" class="btn btn-secondary mt-1 me-1"><i class="fad fa-external-link-square fa-fw"></i>
           Visit {{ item.website | remove_first: "www." | split: "/" | slice: 2 }}</a>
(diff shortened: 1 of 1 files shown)


tommytran732:
It's broken: https://deploy-preview-337--privacyguides.netlify.app/browsers/
Tor's privacy policy should show but it doesn't

tommytran732:
Maybe {{page.privacy-policy}} needs the space between them? Like {{ page.privacy-policy }}?

@dngray if you could look at this it would be great, the privacy policy needs to be fixed urgently before we convert more pages to the new format

dngray:
> Maybe `{{page.privacy-policy}}` needs the space between them? Like `{{ page.privacy-policy }}`?

It actually needs to be `{{ item.privacy_policy }}`, and it was in the wrong file.

PR #342, opened 2021-11-18
Search engines revamp

tommytran732:
Revamping the search engine page, removing search engines with bad privacy policies and convert to the new layout

Change log:
1. Removed QWANT due to bad privacy policies - data is collected and shared with third parties: 
https://about.qwant.com/en/legal/confidentialite/
https://tosdr.org/en/service/527

2. Remove Metager due to personalized ads and possible censorship
https://metager.org/datenschutz
https://tosdr.org/en/service/2658

3. Removed Yacy - development has practically been dead for years and has barely picked up recently. The latest release is still from 2016. The search results aren't that good either, it does not sort them based on relevancy.
https://linuxreviews.org/YaCy


rusty-snake:
TBH We should list them in a "Why is search engie X not listed?" section instead of removing.
1. To avoid new suggestions
2. To make it clear that they are not recommended (distinct form "we only list foo and bar to make decisions easier"). Since other "privacy-sites" recommend them (and others like DDG not) this makes it easier to understand the decision PG did.

tommytran732:
> TBH We should list them in a "Why is search engie X not listed?" section instead of removing.
> 
>     1. To avoid new suggestions
> 
>     2. To make it clear that they are not recommended (distinct form "we only list foo and bar to make decisions easier"). Since other "privacy-sites" recommend them (and others like DDG not) this makes it easier to understand the decision PG did.

I do try very hard to list reasons on why something is recommended
IDK how I would do it for something in the questionable zone, cuz like, they are not so bad that they deserve recommendations against them, they just aren't as good as the ones we do recommend

dngray:
> TBH We should list them in a "Why is search engie X not listed?" section instead of removing.

I've been thinking about this and some of the other *extra* information we used to have on the MediaWiki.

Maybe we could use a wiki on this github organization. I rather like the idea of that, because you can check out wiki repos (they're just markdown) and seeing as people are already here thinking about recommending something it kinda makes sense.

A-AhmedYasir:
Am I not able to find it, or did you remove Searx?

tommytran732:
> Am I not able to find it, or did you remove Searx?

It's not ported to the new layout yet.

fxbrit:
@tommytran732 after reading metager policy I don't get the reasoning for removing it. it says non-personalized ads:

> In addition to donations and membership fees, we must finance our operation through non-personalized advertising on the results page. To receive this advertising, we give the first two blocks of the IP in connection with parts of the so-called user agent to our advertising partners

I also don't get the censorship argument, can you guide me? maybe I'm missing something.

tommytran732:
> @tommytran732 after reading metager policy I don't get the reasoning for removing it. it says non-personalized ads:
> 
> > In addition to donations and membership fees, we must finance our operation through non-personalized advertising on the results page. To receive this advertising, we give the first two blocks of the IP in connection with parts of the so-called user agent to our advertising partners
> 
> I also don't get the censorship argument, can you guide me? maybe I'm missing something.

Er, well, here is what they got to say:
"For this purpose alone [anti spam botting they are talking about], we store the full IP address and a timestamp for a maximum of 96 hours."
"We must finance our operation through non-personalized advertising on the results page. To receive this advertising, we give the first two blocks of the IP in connection with parts of the so-called user agent to our advertising partners"

So basically they share the first 2 blocks of your IP and user agent to whichever third party they partner with to display ads. I mean it's not a big issue or anything, I just think it's a little odd why they are sharing the first 2 blocks and the user agent in the first place.

They also said
"We also reserve the right to block web pages with demonstrably incorrect information, web pages of extremely poor quality and other particularly dubious web pages."
https://metager.org/transparency

I mean, its not a huge huge issue with Metager or anything, it's simply that neither DDG or Startpage or Mojeek does this in the first place, so idk if we should bother listing them.

At the end of the day, TOS stuff is not my strongest suite (I tend to never care for them and always layer up my defenses). I am not quite sure where to draw the line. 

If you think these things are acceptable and that Metager should still be listed along with search providers who don't do this, please let me know.

tommytran732:
And yeah, I think probably the non-personalized ads was wrong. I was quite tired when I was reading it on TOSdr. I still think it's odd - if they don't do personalized ads, why would they share those infos in the first place?

fxbrit:
I'm not an expert either tbh, maybe someone can help us figuring this out in a clearer manner. with that being said:

> "For this purpose alone [anti spam botting they are talking about], we store the full IP address and a timestamp for a maximum of 96 hours."

this is ok imo.

> "We must finance our operation through non-personalized advertising on the results page. To receive this advertising, we give the first two blocks of the IP in connection with parts of the so-called user agent to our advertising partners"

I think this is not that invasive, but of course it depends on what PG considers acceptable. I personally don't think this is a big deal.

> "We also reserve the right to block web pages with demonstrably incorrect information, web pages of extremely poor quality and other particularly dubious web pages."

that's their way of saying they will not tolerate fake news I guess. keep in mind that other engines that get their results from bing or google are subject to the censorship made by these engines.

> why would they share those infos in the first place?

they probably want to show relevant ads, but not targeted ones. but don't quote me.

so you get my point, I don't feel qualified to say you're making I mistake but it's probably worth giving it a second look before going on. what do you think?

tommytran732:
Well, I simply think along the lines of

"Hey they are not that bad at all. But.. the other ones are better, and they are certainly not as good as the other ones we recommend... so why would we even bother?" 

Would be great if a team member comment on this though because I am not sure tbh. I got no strong feelings either ways.

dngray:
I'm thinking we should be very clear about what is recorded on each search engine.

Tbh I'm not a fan of targeted advertising, that's literally what Google does.

tommytran732:
This is ready for review and merge.

dngray (approved):

PR #356, opened 2021-11-21
Add back PeerTube, Invidious, remove LBRY

julianfairfax:
## Description

#274 removed PeerTube under the grounds of it not providing a way to have a private subscription list, and replaced it with LBRY, [a platform that condones and promotes misinformation and nazi content](https://en.wikipedia.org/wiki/LBRY#Moderation). It also removed Invidious, confusing many, and accomplishing little.

This PR adds back PeerTube on the grounds of it being the best technical implementation of an alternative to YouTube despite the private subscription list, removes LBRY due to it's objectively deplorable moderation team, and reinstates Invidious as a viable frontend to YouTube on the web. Both PeerTube and Invidious can be self-hosted, alleviating supposed issues, whereas LBRY may be possible to self-host, but this isn't common practice.

PR #358, opened 2021-11-21
video-streaming: delist LBRY until proper process

Mikaela:
PrivacyGuides has a policy of not de/listing software without two team members approving it. #274 was force-merged outside of that process without @privacyguides/team having a chance of properly reviewing it.

This PR removes LBRY from the site and may be relisted later when it has been sufficiently researched and discussed.

Ref: #356.


julianfairfax:
While I disagree with LBRY being included in the future, I agree it should follow proper process if it is to be.

tommytran732:
I disagree that it should be delisted solely based on the basis that bad people abuse it. Unless something is explicitly made for malicious purposes, the primary concern should be the technology itself, not social issues around it.

Under the same logic, you would have to delist tor, i2p, because guess what? Bad people use them. A lot.

dngray:
I don't believe technology should be evaluated on the worst actors, (by that logic we'd basically be left with only centralized services by well resourced companies) so I don't support removing it either.

It's worth noting that *only* the Odysee  website is centralized, and they do have a community guideline of what they allow: https://odysee.com/@OdyseeHelp:b/Community-Guidelines:c

Mikaela:
The issue here is that LBRY was merged in without proper process. If you have any other issue, please discuss it somewhere more appropiate.

dngray:
> The issue here is that LBRY was merged in without proper process. If you have any other issue, please discuss it somewhere more appropiate.

I don't think we need to de-list it to decide that, that just seems like more work, delist, it, someone puts in a PR to re-list it.. what's the point in that?

dngray:
So I've done some more digging in to this today.

- A lot of major YT channels actually mirror there stuff here.
- I disagree with delisting something purely because "some bad people" used it. It's also worthwhile noting there are plenty of other decentralized video platforms pushing a crypto currency (some of them may see LBRY as a competitor), so it's entirely possible bad press will be given oxygen to make the most out of it. Its worth noting those platforms generally are centralized with a coin system just to make money as opposed to LBRY which indexes on their blockchain
- We're trying to remain focused on privacy (and that intersects with security) keep that in mind when evaluating choices we recommend
- It's censorship resistant, which is nice for a change when compared to YT that often makes false positives. There will be plenty of YT creators that will point out their worst fears regarding false positive deplatforming or DMCA complaints by copyright trolls
- The software does work through Tor, (as does the website)
- By citing "nazis" as the only reason we're basically labeling the network as the place for those people to go, additionally we're ignoring any of the benefits it has to privacy
  - It's also worth noting if we go down that path we might as well de-list Tor (because child porn and drugs), IPFS (because people post objectionable content on there), Freenet, and any other decentralized platform. Also in my experience when I moderated https://tube.privacytools.io there was a fair bit of "questionable" content (antivaxxer, nazi) material posted on other instances.
  - Major YT channels do use it, this includes Techlore, (which was on the front page of Odysee), Pine64, Purism and many open source/Linux tech channels. Even Braxman is there... I even some anti-alt-right videos on the front page.
Screenshot 2021-11-22 at 19-24-19 odysee com

PR #390, opened 2021-11-26
Revamping the Android section

tommytran732:
Migrating Android into it's own navbar item using the new format.

Closes: https://github.com/privacyguides/privacyguides.org/pull/79

Link: https://deploy-preview-390--privacyguides.netlify.app/android/


SkewedZeppelin:
Some notes:
- DivestOS includes Mulch WebView on all versions which is largely based on the Vanadium WebView. It is only the only known WebView for 32-bit that enables CFI.
- DivestOS includes all of the legacy AndroidHardening kernel hardening patches and the bulk of the modern GrapheneOS kernel hardening patches.
  - DivestOS goes even further with defconfig hardening too: https://github.com/Divested-Mobile/DivestOS-Build/blob/master/Scripts/Common/Functions.sh#L698 This is far more effective than most people consider as the AOSP trees often have security features backported but disabled, eg. page poisoning is available on most 4.4 trees despite being a 4.6 feature
- DivestOS includes the GrapheneOS hardened_malloc on 16.0 and 17.1 for all 64-bit devices. I would like to bring this to 18.1 eventually.
- DivestOS has the exec_spawning feature on 16.0 and 17.1 but is disabled due to breaking things like VoLTE, however I very likely ported it wrong. This is lower priority, as effective as it is, it caused older devices like the Nexus 4 to take very long to open apps.
- New AdAway has a VPN mode fwiw

tommytran732:
Right I will add a note for this

SkewedZeppelin:
I'd note that the version of Orbot available on their F-Droid repo and Google
Play is often out-of-date.
The latest version from GitHub, currently 16.5.2, has many performance,
reliability, security, and arguable privacy improvements.

https://github.com/guardianproject/orbot/releases

tommytran732:
> I'd note that the version of Orbot available on their F-Droid repo and Google Play is often out-of-date. The latest version from GitHub, currently 16.5.2, has many performance, reliability, security, and arguable privacy improvements. https://github.com/guardianproject/orbot/releases edit: also I don't think it was ever confirmed, but I'd recommend isolate destination addresses when using the app vpn mode. edit2: also perhaps recommend the order of use for an app be vpn mode first, then socks5, then http. an app using http proxy won't route dns through tor iirc

That was the intension lmao

ph00lt0:
If we recommend people to buy preloaded GrapheneOS phones, we should add some explanation on how to verify the devices integrity.

ph00lt0:
In the device shopping section we might need to add recommendation to pay in cash. This prevents the IMEI number to be linked to you. This obviously also requires you not to use a phone plan linked to you.

SkewedZeppelin:
I like it! Thank you @tommytran732 for this page among others.
You are raising the bar, hopefully the communities will realize it.

dngray:
> In the device shopping section we might need to add recommendation to pay in cash. This prevents the IMEI number to be linked to you. This obviously also requires you not to use a phone plan linked to you.

In practice though I'm pretty sure most countries require drivers license/details when activating a SIM card though don't they?

Its also worth noting a SIM card keeps a record of every IMEI it has come in contact with and [regularly transmits that](https://www.youtube.com/watch?v=0Em-J_3QYu4).

ph00lt0:
> In practice though I'm pretty sure most countries require drivers license/details when activating a SIM card though don't they?
> 
> Its also worth noting a SIM card keeps a record of every IMEI it has come in contact with and [regularly transmits that](https://www.youtube.com/watch?v=0Em-J_3QYu4).

https://www.comparitech.com/blog/vpn-privacy/sim-card-registration-laws/

and yes it does. So you shouldn't use the same sim card in a new phone.

tommytran732:
I think that would be highly dependant on the threat model (do you wanna be anonymous or do you just wanna private with your data)

Plus its a general phone recommendation thing, not an Android thing, so i think we're fine eithout mentioning it

tommytran732:
Oh by the way we will have an iOS section at some point too, and hopefully with the same kind of content (here are some apps to use, which iphones are acceptable, what you should avoid doing, blah blah)

SkewedZeppelin:
> DivestOS does not handle firmware updates well currently.

Here is the status of that: https://gitlab.com/divested-mobile/firmware-empty/-/blob/master/STATUS

tommytran732:
@SkewedZeppelin So I noticed LineageOS added the toggles for data and wifi connection in the application information area for each app. I just wanna confirm - Divest also inherits the same settings, right?

SkewedZeppelin:
> @SkewedZeppelin So I noticed LineageOS added the toggles for data and wifi connection in the application information area for each app. I just wanna confirm - Divest also inherits the same settings, right?

Indeed.
https://divestos.org/images/screenshots/11/Network.png

dngray:
> IMEIs starting with 35 cannot be bootloader unlocked

Is this a standard thing? Or are there other IMEIs that don't start with 35 that also can't be unlocked?

collections/_evergreen/android.html
```diff
@@ -101,6 +101,8 @@ <h5><strong>Android Permissions</strong></h5>
 <h5><strong>Android Device Shopping</strong></h5>
 <p>Google Pixels are known to have good security and properly support <a href="https://source.android.com/security/verifiedboot">verified boot</a>. Some other phones such as the Fairphone and Oneplus devices do also support custom Android verified boot (AVB) key enrollment however, there have been issues with their older models. In the past they were using <a href="https://social.coop/@dazinism/105346943304083054">test keys</a> or not doing proper verification, making verified boot on those devices useless.</p>
 
+<p>Avoid buying phones from mobile network operators. Often these have a <strong>locked bootloader</strong> that does not support <a href="https://source.android.com/devices/bootloader/locking_unlocking">OEM unlocking</a>. These phone variants will prevent you from installing any kind of alternative Android distribution. That includes any phones from purchased from Verizon, Telus, Rogers, EE etc.</p>
```
SkewedZeppelin, line 104: The 35xxx IMEI is the easiest way to discern them. Sellers are often happy to list the devices as carrier unlocked and not know/label what true carrier it came from.

dngray: I'd be careful generally buying any phones from ebay, especially Pixels. Is there any good reason/market where you can't buy a Pixel?

SkewedZeppelin: Where else would you buy them?
New in box from Google? that requires a Google account
New in box from Best Buy? That requires $300+.
Not everyone can afford these devices, one of the primary reasons I made DivestOS.

dngray: tbh i wouldn't buy a secondhand phone from ebay, possible the complete IMEI might even be blacklisted if it was stolen.

SkewedZeppelin: Nah, most of the bigger sellers actually buy them from the carriers or the carrier warranty partnerships.
I've bought at least 3 dozen phones off of eBay and never has one been blocked.

freddy-m (approved):
Give it one final check, but it looks good to me!

SkewedZeppelin:
Nit:
> includes kernel patches from GrapheneOS, enabling security features in defconfig.

These are two difference things, needs an `and`.

dngray:
> These are two difference things, needs an `and`.

Done.

PR #392, opened 2021-11-27
Add non Javascript frontends for video streaming

tommytran732:
Added Invidious since it works great without JS.

I thought Piped worked without JS, but this does not seem to be the case.

image


tommytran732:
Ready for review and merge,

hook9 (changes requested):
The privacy benefits of Invidious (and similar frontends) is highly dependent on how trustworthy the instance operator is, so I personally don't like recommending lists of instances that are relatively easy to get onto. It's important that users evaluate how much they trust an instance operator, otherwise there's little to no benefit and they're actually potentially hindering their privacy.

dngray:
> The privacy benefits of Invidious (and similar frontends) is highly dependent on how trustworthy the instance operator is, so I personally don't like recommending lists of instances that are relatively easy to get onto. It's important that users evaluate how much they trust an instance operator, otherwise there's little to no benefit and they're actually potentially hindering their privacy.

And with searx instances the same applies, there just is no way to really know. One could argue though if you were looking for a video on YT about herpes, using a .onion instance with `&local=true` least an instance operator would only know someone on Tor looked at a video on herpes vs Google potentially knowing exactly if a person visited YT directly with their logged in Google account.

freddy-m:
Would love it if you could open GitHub discussions/issues before making these PRs. 👍🏼

tommytran732:
> Would love it if you could open GitHub discussions/issues before making these PRs. 👍🏼

There was a discussion for it in another PR https://github.com/privacyguides/privacyguides.org/pull/356

dngray:
> Would love it if you could open GitHub discussions/issues before making these PRs. 👍🏼

Normally I'd say that's the case, but there is a very strong case for Invidious in that PR, it's also something we kinda indirectly mention anyway with Freetube.

dngray:
I'm thinking this might be better organized if the frontends/clients were split from LBRY.

LBRY is a decentralized distribution network that works through Tor, whereas the front ends/clients are just front ends/clients to YT.

I don't think we'll be re-adding PeerTube as there's nothing private about that. It doesn't support .onion services https://github.com/Chocobozzz/PeerTube/issues/1569 and enforces https https://github.com/Chocobozzz/PeerTube/issues/2899 which isn't desirable for a .onion service operators.

tommytran732:
> I'm thinking this might be better organized if the frontends/clients were split from LBRY.
> 
> LBRY is a decentralized distribution network that works through Tor, whereas the front ends/clients are just front ends/clients to YT.
> 
> I don't think we'll be re-adding PeerTube right now as there's nothing private about that. It doesn't support .onion services [Chocobozzz/PeerTube#1569](https://github.com/Chocobozzz/PeerTube/issues/1569) and enforces https [Chocobozzz/PeerTube#2899](https://github.com/Chocobozzz/PeerTube/issues/2899) which isn't desirable for a .onion service operators. It also can't be used without JavaScript.

I don't think we should ever recommend anything as a network. We should just stick to clients (desktop/mobile) and Non-JS front ends. This should also be applied to social media and what not too once we get to it (Nitter as a Non-JS front end for Tiwitter, and Fritter as a client for example)

dngray:
> I don't think we should ever recommend anything as a network. We should just stick to clients (desktop/mobile) and Non-JS front ends. This should also be applied to social media and what not too once we get to it (Nitter as a Non-JS front end for Tiwitter, and Fritter as a client for example)

I'm inclined to agree with direction, after all what we really care about is privacy. Getting the most out of the tools that give us this is really what our mission is.

dngray:
> The privacy benefits of Invidious (and similar frontends) is highly dependent on how trustworthy the instance operator is, so I personally don't like recommending lists of instances that are relatively easy to get onto. It's important that users evaluate how much they trust an instance operator, otherwise there's little to no benefit and they're actually potentially hindering their privacy.

I think we've added a suitable warning there. It's also worth noting that Invidious has a criteria for officially listed instances.

hook9 (approved):

PR #442, opened 2021-12-04
Typo

Positron832:
## Description
Fixed another typo.
Resolves: more eye pain?


dngray:
> more eye pain?

I'll be blind at this rate!

PR #446, opened 2021-12-07
Create virtual-insanity

freddy-m:
## Description

Blog post.


tommytran732:
Hm... I think you should update the main page and the blog page as well.

freddy-m:
Yeah, that'd probably be a good idea lol

PR #448, opened 2021-12-08
Update Providers, Software root pages

elitejake:
## Description

Updated list items and icons in [`software.html`](https://github.com/privacyguides/privacyguides.org/blob/main/legacy_pages/software.html) and [`providers.html`](https://github.com/privacyguides/privacyguides.org/blob/main/legacy_pages/providers.html).


dngray:
This one must have been missed

PR #456, opened 2021-12-11
Brave

tommytran732:
Discussion: https://github.com/privacyguides/privacyguides.org/discussions/453


freddy-m:
I don't think that the linked discussion was conclusive. Are we sure we want to list Brave @privacyguides/team?

tommytran732:
> Closing this for now, as the team sees little reason to list Brave presently.

Hi,
I am still discussing this with dngray privately and we are planning to expend this later. The arguments for Brave include:
1. Firefox ETP + uBO is not enough to protect the user. The user should use Firefox + Arkenfox or it is pretty much futile.
2. Arkenfox is a pain to setup and use for some users, and some will likely just use ETP + uBO (which again, is insufficient).
3. For those users, Brave is a much better choice.

tommytran732:
Could you please reopen the PR?

dngray:
I've re-opened it (not merging into main though). It's a draft PR still. This should allow you to work on it, we may/may not merge into main at a later date.

rnsjunkmail:
If privacy guides wants to have credibility, then all browsers with 10 million + users need to be addressed. Tests users can run to verify privacy efficacy of browsers need to be included as well as the rationale used to choose, rank,  and rate browsers needs to be clearly related.

PR #192, opened 2021-10-13
Fix #57 : Overhaul real-time communications page + add Session messenger

lrq3000:
## Description

Resolves: #57, #138, #50

* Add audits systematically whenever available.
* Separate descriptions of networks types in a separate tutorial section
* Tag by network types instead of sections only for instant messengers)
* Add illustrations of networks types.
* Rework a bit the definitions and pros/cons.
* Add Anonymous Routing as a network type.
* Add Session (ex Loki Messenger)
* Upgrade Status messenger (ex status.im messenger) into a recommendation (needs more testing to confirm).

Following discussions below:
* Remove non audited or non E2EE apps: Jami, Mumble.
* Add lots of warning as notes for almost all software messengers.
* Migrate template from legacy to evergreen
* Add labels to evergreen template (more flexible alternative to info and warning items, necessary to implement the tags based classification of messengers by networks types)

Todo before merge:
- [ ] Check if Status is worthy of being a recommendation (usability, security concerns since after the 2019 audit, etc).
- [ ] Consider Jami as a videoconferencing recommendation (need to confirm how it E2EE encrypts group calls).

This is a proposition, to overhaul the real-time communications page.

Currently, instant messengers are separated in different categories. This forces to dichotomize messengers into one category or the other, which doesn't fit well for some messengers that implement multiple networks (eg, Briar which is both P2P and Anonymous Routing). Furthermore, only instant messengers are identified by their network type, but this should also apply to video and voice messengers as well as teamchat.

I propose here following the discussion in #57 to redesign the page by removing the network type sections, and instead just present a list of instant messengers, but with tags in each description instead of sections. This allows for a more streamlined overview, similar to other listings in other pages, while not losing any information. In fact, there are more infos now since we can assign multiple networks types.

The layman descriptions of the differences between each network type is not lost either, but moved at the bottom of the page. It could be moved at the top too depending on what you want to present first (the selection of apps or a tutorial to help the user choose). Illustrations have been added, as well as an Anonymous Routing network type.

Note: the PR may be buggy (I will check with Netlify, build toolchain fails on my computer for some reason...). I will update this post when fully ready.

/EDIT: somewhat ready now to discuss further. Probably needs some more polishing if this proposition is accepted.

/EDIT2: Note that [Jami has not yet been audited](https://dt.gl/messnger-provider-review-2021-jami/).

/EDIT3: for future reference, all icons that can be used for tags are listed here: https://www.angularjswiki.com/fontawesome/

PR #488, opened 2021-12-21
Update Jitsi Meet instances link

elitejake:
## Description

Jitsi Meet's wiki has been migrated to [jitsi.github.io/handbook](https://jitsi.github.io/handbook/).<sup>[ref]</sup>

[ref]: https://github.com/jitsi/jitsi-meet/wiki


ph00lt0:
resolved in: https://github.com/privacyguides/privacyguides.org/pull/192

tommytran732:
Ah yeah we are not recommending Jitsi Meet anymore since it's end to end encryption support for group call is in alpha and it hasn't been audited yet, so I am closing this.

ph00lt0:
@tommytran732 although i agree that it would be important to do so, do we have any other recommendations for conference calling? A feature I really like is that I can give people the Jitsi meet phone number and meeting pin so that they call over phone as they request/demand without having to set up yet another temporary phone number.

ph00lt0:
> @tommytran732 although i agree that it would be important to do so, do we have any other recommendations for conference calling? A feature I really like is that I can give people the Jitsi meet phone number and meeting pin so that they call over phone as they request/demand without having to set up yet another temporary phone number.

Maybe this would be more suited in a VOIP and conferencing call section, which was suggested earlier. Most of us need to do these type of normal and groups calls, so it is still worth to recommend the best options for these imo.

tommytran732:
> > @tommytran732 although i agree that it would be important to do so, do we have any other recommendations for conference calling? A feature I really like is that I can give people the Jitsi meet phone number and meeting pin so that they call over phone as they request/demand without having to set up yet another temporary phone number.
> 
> Maybe this would be more suited in a VOIP and conferencing call section, which was suggested earlier. Most of us need to do these type of normal and groups calls, so it is still worth to recommend the best options for these imo.

I think it is better to just not recommend anything at all than recommending questionable products. When Jitsi Meet is ready, we will add it back.

ph00lt0:
@tommytran732 i think that is pretty strong worded for an open source project that actually tries to do right. I get where you  are coming from but if the privacy community wouldn't recommend alternatives everyone would still be using Zoom. Same goes for VOIP, it isn't perfect, but better then calling with your own number. When no prefect solution is available it can still be beneficial to look into how to layer defenses and limit the amount of data you'd have to provide.
I don't think we should compare Jitsi to something like element, they serve a different purpose. I would urge to recommend the best options for everyday internet users, not only tools who provide the highest standard. 
For the same reason, we should be telling people it is better to use a different OS if your on windows, but still provide windows users with guidance on how to stop telemetry etc.

ph00lt0:
Moved this discussion here: https://github.com/privacyguides/privacyguides.org/discussions/498

PR #499, opened 2021-12-25
Update email-clients.html

LinuxSnapshot:
Remove proprietary software.

## Description
While good for privacy, it is "proprietary" software.


dngray:
The reason this was an exception was because there were no decent PGP clients for iOS.

iOS is proprietary software, so this makes no sense.

PR #500, opened 2021-12-25
Update operating-systems.html

LinuxSnapshot:
"Windows 10" to "Windows"

## Description

Windows 11 is already released. Change "Windows 10" to "Windows" accordingly.


dngray:
This whole page is going to be re-done, so changing that doesn't make it apply to windows 11, etc.

PR #517, opened 2021-12-30
Remove Worth Mentioning (Self Contained Networks)

freddy-m:
## Description

We're removing the 'Worth Mentioning' sections, and this one is particularly messy.


jonaharagon (approved):

tommytran732:
Just in case anyone is wondering:

The "self-contained" networks should be providing some privacy/anonymity benefits to really be recommended. They must be "ready for production" too.

Zeronet: Not anonymous 
Retroshare: Your IP is visible on the network, like Torrent
I2P Bote: Doesn't seem to be maintained at all, the last release was 2 years ago
GNUnet: Not ready for production according to their own FAQ
IPFS: Same deficiencies as a Torrent network + supercookie
Yggdrasil: Not anonymous

PR #529, opened 2022-01-01
Markdown conversion

jonaharagon:


dngray:
The only cards I wasn't able to completely convert to markdown were the browser, and mobile browser cards where we used lists without any bullets ie:

```html
  <ul style="list-style-type:none;padding-left:0;">
    <li>xxxx</li>
  </ul>
```

The other thing was the Firefox IOS card where we used:

```html
<img alt="iOS Garbage Bin icon" width="16" height="16" src="/assets/img/browsers/ios-trash.svg"/>
```

dngray (approved):
Hopefully we can get this merged soon, so that we can prevent further diverging changes needing to be updated here and elsewhere.

PR #530, opened 2022-01-02
add zulucrypt

ail6b:
https://github.com/mhogomchungu/zuluCrypt
description based on the github project description

## Description

Resolves: no issue opened 

I am adding a software recommendation: zulucrypt. It is a tool for creating encrypted containers. The website linked is the github repo https://github.com/mhogomchungu/zuluCrypt. It also seems to have an official website https://mhogomchungu.github.io/zuluCrypt/, which is easy to find from the github repo. The tool is similar to tomb (which is already recommended), but it does more from what I can see. It definitely seems like a tool more suited for advanced users, but I see no harm in including an extra recommendation.

You might also be interested on the whonix wiki's opinion on it:
https://www.whonix.org/wiki/Full_Disk_Encryption#Zulucrypt

I have no conflict of interest.

### DIFF
--- a/_includes/legacy/sections/file-encryption.html
+++ b/_includes/legacy/sections/file-encryption.html
@@ -43,4 +43,5 @@
   <li><a href="https://www.dyne.org/software/tomb/">Tomb</a> - A simple zsh script for making LUKS containers on the commandline.</li>
   <li><a href="https://hat.sh/">Hat.sh</a> - A web application that provides secure client-side file encryption in your browser. It can also be selfhosted.</li>
   <li><a href="https://www.kryptor.co.uk/">Kryptor</a> - Free and open source file encryption software for Windows, macOS, and Linux.</li>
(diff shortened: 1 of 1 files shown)


dngray (changes requested):
Likely we won't do this as worth mentioning, but instead will have different categories, such as "OFTF volume encryption" and "file encryption".

ail6b:
To add to the discussion, if there's a new file encryption page, perhaps sirikali (same author as zulucrypt) could be also considered? Maybe the comparison at https://www.cryfs.org/comparison could be also a good food for thought for the discussion on the inclusion/exclusion of any of the tools and/or further recommendations for valid and invalid use cases for each of them.

ail6b:
furthermore could be relevant to creating LUKS partitions for containers of encrypted data
https://sockpuppet.org/blog/2014/04/30/you-dont-want-xts/

freddy-m:
Closing this until we a relevant discussion is opened and a conclusion is found.

Issue #532, opened 2022-01-02
Website Issue | No contrast between sections on index


RandPotato:
## Description
After the deletion of legacy pages all sections on the main page (except the first) are white, leaving no visual difference or contrast between them

# Screenshots
![image](https://user-images.githubusercontent.com/82942910/147867417-0e7b85cb-ebe2-4dfc-b280-47398f050f8f.png)

PR #533, opened 2022-01-02
Fix contrast between sections on main page

RandPotato:
## Description

Resolves: #532


dngray:
Not needed anymore https://github.com/privacyguides/privacyguides.org/pull/829

PR #548, opened 2022-01-10
Update CalyxOS comparison

tommytran732:
CalyxOS Android 12 is in beta. This PR just removes outdated information about them being on Android 11. This should be merged when CalyxOS Android 12 hits the stable channel.


stepech:
Hey there,
it appears that you are removing whole text category, related only partially to Android 12. In case of GrapheneOS  `Device Patch Level` contains information also about what devices are supported and how long. The same can go to CatalyxOS, instead of removing the category completely just because it mentions unsupported Android 12, wouldn't it be better to just update the text, maybe include same information/supported device list link, as in GrapheneOS?
You are also updating paragraph mentioning Aurora store and their PR#153, which was already completed, might be worth looking into too as the info is thus obsolete (Not necessarily in this PR).

tommytran732:
> Hey there,
> it appears that you are removing whole text category, related only partially to Android 12. In case of GrapheneOS `Device Patch Level` contains information also about what devices are supported and how long. The same can go to CatalyxOS, instead of removing the category completely just because it mentions unsupported Android 12, wouldn't it be better to just update the text, maybe include same information/supported device list link, as in GrapheneOS?
> You are also updating paragraph mentioning Aurora store and their PR#153, which was already completed, might be worth looking into too as the info is thus obsolete (Not necessarily in this PR).

1. That link about which device is supported for how long is already mentioned in other places on the Android page.
The patch level comparison is no longer applicable after CalyxOS catches up, so I don't see any reason to leave anything there.

2. The Aurora one will be mentioned when it is actually released in the stable version of Aurora Store. It is merged, but it is not in the stable version yet. The stable version was out like 6 months ago I think?

stepech:
Then maybe remove the heading too? (Line 156) This doesn't look right. But otherwise :+1:

image

tommytran732:
> Then maybe remove the heading too? (Line 156) This doesn't look right. But otherwise 👍
> 
> image

Fixed, ty

SkewedZeppelin:
Maybe also add a note that CalyxOS is currently shipping a version of Chromium from October:
https://chromereleases.googleblog.com/2021/10/chrome-for-android-update_01703513036.html

https://omahaproxy.appspot.com/

latest stable is 97.0.4692.98
GrapheneOS Vanadium is at 97.0.4692.98
DivestOS Mulch is at 97.0.4692.98
CalyxOS is at 95.0.4638.50
LineageOS is at 96.0.4664.104 with 97.0.4692.87 pending
/e/OS is at 96.0.4664.54
Bromite is at 96.0.4664.54

tommytran732:
I wonder what version they have on their testing channel... this is bad...

tommytran732:
Still looks like 95 actually

SkewedZeppelin:
> this is bad...

https://divestos.org/misc/ch-dates.txt
:rofl: :rofl: :rofl: :rofl:

flawedworld:
You may wish to refer to https://chromiumdash.appspot.com/releases?platform=Android for a clear overview of Chromium release history.

CalyxOS's chromium build in their 12 release is currently affected by over 100 CVE's, and they are 92 days behind on patches.

CVE-2021-4102 is currently being exploited in the wild (fixed in 96.0.4664.104 onwards).

SkewedZeppelin:
I calculated it at ~121 known security issues.

flawedworld:
> I calculated it at ~121 known security issues.

And that's just the CVEs.... theres likely tons more which were found internally by fuzzing etc and fixed

tommytran732:
Yeah I made a PR for a warning. This is really raelly bad. Let's hope it doesn't take them a few more months to fix it.

PR #578, opened 2022-01-23
Add a warning for CalyxOS

tommytran732:
I don't want to completely remove CalyxOS just yet, but we really cannot recommend it in its current state. The firmware patch level is 4 months out of date on both the stable and beta branch (yes, I am aware that it is fixed on the testing branch), and Chromium browser & webview are completely outdated on all branches. https://github.com/privacyguides/privacyguides.org/pull/548#issuecomment-1018245074. 

I'd like to give them the benefit of the doubt and wait a bit longer to see how things work out, but in its current state, I don't think Calyx is a good recommendation anymore.


DeiAsPie:
I say wait for the first week of Feb. If still things don't get well then do as you like.
They are currently dealing with a major migration of all devices to Android 12 so pardon them for being not so updated for a while.

dngray:
> I say wait for the first week of Feb. If still things don't get well then do as you like.

We can place a warning and remove it later, although if it goes on too much longer we might not recommend it at all.

SkewedZeppelin:
They've recently updated... to 97: https://review.calyxos.org/c/CalyxOS/platform_prebuilts_calyx_chromium_arm64/+/8654

tommytran732:
Warning removed. Thank you for notifying us :)

Issue #553, opened 2022-01-11
VPN section: saying Bitcoin is an anonymous payment option


userdaphi:
## the site is saying that Bitcoin (BTC) is an anonymous payment option. But this is not true. The ONLY privacyfriendly cryptocurrency is Monero (XMR). It is the only cryptocurrency which is private-by-design.

I give you an example:
IVPN
> Accepts Bitcoin
In addition to accepting credit/debit cards and PayPal, IVPN accepts Bitcoin, Monero and cash/local currency (on annual plans) as anonymous forms of payment.

People might think that Bitcoin is a private payment option. This is not true.

To include Bitcoin as a requirement for a private VPN provider is simply nonsense. It would be better to include Monero.
I would also suggest an explanation for that fact, why Bitcoin or any other cryptocoin is not private at all (except Monero of course)
**https://www.privacyguides.org/providers/vpn/**

m77e4t:
Claiming 'only' Monero might be too much. There are other privacy coins like XMR which work similar to it or use other cryptographic methods to make them equal or more anonymous/private than XMR. https://www.coingecko.com/en/categories/privacy-coins

Still, the info regarding payment on VPN page as of 2/3 Feb 2022 seems pretty good.

samuel-lucas6:
> There are other privacy coins like XMR 

Monero is XMR. The only other privacy coin I've heard of is Zcash, not that I'm following cryptocurrencies. If I remember correctly, some Zcash designers made some claims about potentially helping the police, which made people scared that it had a backdoor.

m77e4t:
My bad. I made a mistake while writing the comment. I wanted to list a few privacy coins, but I accidentally wrote 'XMR' where it should not have been written (possibly I was doing some other stuff while writing the comment). Furthermore, I have corrected and updated my comment.

dngray:
> People might think that Bitcoin is a private payment option. This is not true.

I've been thinking about this one. Now that we've launched the new site, I wouldn't mind a page that discusses cryptocurrencies in regard to their application of privacy (and not).

We've discussed it briefly within the team rooms. There is quite a lot of misconception about Bitcoin and how it relates to privacy, particularly the fact that it **is not private** and there is a **public ledger**.

It'd be an entry-level article, that discusses these aspects.

dngray:
Merged in https://github.com/privacyguides/privacyguides.org/commit/dc94c77dadc72b4f09e0488ea4240a45bf8d4fee

Issue #573, opened 2022-01-22
Website Issue | HTML tags on meta description


1hiking:
## Description

I am pretty sure html tags shouldn't be put inside meta description tags. For example on the Android page, there is hyperlink tags.

![image](https://user-images.githubusercontent.com/73438877/150652799-ed8c4c99-32d1-4321-b59d-77b067e37f76.png)

DeiAsPie:
I don't anything wrong here 🤷

Issue #545, opened 2022-01-07
Mailvelope is not a desktop Email client


1hiking:
## Description

This tool is not an email client but a browser addon for a web-client

**URL of affected page:**
https://privacyguides.org/software/email/#desktopclients

Loki-L1130:
I agree with you.

freddy-m:
We wanted to add it to the section, but we weren't really sure where to put it. Do you have any suggestions @1hiking? cc @privacyguides/team

Loki-L1130:
How about keeping it under Email Providers ?

dngray:
I think with this one, we want to split that page up in the legacy revamp, essentially allowing for native clients per platform, we could then simply have a web section.

dngray:
I think i'll leave this one under desktop clients. Technically it is, as it only works on desktop browsers.

dngray:
Fixed in https://github.com/privacyguides/privacyguides.org/pull/829, it only works on desktop browsers.

Issue #538, opened 2022-01-04
LBRY F-Droid


adam-stambaugh:
The site states:

> The Odysee website and the LBRY clients in F-Droid, Play Store and the AppStore have mandatory synchronization and telemetry.

That isn't true as far as I can tell for the F-Droid branch. It was made compliant to not have anti-features on F-Droid as mentioned here:

https://github.com/lbryio/lbry-android/issues/1047

And setup a separate branch:

https://github.com/lbryio/lbry-fdroid

Even F-Droid says there aren't any anti-features. Looking in the options, anything tracking related is disabled by default. It also doesn't require an account, same as the desktop client.

https://privacyguides.org/video-streaming/

PR #579, opened 2022-01-24
Fix various GH issues

ElijahPepe:
## Description

Resolves: #553, #545, #573, #538, #531, #424, #353, #159, #128, #81

The commits detail what I've fixed, but I've tried to fix a ton of the issues on GitHub Issues.


DeiAsPie:
The highlighted one is fine but the ones not currently highlighted are barely readable. Maybe attempt some other way of showing something is highlighted like changing colors themselves rather than using different shades of the same color.

dngray:
While we appreciate the contributions, this would have really better been split up into individual PRs for each issue.

Makes it difficult to review, and possibly reverse if there is some issue.

PR #588, opened 2022-01-31
Add NewPipe x SponsorBlock

tommytran732:


freddy-m:
Not sure if this should be on the site @privacyguides/team? Seems beyond the scope...

tommytran732:
This fork provides no privacy benefit over Newpipe, but it does'nt make anything worse either, and it seems well-maintained. It does provide a feature that a lotnof people apprefiate, so i thought "Why not?".

It is a lot better than having people install YouTube Vanced if all they want is SponsorBlock.

elitejake:
> but it doesn't make anything worse either,

Doesn't it? I don't think NewPipe with SponsorBlock store offline databases, instead it uses this ([sponsor.ajay.app](https://sponsor.ajay.app/)) public API right?

You should add the [privacy policy](https://gist.github.com/ajayyy/aa9f8ded2b573d4f73a3ffa0ef74f796) of SponsorBlock in the section too.

tommytran732:
> > but it doesn't make anything worse either,
> 
> Doesn't it? I don't think NewPipe with SponsorBlock store offline databases, instead it uses this ([sponsor.ajay.app](https://sponsor.ajay.app/)) public API right?
> 
> You should add the [privacy policy](https://gist.github.com/ajayyy/aa9f8ded2b573d4f73a3ffa0ef74f796) of SponsorBlock in the section too.

SponsorBlock hases the URL, send the first few characters to SponsorBlock, retrieve the list, then search for the exact hash locally. There is not really a privacy issue imo.

Besides, this fork doesn't even have a default sponsorblock instance. If we were to add sponsorblock's TOS, we'd be adding it to FreeTube (which does use sponsor.ajay.app by default), not Newpipe x Sponsorblock.

PR #614, opened 2022-02-12
changing legacy html pages to markdown

DeiAsPie:


tommytran732:
This is already done.

DeiAsPie:
But I still see html files in legacy_pages

PR #634, opened 2022-02-21
Add Piped to Video Streaming

mfwmyfacewhen:
Added Piped to the Video Streaming section.


dngray:
Main reason is there are things Piped does which Invidious does not.

Invidious may be better for privacy (html only, supports .onion instances) however, it doesn't support high resolutions https://github.com/iv-org/invidious/issues/34 of course this could change in the future.

ph00lt0:
@dngray alright all good, thanks for explaining!

PR #642, opened 2022-02-22
Fix error in mobile browser recommendations

dngray:


SkewedZeppelin:
Can "site isolation" be clarified that it is "per-site process isolation" and not to do with data?
I've read many a comments of people thinking it means sites can read your cookies of other sites due to this ambiguity.

PR #644, opened 2022-02-22
Add information about F-Droid

tommytran732:


akumeilol:
> Some applications like [IVPN](https://www.ivpn.net/) do not include some features (adblocking in IVPN’s case) in their F-Droid build out of fear of censorship by Google. Evaluate whether the additional features in the F-Droid build is worth the slower update...

This doesn't make sense to me. IVPN supposedly doesn't include adblocking in the F-Droid build yet we are told to evaluate the additional features in the F-droid build? Did you mean they withhold the feature from the Google Play Store?

PR #671, opened 2022-03-02
Fix author frontmatter in Firefox 2021 blog post

ghost:
Currently the author for the Firefox 2021 blog post shows up as `["Daniel Gray"]` in the RSS feed—this PR aims to fix that by changing it to `Daniel Gray`.

image

Apologies for opening a PR for something not really in *need* of a fix, but I'm required to submit a number of PRs in public repositories for my computer science class. I figured I might as well contribute to a project that I believe in :P

### DIFF
--- a/collections/_posts/2021-12-01-firefox-privacy-2021-update.md
+++ b/collections/_posts/2021-12-01-firefox-privacy-2021-update.md
@@ -1,7 +1,6 @@
 ---
 title: 'Firefox Privacy: 2021 update'
-author:
-  - Daniel Gray
+author: Daniel Gray
 cover: firefox-privacy-2021.png
 color: "linear-gradient(to bottom, rgba(158, 82, 19, 0.5), rgba(158, 82, 19, 0.9))"
 layout: post


arcticlex:
Jekyll feed is what's generating the RSS feed, as is stated at the top of [https://privacyguides.org/feed.xml](https://privacyguides.org/feed.xml)

I can think of a couple possible solutions to allowing for multiple authors.
1. Generate a JSON feed instead, using the 'jekyll-json-feed' gem, since as you stated it parses correctly into JSON.
2. Modify the way the jekyll-feed plugin outputs, using something like [this](https://gist.github.com/arcticlex/b0f61248e5fb6c5f440a46c325cbd1b4)
3. Take the lazy approach and just list all the authors as a single author, like "Daniel and Freddy"

I will try 1 & 2 tomorrow.

I also prematurely readied this PR as the 'Welcome to Privacy Guides' blog post still has this author display issue. But I guess that blog post will be a good testing ground to fixing this.

dngray:
> Jekyll feed is what's generating the RSS feed, as is stated at the top of https://privacyguides.org/feed.xml

Hmm, I wasn't aware we had that, admittedly I hadn't looked too hard. We should probably add:
```html
<link rel="alternate" type="application/rss+xml" 
  title="Blog | Privacy Guides" 
  href="https://privacyguides.org/feed.xml" />
```

> 1. Generate a JSON feed instead, using the 'jekyll-json-feed' gem, since as you stated it parses correctly into JSON.

No reason we couldn't offer that too. https://stevenwestmoreland.com/2017/05/adding-json-feed-to-your-jekyll-site.html seems to describe how to do that.

> 2. Modify the way the jekyll-feed plugin outputs, using something like [this](https://gist.github.com/arcticlex/b0f61248e5fb6c5f440a46c325cbd1b4)

I'm thinking maybe we should look at https://github.com/jekyll/jekyll-feed#author-information

> 3. Take the lazy approach and just list all the authors as a single author, like "Daniel and Freddy"

Nah, I'd rather not do that.

arcticlex:
I've spent some time reading all the Jekyll author information I can find but it still doesn't make sense to me, and I'm not sure I'm whiling to dedicate enough time to figuring it out. It all seems very unintuitive and unnecessarily complex, especially for a site like Privacy Guides where the blog is not the main focus.

However, I think I would be able to make it work using a custom generator instead of Jekyll feed. Something like https://www.johnvincent.io/jekyll/rss-feed-with-jekyll/, which is a similar approach to the JSON generator that you linked to.

I tried building the PG site myself, but I couldn't get it to work after quite a bit of trial and error. Would it be okay if I create a draft PR and make extensive use of the Netlify bot that builds the previews? I had to use Netlify in a web dev course and I remember there being limits to the number of builds, so I would hate to use up your guys' quota fixing quite a trivial issue.

PR #680, opened 2022-03-07
Adding Advanced Protection Program

tommytran732:
Adding this to the Android page for now, as it does affect stock Android. When we discuss ChromeOS/ChromeOS Flex, or other Google products, we could consider making a page dedicated to recommendations with the Google Account and linking to that page instead.

### DIFF
--- a/collections/_evergreen/android.md
+++ b/collections/_evergreen/android.md
@@ -148,7 +148,23 @@ We have these general tips:
 * Check if an app is available on the [IzzyOnDroid](https://apt.izzysoft.de/fdroid/) repository. The IzzyOnDroid repository pulls builds directly from GitHub and is the next best thing to the developers' own repositories. We recommend that you download the GitHub builds and install them manually first, then use IzzyOnDroid for any subsequent updates. This will ensure that the signature of the applications you get from IzzyOnDroid matches that of the developer and the packages have not been tampered with.
 * Check if there are any differences between the F-Droid version and the Google Play Store version. Some applications like [IVPN](https://www.ivpn.net/) do not include certain features (eg [AntiTracker](https://www.ivpn.net/knowledgebase/general/antitracker-faq/)) in their Google Play Store build out of fear of censorship by Google.
 
-Evaluate whether the additional features in the F-Droid build is worth the slower updates or whether faster updates from the Google Play is worth the potential privacy issues in your [threat model](/threat-modeling/).
+Evaluate whether the additional features in the F-Droid build are worth the slower updates. Also think about whether faster updates from the Google Play Store are worth the potential privacy isues in your [threat model](/threat-modeling/).
+
+## Advanced Protection Program
+If you have a Google account we suggest enrolling in the [Advanced Protection Program](https://landing.google.com/advancedprotection/). It is available at no cost to anyone with two or more hardware security keys with [U2F](https://en.wikipedia.org/wiki/Universal_2nd_Factor) support.
+
+The Advanced Protection Program provides enhanced threat monitoring and enables:
+
+* Stricter two factor authentication; e.g. that [U2F](https://en.wikipedia.org/wiki/Universal_2nd_Factor) **must** be used and disallows the use of [SMS OTPs](https://en.wikipedia.org/wiki/One-time_password#SMS), [TOTP](https://en.wikipedia.org/wiki/Time-based_one-time_password), and [OAuth](https://en.wikipedia.org/wiki/OAuth).
+* Only Google and verified third party apps can access account data
+* Scanning of incoming emails on Gmail accounts for [phishing](https://en.wikipedia.org/wiki/Phishing#Email_phishing) attempts
+* Stricter [safe browser scanning](https://www.google.com/chrome/privacy/whitepaper.html#malware) with Google Chrome
+* Stricter recovery process for accounts with lost credentials
+
+For users that are using the privileged Play Services (common on stock operating systems), the Advanced Protection Program also comes with [additional benefits](https://support.google.com/accounts/answer/9764949?hl=en) such as:
+* Not allowing app installation outside of the Google Play Store, the OS vendor's app store, or via [`adb`](https://en.wikipedia.org/wiki/Android_Debug_Bridge)
+* Mandatory automatic device scanning with [Play Protect](https://support.google.com/googleplay/answer/2812853?hl=en#zippy=%2Chow-malware-protection-works%2Chow-privacy-alerts-work)
+* Warning the user about unverified applications
 
 ## Security comparison of GrapheneOS and CalyxOS
 


collections/_evergreen/android.md
```diff
@@ -150,6 +150,20 @@ We have these general tips:
 
 Evaluate whether the additional features in the F-Droid build is worth the slower updates or whether faster updates from the Google Play is worth the potential privacy issues in your [threat model](/threat-modeling/).
```
mfwmyfacewhen, line 151: Maybe split this up into two sentences, like "Evaluate whether the additional features in the F-Droid build are worth the slower updates. Also think about whether faster updates from the Google Play Store are worth the potential privacy isues in your threat model."

collections/_evergreen/android.md
```diff
@@ -150,6 +150,20 @@ We have these general tips:
...
 
+## Advanced Protection Program
+We highly recommend that you enroll in the [Advanced Protection Program](https://landing.google.com/advancedprotection/) if you have a Google account. It is free and available to anyone with 2 or more hardware security keys with U2F support. The program puts your account under enhanced threat monitoring and has security benefits such as:
+
+* Enforcing that U2F be used for two factor authentication. SMS, TOTP, and OATH are not allowed.
```
mfwmyfacewhen, line 156: Either remove the periods at the end of each bullet point or add them for consistency

mfwmyfacewhen (changes requested):

collections/_evergreen/android.md
```diff
@@ -150,6 +150,20 @@ We have these general tips:
...
+* Enforcing that only Google apps and verified third party apps can access your account data.
+* Enhanced scanning of incoming emails on GMail accounts for phishing attempts
+* More aggressive safe browsing scanning with Google Chrome
+* Stricter process for account recovery
+
+For users that are using the privileged Play Services (common on stock operating systems), the Advanced Protection Program also comes with [additional benefits](https://support.google.com/accounts/answer/9764949?hl=en) such as:
+* Disallowing app installation outside of the Google Play Store, the OS vendor's app store, and adb
+* Mandatory automatic device scanning with [Play Protect](https://support.google.com/googleplay/answer/2812853?hl=en#zippy=%2Chow-malware-protection-works%2Chow-privacy-alerts-work)
```
ph00lt0, line 164: Although I like the advanced protection program, I really wouldn't encourage users who want privacy to enable this. That seems counter productive. 
Google will receive:
 - Your device's network connections 
 - Potentially harmful URLs 
 - Operating system, and apps installed on your device through Google Play or other sources.

tommytran732: Doesn't matter, as Google already have all of that if you are using privileged Play Services. You do not get to choose whether to install the Play Services or not, the OS vendor does it for you.

If you are not using the privileged play service, Google cannot block app install or auto remove malicious apps, because it doesn't have the permission to do so.

ph00lt0: Okay i see where your coming from. I agree using a different OS would be a much better idea, but saying that it doesn't make a difference is not true AFAIK.

It seems your implying that scanning isn't disabled when the users does disable play protect? I know it's enabled by default (at least it always was) but when the user disables the scanning from what I understand it stops the framework from reporting installed applications. The applications you use are often quite a unique pattern to identify users, as network activity can be as well.

Security wise I would fully agree with you, privacy however I am not sure if this is the best advice.

tommytran732: > Okay i see where you're coming from. I agree using a different OS would be a much better idea, but saying that it doesn't make a difference is not true AFAIK.
> 
> It seems your implying that scanning isn't disabled when the users does disable play protect? I know it's enabled by default (at least it always was) but when the user disables the scanning from what I understand it stops the framework from reporting installed applications. The applications you use are often quite a unique pattern to identify users, as network activity can be as well.
> 
> Security wise I would fully agree with you, privacy however I am not sure if this is the best advice.

An app can check what other apps you have installed, privileged or not. In fact, yes, even with Play Protect off, Google still scans your installed app anyways. How do you think it does app updates? It has to know what you have installed, what version you have installed, then compare it to the latest version of those apps and pull the latest updates.

ph00lt0: @tommytran732 you are probably right, didn't really think of that. Was hoping the data wouldn't be used in that way. Same goes for safe browsing, that also is a complex recommendation if your asking me.

tommytran732: > @tommytran732 you are probably right, didn't really think of that. Was hoping the data wouldn't be used in that way. Same goes for safe browsing, that also is a complex recommendation if your asking me.

Safe Browsing is more nuanced than Play Services scanning.

In Chrome and various other forks, the data is sent directly through Google. Its not so great, but if you are using it anyways, you might as well want it to be aggressive like on the advanced protection program. If you don't want it, use a browser that lets you disable it (and iirc it can be disabled in Chrome as well).

If you are using Brave, then they proxy the requests through their server before sending it to Google. This is bette, but not idea.

With Firefox + additional configuration, you can make it do Safe Browsing checks locally only, which would be the best implementation of it.

collections/_evergreen/android.md
```diff
@@ -150,6 +150,22 @@ We have these general tips:
 
 Evaluate whether the additional features in the F-Droid build is worth the slower updates or whether faster updates from the Google Play is worth the potential privacy issues in your [threat model](/threat-modeling/).
 
+## Advanced Protection Program
+If you have a Google account we suggest enrolling in the [Advanced Protection Program](https://landing.google.com/advancedprotection/). It is available at no cost to anyone with two or more hardware security keys with [U2F](https://en.wikipedia.org/wiki/Universal_2nd_Factor) support.
+
+The Advanced Protection Program provides enhanced threat monitoring and enables:
+
+* Stricter two factor authentication; e.g. that [U2F](https://en.wikipedia.org/wiki/Universal_2nd_Factor) **must** be used and disallows the use of [SMS OTPs](https://en.wikipedia.org/wiki/One-time_password#SMS), [TOTP](https://en.wikipedia.org/wiki/Time-based_one-time_password), [OAuth](https://en.wikipedia.org/wiki/OAuth).
```
mfwmyfacewhen, line 158: add "and" before 0Auth

mfwmyfacewhen (changes requested):

collections/_evergreen/android.md
dngray: Agreed, I think that does sound a bit better.

collections/_evergreen/android.md
dngray: Done.

collections/_evergreen/android.md
dngray: Done.

PR #682, opened 2022-03-07
Adding Phone by Google

tommytran732:
Can't say no to automatic e2ee :)


_data/operating-systems/android-applications/4_phone-by-google.yml
```diff
@@ -0,0 +1,11 @@
+title: Phone by Google
+type: Recommendation
+logo: /assets/img/android/phone-by-google.png
+description: |
+  **Phone by Google** is a proprietary phone application with support for [end to end encryption](https://fi.google.com/about/end-to-end-encrypted-calls/) for [Google Fi](https://fi.google.com/about/) users, CallerID and spam blocking.
```
mfwmyfacewhen, line 5: change "end to end" to "end-to-end", add comma after CallerID

_data/operating-systems/android-applications/4_phone-by-google.yml
```diff
@@ -0,0 +1,11 @@
...
+
+  If you are a Google Fi customer, and your device meets the [requirements](https://support.google.com/fi/answer/11295314?hl=en), we highly recommend that you use this app to leverage the automatic end to end encryption feature.
```
mfwmyfacewhen, line 7: change "end to end" to "end-to-end"

mfwmyfacewhen (changes requested):

ph00lt0:
As many people said, the meta data is worth much more... I really do not agree with this listing and if you do you should also be listing WhatsApp, there is not different. Actually there is one, WhatsApp has a lot more users than Google Fi. So if you care about just the encrypted call you are likely better of using them. 

We should encourage people to use actually good encrypted solutions such as Signal not a google app that sends all about who you are contacting to them. That is quite the opposite of being something privacy friendly. 

So yes we can definitely say no to this, just like we do to WhatsApp.

ph00lt0:
The fact that Google is adding this only confirms that meta-data now much more valuable than actual data, as Snowden once stated.

tommytran732:
> As many people said, the meta data is worth much more... I really do not agree with this listing and if you do you should also be listing WhatsApp, there is no difference. Actually there is one, WhatsApp has a lot more users than Google Fi. So if you care about just the encrypted call you are likely better of using them.
> 
> We should encourage people to use actually good encrypted solutions such as Signal not a google app that sends all about who you are contacting to them. That is quite the opposite of being something privacy friendly.
> 
> So yes we can definitely say no to this, just like we do to WhatsApp.

Google is the Carrier, they already have the metadata. Not sure what your point is. 
WIth whatsapp, you have to go out of your way to sign up for it. It's not so great.
WIth the phone app, if you are already on Google Fi, it makes no sense to not use it. If your friends are on Signal, they should be contacting you via Signal. If they aren't, then the auto e2ee is awesome.

ph00lt0:
> > As many people said, the meta data is worth much more... I really do not agree with this listing and if you do you should also be listing WhatsApp, there is no difference. Actually there is one, WhatsApp has a lot more users than Google Fi. So if you care about just the encrypted call you are likely better of using them.
> > We should encourage people to use actually good encrypted solutions such as Signal not a google app that sends all about who you are contacting to them. That is quite the opposite of being something privacy friendly.
> > So yes we can definitely say no to this, just like we do to WhatsApp.
> 
> Google is the Carrier, they already have the metadata. Not sure what your point is. WIth whatsapp, you have to go out of your way to sign up for it. It's not so great. WIth the phone app, if you are already on Google Fi, it makes no sense to not use it. If your friends are on Signal, they should be contacting you via Signal. If they aren't, then the auto e2ee is awesome.

Well it will only work if you encourage others to user Google Fi. That doesn't seem like a good idea. I agree with you that it's great they are adding this, but in order to make use of this you would need to encourage friends family to get Google Fi, it would probably even be easier to convince someone to get Signal.

ph00lt0:
By adding this on the website, it looks like we recommend Google Fi as provider, I don't think any provider is super privacy friendly but it just doesn't seem like a good idea.

tommytran732:
> By adding this on the website, it looks like we recommend Google Fi as provider, I don't think any provider is super privacy friendly but it just doesn't seem like a good idea.

We are not officially endorsing Google Fi, but think about it for a moment. Your carrier is going to have your metadata anyways, so it's only a matter of trusting who your carriers are. Google is not so bad when it comes to this. Some carriers don't even have e2ee enabled for their accounts, others like T-Mobile gets hacked non-stop. Google at the very least is competent with security - you have the advanced protection program, sim swapping is impossible without having your Google account itself and whatnot.

Also, why would you ever not take opportunistic e2ee over no e2ee at all?

ph00lt0:
> > By adding this on the website, it looks like we recommend Google Fi as provider, I don't think any provider is super privacy friendly but it just doesn't seem like a good idea.
> 
> We are not officially endorsing Google Fi, but think about it for a moment. Your carrier is going to have your metadata anyways, so it's only a matter of trusting who your carriers are. Google is not so bad when it comes to this. Some carriers don't even have e2ee enabled for their accounts, others like T-Mobile gets hacked non-stop. Google at the very least is competent with security - you have the advanced protection program, sim swapping is impossible without having your Google account itself and whatnot.
> 
> Also, why would you ever not take opportunistic e2ee over no e2ee at all?

Oh yeah I am definitely not challenging that. (Certain) providers (especially outside EU) can be horrible when it comes to data and they surely do not have good reputations. I just would feel very uncomfortable letting Google combine yet another source of data to your identity. I would argue it may be a better idea to use a provider that doesn't connect your data to other parts of your digital life.  

I am not sure how big Google Fi is but what is the chance that you call with someone who does and uses android and uses this app? I mean that seems very low and hard to explain and ask for. Establishing a encrypted connection is at that point completely random, I would not want to rely on that. 

Also I am generally recommending people to use VOIP rather then sim/esim for calling that has a lot of advantages when it comes to both privacy and security.

dngray (changes requested):
This  one needs further discussion.

dngray:
We discussed this as a team, while there is some merit in the fact that you get e2ee, with this and [Google Messages](https://en.wikipedia.org/wiki/Messages_(Google)), it doesn't really need endorsement by PrivacyGuides over platform, carrier independent options like [Signal](https://en.wikipedia.org/wiki/Signal_(software)), which provide E2EE with a lot less metadata.

ph00lt0:
For reference: https://www.scss.tcd.ie/doug.leith/privacyofdialerandsmsapps.pdf

dngray:
> For reference: https://www.scss.tcd.ie/doug.leith/privacyofdialerandsmsapps.pdf

That paper comes to some pretty poor conclusions about "privacy". They talk about "anonymity" for example, and claim that Google messages is not anonymous, meanwhile recommending users install "generic" SMS applications without encryption.

Please explain to me how sending entire copies of communicated data unencrypted to your telephone carrier is "more private" than sending a hash to Google. I'm not by any means saying that Google Messages is the "best" tool out there, but it's not any worse than trusting notoriously untrustworthy operators.

I might add, the reason that Google invested in RCS, and Google Messaging, comes back to lazy operators not rolling this ITU standard out earlier, leaving users with nothing.

PR #690, opened 2022-03-08
PrivacyBlur

dngray:
Modern application that can redact photos.


mfwmyfacewhen (commented):
Looks good

lx-is (changes requested):
You might as well add the iOS link. Even though it's on an Android page, I don't see why you wouldn't :P

```
  - icon: fab fa-app-store-ios
     url: 'https://apps.apple.com/app/privacyblur/id1536274106'
```

There's also a higher resolution image for the logo. It's only 1KB more and looks a lot sharper.
privacyblur

dngray:
> You might as well add the iOS link. Even though it's on an Android page, I don't see why you wouldn't :P

I was thinking about that, but the plan is to have a dedicated iOS page and then this will feature there the same.

> There's also a higher resolution image for the logo. It's only 1KB more and looks a lot sharper.

Its simple enough I might make a SVG version lol.

lx-is:
I actually made a svg version, but it turned out to be about 20KB so I figured it wasn't worth it. Maybe you can do better though

dngray:
> I actually made a svg version, but it turned out to be about 20KB so I figured it wasn't worth it. Maybe you can do better though

Do you want to post it?

lx-is:
Sorry, I already deleted it 😞

lx-is:
output

here's a svg without the background, forgot the site has dark mode

dngray:
Any idea why neither of these show up in Inkscape?

lx-is:
Here's the non-optimized version. Does this show up in Inkscape?
input

dngray:
No it doesn't, weird.

lx-is:
That's weird. I've never used Inkscape so I'm not sure why that would be. I created these in Figma.

dngray:
> I recreated it in Sketch, which I probably should've used in the first place anyway. This is already optimized with Scour. Lemme know if it works

That one works :+1:

Tvirin:
Here is video explain for unredacted

https://www.youtube.com/watch?v=3fBhPP2FD-c

PR #692, opened 2022-03-09
Fix position of Mailcow

lx-is:
## Before
before

## After
after

---

Affected pages: https://privacyguides.org/providers/email/#selfhosting


lx-is:
> *positing*

lol that typo

PR #701, opened 2022-03-13
Remove firefox ios and focus, add safari

mfwmyfacewhen:
Removing Firefox iOS and Focus, replacing with Safari

All browsers on iOS [must use Apple's WebKit](https://developer.apple.com/app-store/review/guidelines/#2.5.6) to be allowed on the App Store. As a result, any browser on the app store is essentially just Safari under the hood. Safari also offers features like blocking cross-site tracking and extensions, which essentially replaces the functionality of Firefox iOS. Each private tab in Safari is isolated from every other tab, and when the user closes a private tab, all history and cookies related to that tab are deleted. This means each private tab in Safari replicates the functionality of Focus. These apps do not add anything that Safari doesn't already offer, so I'm replacing them with a Safari guide.

Related to this discussion: https://github.com/privacyguides/privacyguides.org/discussions/662


lx-is:
For the record, I still disagree with the removal of Firefox Focus 😛

mfwmyfacewhen:
> For the record, I still disagree with the removal of Firefox Focus stuck_out_tongue

Is there something focus does that safari doesn't? If so I'll keep it.

lx-is:
> > For the record, I still disagree with the removal of Firefox Focus stuck_out_tongue
> 
> Is there something focus does that safari doesn't? If so I'll keep it.

Custom search engines. And please don't say you can add a bookmark to a search engine on your home screen with Safari, that is the jankiest solution imaginable.

ph00lt0:
I also don't think this is a good idea. Firefox Focus offers in-build ad-blocking which is much easier then with Safari + something like Adguard.

lx-is:
Yup. I think it can be summarized as just "easy to use, easy to recommend."

ph00lt0:
This is currently a big topic in browsers in our community. Some people seem to think that blocking ads makes you less private. I really disagree with this. Blocking trackers and ads by default will block so many parties from collecting your data. Obviously there are also more sophisticated tracking technologies such as fingerprinting but if that is your thread you need to use Tor and probably Tails. Lets not create the idea that you can prevent this in general browsers by not recommending people to block as many as you can. Stopping trackers and ads from loading at all still benefits you the most as most of the internet still uses this. It's fairly easy to block all those nasty companies from loading their scripts in to your system, which is both great for privacy and security.

mfwmyfacewhen:
> This is currently a big topic in browsers in our community. Some people seem to think that blocking ads makes you less private. I really disagree with this. Blocking trackers and ads by default will block so many parties from collecting your data. Obviously there are also more sophisticated tracking technologies such as fingerprinting but if that is your thread you need to use Tor and probably Tails. Lets not create the idea that you can prevent this in general browsers by not recommending people to block as many as you can. Stopping trackers and ads from loading at all still benefits you the most as most of the internet still uses this. It's fairly easy to block all those nasty companies from loading their scripts in to your system, which is both great for privacy and security.

Similar to Tor browser, you can disable javascript and the ads won't load at all.

ph00lt0:
> > This is currently a big topic in browsers in our community. Some people seem to think that blocking ads makes you less private. I really disagree with this. Blocking trackers and ads by default will block so many parties from collecting your data. Obviously there are also more sophisticated tracking technologies such as fingerprinting but if that is your thread you need to use Tor and probably Tails. Lets not create the idea that you can prevent this in general browsers by not recommending people to block as many as you can. Stopping trackers and ads from loading at all still benefits you the most as most of the internet still uses this. It's fairly easy to block all those nasty companies from loading their scripts in to your system, which is both great for privacy and security.
> 
> Similar to Tor browser, you can disable javascript and the ads won't load at all. And of course any adblocker you want can be added to Safari, whereas focus you're stuck with Mozilla's.

Afaik you can't install extensions on Safari in Ios or has this changed recently? The only way to do it was always to install Adguard which for non technical users is a pain and complex process. Similar to disabling JavaScript, that isn't something non technical users will do and understand.

lx-is:
> > > This is currently a big topic in browsers in our community. Some people seem to think that blocking ads makes you less private. I really disagree with this. Blocking trackers and ads by default will block so many parties from collecting your data. Obviously there are also more sophisticated tracking technologies such as fingerprinting but if that is your thread you need to use Tor and probably Tails. Lets not create the idea that you can prevent this in general browsers by not recommending people to block as many as you can. Stopping trackers and ads from loading at all still benefits you the most as most of the internet still uses this. It's fairly easy to block all those nasty companies from loading their scripts in to your system, which is both great for privacy and security.
> > 
> > 
> > Similar to Tor browser, you can disable javascript and the ads won't load at all. And of course any adblocker you want can be added to Safari, whereas focus you're stuck with Mozilla's.
> 
> Afaik you can't install extensions on Safari in Ios or has this changed recently? The only way to do it was always to install Adguard which for non technical users is a pain and complex process. Similar to disabling JavaScript, that isn't something non technical users will do and understand.

You can install extensions. Extension functionality on Safari iOS was recently expanded quite significantly and now matches the capabilities of Safari macOS.

ph00lt0:
> You can install extensions. Extension functionality on Safari iOS was recently expanded quite significantly and now matches the capabilities of Safari macOS.

Oh that is completely new to me. Does this mean we can install ublock origin in safari on ios? That would be great news. I wonder if this will mean that other webkit browsers will get this ability as well.

lx-is:
> Oh that is completely new to me. Does this mean we can install ublock origin in safari on ios? That would be great news. I wonder if this will mean that other webkit browsers will get this ability as well.

No, Safari still doesn't have all the necessary APIs for uBlock Origin despite them being technically available in WebKit.

mfwmyfacewhen:
> > You can install extensions. Extension functionality on Safari iOS was recently expanded quite significantly and now matches the capabilities of Safari macOS.
> 
> Oh that is completely new to me. Does this mean we can install ublock origin in safari on ios? That would be great news. I wonder if this will mean that other webkit browsers will get this ability as well.

Doesn't look like ublock origin is available yet.

lx-is:
> Doesn't look like ublock origin is available yet. There is adguard though.

AdGuard limits its advanced protection (aka it's actual extension) behind a paywall. Free users are stuck with content blockers. I'm not sure how it compares from Firefox Focus' content blocking.

ph00lt0:
> Doesn't look like ublock origin is available yet. There is adguard though.

Yeah i just checked this fork that i knew of: https://github.com/el1t/uBlock-Safari/issues/158 seems to be dead.

ph00lt0:
I guess we could recommend people to use safari with that. Not against this at all, but removing FF seems not the right moment if no good adblocking solution is available atm.

mfwmyfacewhen:
> I guess we could recommend people to use safari with that. Not against this at all, but removing FF seems not the right moment if no good adblocking solution is available atm.

There are quite a few, including focus's blocking itself available as an extension. The user can easily pick whichever one they want.

ph00lt0:
> > I guess we could recommend people to use safari with that. Not against this at all, but removing FF seems not the right moment if no good adblocking solution is available atm.
> 
> There are quite a few, including focus's blocking itself available as an extension. The user can easily pick whichever one they want. Safari actually does some anti-fingerprinting and adblockers will make you more unique. I really think it's a bad idea to recommend them.

I can't find this focus extension, what is the name of it? 

Yeah the unique part we really disagree on. If you block trackers from loading at all they cannot fingerprint you. In most cases you will have blocked them so this argument really doesn't go up for me. Eitherway if that is what you are defending for, you need to use TOR.

mfwmyfacewhen:
> > > I guess we could recommend people to use safari with that. Not against this at all, but removing FF seems not the right moment if no good adblocking solution is available atm.
> > 
> > 
> > There are quite a few, including focus's blocking itself available as an extension. The user can easily pick whichever one they want. Safari actually does some anti-fingerprinting and adblockers will make you more unique. I really think it's a bad idea to recommend them.
> 
> I can't find this focus extension, what is the name of it?
> 
> Yeah the unique part we really disagree on. If you block trackers from loading at all they cannot fingerprint you. In most cases you will have blocked them so this argument really doesn't go up for me. Eitherway if that is what you are defending for, you need to use TOR.

Well safari already does tracker blocking by default, just not adblocking. The extension is available when you install the focus app.

_data/software/browsers-mobile/2_safari.yml
```diff
@@ -0,0 +1,50 @@
...
+  **Safari** is the default browser in iOS. It includes privacy features such as tracking protection, isolated private tabs, iCloud Private Relay to hide your ip address from trackers, automatic https upgrade, and the option to disable javascript.
+
+  These options can be found in *Privacy and Security* (⚙️ Settings → Safari → Privacy and Security).
+
+  #### **Cross-Site Tracking Prevention**
+  <ul style="list-style-type:none;padding-left:0;">
+    <li>Toggle On: <strong>Prevent Cross-Site Tracking</strong>.</li>
+  </ul>
+
+  #### **Hide IP Address**
```
lx-is, line 14: I think renaming this section to "Private Relay" would be better. And then more information on PR should be added, such as noting the difference between paid PR with iCloud+ and the free version.

I have iCloud+, so if you have any questions regarding it lmk.

It should also be noted that PR is **not** available in all countries and regions and is still in beta. https://support.apple.com/en-us/HT212614

ph00lt0:
> > > > I guess we could recommend people to use safari with that. Not against this at all, but removing FF seems not the right moment if no good adblocking solution is available atm.
> > > 
> > > 
> > > There are quite a few, including focus's blocking itself available as an extension. The user can easily pick whichever one they want. Safari actually does some anti-fingerprinting and adblockers will make you more unique. I really think it's a bad idea to recommend them.
> > 
> > 
> > I can't find this focus extension, what is the name of it?
> > Yeah the unique part we really disagree on. If you block trackers from loading at all they cannot fingerprint you. In most cases you will have blocked them so this argument really doesn't go up for me. Eitherway if that is what you are defending for, you need to use TOR.
> 
> Well safari already does tracker blocking by default, just not adblocking. The extension is available when you install the focus app.

So seems you would still need the focus app? Yeah safari does some 'limited' tracking prevention. But not much regarding ads, who often track everything they can.

_data/software/browsers-mobile/2_safari.yml
```diff
@@ -0,0 +1,50 @@
...
+
+  This is Apple's built-in VPN. It sends your connection through two relays operated by different entities. The first knows your Internet Protocol (IP) address, but not the website you are visiting. The second knows the website you are visiting, but not your IP address, instead providing a generalized identity and location to the destination website. This way, no single entity has the information to identify both you and the sites you visit.
```
lx-is, line 16: The "second" parties should be listed. These parties are Akamai, Cloudflare, and Fastly. https://www.streamingmediablog.com/2021/06/apple-private-relay.html

_data/software/browsers-mobile/2_safari.yml
```diff
@@ -0,0 +1,50 @@
+title: Safari
+type: Recommendation
+logo: /assets/img/browsers/safari.svg
+description: |
+  **Safari** is the default browser in iOS. It includes privacy features such as tracking protection, isolated private tabs, iCloud Private Relay to hide your ip address from trackers, automatic https upgrade, and the option to disable javascript.
```
lx-is, line 5: Capitalize IP

_data/software/browsers-mobile/2_safari.yml
lx-is, line 16: I also don't think the IP abbreviation needs to be specified. If anything, do it when it was first mentioned in the section.

mfwmyfacewhen:
> > > > > I guess we could recommend people to use safari with that. Not against this at all, but removing FF seems not the right moment if no good adblocking solution is available atm.
> > > > 
> > > > 
> > > > There are quite a few, including focus's blocking itself available as an extension. The user can easily pick whichever one they want. Safari actually does some anti-fingerprinting and adblockers will make you more unique. I really think it's a bad idea to recommend them.
> > > 
> > > 
> > > I can't find this focus extension, what is the name of it?
> > > Yeah the unique part we really disagree on. If you block trackers from loading at all they cannot fingerprint you. In most cases you will have blocked them so this argument really doesn't go up for me. Eitherway if that is what you are defending for, you need to use TOR.
> > 
> > 
> > Well safari already does tracker blocking by default, just not adblocking. The extension is available when you install the focus app.
> 
> So seems you would still need the focus app? Yeah safari does some 'limited' tracking prevention. But not much regarding ads, who often track everything they can.

Btw you mentioned tor, Safari has a sort of tor-lite (private relay) that puts your connection through two hops. I explain it in my pr. So I think there is a point to be made for blending in with other Safari users.

collections/_evergreen/browsers.md
```diff
@@ -31,7 +31,7 @@ description: |
 
 On Android, Mozilla's engine [GeckoView](https://mozilla.github.io/geckoview/) has yet to support [site isolation](https://hacks.mozilla.org/2021/05/introducing-firefox-new-site-isolation-security-architecture) or enable [isolatedProcess](https://bugzilla.mozilla.org/show_bug.cgi?id=1565196). Firefox on Android also doesn't yet have [HTTPS-Only mode](https://github.com/mozilla-mobile/fenix/issues/16952#issuecomment-907960218) built-in. We do not recommend Firefox or any Gecko based browsers at this time.
 
-On iOS all web browsers use the [WebKit](https://en.wikipedia.org/wiki/WebKit) browser engine, including Firefox. However, Firefox includes a few extra features like Tracking Protection and the ability to add search engines.
+On iOS all web browsers use the [WebKit](https://en.wikipedia.org/wiki/WebKit) browser engine, so all browsers on the App Store are essentially Safari under the hood.
```
lx-is, line 34: All iOS web browsers make use of [WKWebView](https://developer.apple.com/documentation/webkit/wkwebview), not broadly WebKit.

lx-is (changes requested):

_data/software/browsers-mobile/2_safari.yml
mfwmyfacewhen: Gotcha I'll update it

_data/software/browsers-mobile/2_safari.yml
mfwmyfacewhen: I copied it straight off of apple's site but yeah I think I'll delete that.

lx-is:
> For the record, I still disagree with the removal of Firefox Focus 😛

Sorry, I don't think I mentioned this earlier despite having the idea — I think Focus would fit as an honorable mention, not a full card. Just a link and a quick snippet stating it offers custom search engine functionality and built-in ad blocking.

mfwmyfacewhen:
> > For the record, I still disagree with the removal of Firefox Focus stuck_out_tongue
> 
> Sorry, I don't think I mentioned this earlier despite having the idea — I think Focus would fit as an honorable mention, not a full card. Just a link and a quick snippet stating it offers custom search engine functionality and built-in ad blocking.

If enough people want it I'll just keep it like it is now.

lx-is:
> PG has moved away from honorable mentions. Only full card worth tools are listed. Both Safari and Focus aren't ideal situations. But let's hope for a uBo version for Safari on ios.

Ah, wasn't aware of that.

I'm not sure if you saw my comment earlier, but UBO for Safari won't be coming unless Safari makes a more APIs available. https://github.com/uBlockOrigin/uBlock-issues/issues/1444 Though there is the Orion browser that circumvents this, it's an in-beta Safari clone that supports both Chrome & Firefox extensions on iOS (not sure how). It'll be interesting to see how the Safari extension space evolves over time.

_data/software/browsers-mobile/2_safari.yml
```diff
@@ -0,0 +1,56 @@
...
+    <li>Toggle On: <strong>Prevent Cross-Site Tracking</strong>.</li>
+  </ul>
+
+  #### **iCloud Private Relay**
+
+  This is Apple's [built-in VPN](https://support.apple.com/en-us/HT212614). It sends your connection through two relays operated by different entities. The first knows your [Internet Protocol (IP) address](https://en.wikipedia.org/wiki/IP_address), but not the website you are visiting. The second knows the website you are visiting, but not your IP address, instead providing a generalized identity and location to the destination website. This way, no single entity has the information to identify both you and the sites you visit.
+
+  iCloud Private Relay is not available in all countries and it is currently in beta.
+
+  It is available with an [iCloud+ subscription](https://support.apple.com/guide/icloud/icloud-overview-mmfc854d9604/icloud), however you can use it for trackers Safari detects for free.
```
lx-is, line 20: I think "Apple allows any qualifying user to pipe connections to known trackers through Private Relay. However, an iCloud+ subscription is required to pipe all Safari traffic through Private Relay" or something similar is a better way to word it.

_data/software/browsers-mobile/2_safari.yml
```diff
@@ -0,0 +1,56 @@
...
+  **Safari** is the default browser in iOS. It includes privacy features such as tracking protection, isolated private tabs, iCloud Private Relay to hide your IP address from trackers, anti-fingerprinting measures, and automatic https upgrade.
+
+  These options can be found in *Privacy and Security* (⚙️ Settings → Safari → Privacy and Security).
+
+  #### **Cross-Site Tracking Prevention**
+  <ul style="list-style-type:none;padding-left:0;">
+    <li>Toggle On: <strong>Prevent Cross-Site Tracking</strong>.</li>
+  </ul>
+
+  #### **iCloud Private Relay**
```
lx-is, line 14: It should probably be mentioned that Private Relay is not considered a replacement for a reputable VPN, especially because it only affects Safari traffic.

tommytran732:
I am **personally** not a big fan of extensions, I don't even like uBlockOrigin to begin with. IMO adblocking is a convenience thing and doesn't do anything for privacy & security. It is not a systematic way to solve any problem. 

Besides, on macOS safari incognito mode starts every new tab without any cookies or site data. Not sure if this is the case on iOS (I don't have an iPhone right now), but that should solve the problem with trackers.

lx-is:
> I am **personally** not a big fan of extensions, I don't even like uBlockOrigin to begin with. IMO adblocking is a convenience thing and doesn't do anything for privacy & security. It is not a systematic way to solve any problem.
> 
> Besides, on macOS safari incognito mode starts every new tab without any cookies or site data. Not sure if this is the case on iOS (I don't have an iPhone right now), but that should solve the problem with trackers.
> 
> As far as I understand it, Private Relay is a proxy and not a VPN. Please double check this.

Yeah, it's the same on iOS — just tested it.

And as for PR, from my understanding it functions more as a VPN, not a proxy.

lx-is:
> > I am **personally** not a big fan of extensions, I don't even like uBlockOrigin to begin with. IMO adblocking is a convenience thing and doesn't do anything for privacy & security. It is not a systematic way to solve any problem.
> > Besides, on macOS safari incognito mode starts every new tab without any cookies or site data. Not sure if this is the case on iOS (I don't have an iPhone right now), but that should solve the problem with trackers.
> > As far as I understand it, Private Relay is a proxy and not a VPN. Please double check this.
> 
> Yeah, it's the same on iOS — just tested it.
> 
> And as for PR, from my understanding it functions more as a VPN, not a proxy.

Actually no, I'm wrong. They are proxies, as stated in this video https://developer.apple.com/videos/play/wwdc2021/10096/

_data/software/browsers-mobile/2_safari.yml
```diff
@@ -0,0 +1,56 @@
...
+
+  This is Apple's [built-in VPN](https://support.apple.com/en-us/HT212614); however, it does not replace a proper VPN as it only affects Safari traffic. It sends your connection through two relays operated by different entities. The first knows your [Internet Protocol (IP) address](https://en.wikipedia.org/wiki/IP_address), but not the website you are visiting. The second knows the website you are visiting, but not your IP address, instead providing a generalized identity and location to the destination website. This way, no single entity has the information to identify both you and the sites you visit.
```
lx-is, line 16: Change to state that this is Apple's two-party proxy service, not built-in VPN.

_data/software/browsers-mobile/2_safari.yml
```diff
@@ -0,0 +1,56 @@
...
+
+  iCloud Private Relay is not available in all countries and it is currently in beta.
+
+  Apple allows any qualifying user to route connections to known trackers through Private Relay. However, an iCloud+ subscription is required to route all Safari traffic through Private Relay.
```
lx-is, line 20: Re-add link to iCloud+ information.

lx-is (changes requested):

_data/software/browsers-mobile/2_safari.yml
lx-is, line 16: > as it only affects Safari traffic

Just learned that it actually effects more than just Safari traffic, so "as it only affects Safari traffic" is slightly incorrect.

With iCloud+, Private Relay routes:
- Safari traffic
- DNS queries
- Subset of app traffic
  - Insecure HTTP traffic generated from an app, for example

lx-is (changes requested):

mfwmyfacewhen:
I think we leave private relay for a different part of the site, yeah?

dngray:
> I think we leave private relay for a different part of the site, yeah?

Yeah this makes sense for the MacOS/iOS pages that will be created.

PR #645, opened 2022-02-22
Add anchors to the appropriate headings previously missing them

akumeilol:
This PR adds anchors to the appropriate headings that were previously missing it.

Preview: https://deploy-preview-645--privacyguides.netlify.app

## Notes

- I tried to work around the existing codebase as much as possible—I didn't want to break anything by modifying anything too heavily. I've never worked on a Jekyll project this large before.
- **Duplicate heading names (i.e. Advantages, Disadvantages, Notes, etc) should use a tier 4 heading or higher.** Also use tier 4 or higher for unimportant headings.
- Sorry Jonah for likely using up a ton of Netlify build minutes! I could not for the life of me get it to build locally.

## Evergreen markdown anchors

https://github.com/privacyguides/privacyguides.org/pull/645/files#diff-9256c9e8ba38aa09096cc4be5bf614026b87485910661016c66e2cfe6fb35d80

https://github.com/privacyguides/privacyguides.org/pull/645/files#diff-f706bb2d5260d4b6b065c9d8065984f748f4e35c773e7f7d89d1bc0efd8bb930

Markdown pages using the Evergreen page layout (cool layout name, btw) make use of the [Jekyll Pure Liquid Heading Anchors](https://github.com/allejo/jekyll-anchor-headings) _includes file. This liquid snippet will automatically add anchors to heading levels 2-3, as per my specification. The anchors make use of the same Font Awesome icon that the legacy page anchors use.

I chose to limit the anchors between heading levels 2-3 because it fits nicely with all the pre-existing content. (i.e. headings with the same name typically all use tier 4 headings)

This _includes file can be used in any other page layout in the future.

## cardv2 anchors

https://github.com/privacyguides/privacyguides.org/pull/645/files#diff-c91b811f40687e522be3303326295aafcbe9faf555e007976b14799bcdb1e0a5

The titles on the cardv2 _includes will now be anchored. An example of this would be the title of Thunderbird in the desktop email clients section.

This works by taking the title of the card, [slug-ifying](https://jekyllrb.com/docs/liquid/filters/#slugify) it, and using the slug-ified title as a link.

## Recommendation card anchors

https://github.com/privacyguides/privacyguides.org/pull/645/files#diff-9bbe2d33f724cdb806827cc4ee7790c3ceae2c81b52cf6cc9cbdb0dac3aa7509

These use the same slug-ify method as cardv2 anchors.

## Specific edits

Not all appropriate headings were anchored using the above additions. So, I went through each page and properly added them.

### Android page

[...]

PR #720, opened 2022-03-15
Restructure how to get apps and align the usage of headers.

ph00lt0:
I open this PR because I noticed that the headers used on this Android page are not aligned. Certain apps have Notes with H3 and others H4 so I fixed this now.

While doing this I noticed that the main point of the "Where to get your apps should actually be above all the clients. It makes much more sense to start on where to get apps from and then list the locations rather then in reverse. By making the clients part of this 'section' it feels more like a guide rather then lose statements.

### DIFF
--- a/_data/operating-systems/android-applications/4_secure_camera.yml
+++ b/_data/operating-systems/android-applications/4_secure_camera.yml
@@ -9,7 +9,7 @@ description: |
   - Use of the new [Media](https://developer.android.com/training/data-storage/shared/media) API, therefore [storage permissions](https://developer.android.com/training/data-storage) are not required
   - Microphone permission not required unless you want to record sound
 
-  ### Notes
+  #### Notes
   Metadata is not currently deleted from video files but that is planned.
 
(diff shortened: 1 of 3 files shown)


akumeilol:
I made this same change in #645

ph00lt0:
> I made this same change in #645

I see now @akumeilol I missed that, but just looking at your PR you have some rebasing to do, it currently resets quite some changes.

dngray:
I think this is fixed now https://github.com/privacyguides/privacyguides.org/commit/00f18af4659cd6dd628ae3fdedf95b532b7a5e64

ph00lt0:
@dngray the order of the page is still strange. Where to get apps should be the main header, instead of fdroid above it. 

The proposed structure would be:
Where to get your apps?
- GrapheneOS
- F-Droid
   - Droid-ify
 - Aurora (add new card)

dngray:
@ph00lt0 check out https://github.com/privacyguides/privacyguides.org/pull/955 is that better?

PR #746, opened 2022-03-19
Make the navbar work without javascript

flexagoon:
## Description

Make the navbar work with javascript disabled. This is achieved by adding a `noscript` class to the navbar which is removed if JS is enabled, and styling the `noscript` class to expand dropdowns on hover.

Resolves: #571

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-expand-lg navbar-light bg-primary">
+<nav class="navbar navbar-expand-lg navbar-light bg-primary noscript">
   <div class="container">
     <a class="navbar-brand" href="/">
       <picture>
--- a/_sass/noscript.scss
+++ b/_sass/noscript.scss
@@ -1,0 +1,13 @@
+.noscript {
+  .dropdown:hover .dropdown-menu {
+    display: block;
+    right: 0;
+    left: auto;
+  }
+  .navbar-toggler {
+    display: none;
+  }
+  #navbarSupportedContent {
+    display: block;
+  }
+}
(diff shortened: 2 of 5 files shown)


elitejake:
In my opinion, `noscript` works better than the original because it shows the drop down menu by just hovering over it.

tommytran732:
This is causing a strange behavior on Mobile. If you open the preview from this PR, the navbar is always extended first, then minimizes itself after a bit. This behavior does not happen on the current site. Could you look into this?

flexagoon:
@tommytran732 I think it's like that because the "noscript" class only gets removed after the page is fully loaded. I don't really know how to fix this while still making the site work on mobile without javascript, but I guess I'll try to find a solution.

Or alternatively I can only make a no-js version for desktop and still require javascript on mobile

tommytran732:
I'd prefer a solution for both mobile and desktop, but worst case scenario we can do desktop only

ph00lt0:
https://tutorialzine.com/2015/08/quick-tip-css-only-dropdowns-with-the-checkbox-hack

This is the way to do it imo

flexagoon:
@ph00lt0 I am aware of this method, but the js version of the website uses Bootstrap dropdown buttons, and I didn't want to re-make the entire navbar with checkboxes and only show that when js is disabled while making the Bootstrap version hidden, since that seems like an even more hacky approach

ph00lt0:
@flexagoon I see your point, I just think it could replace the javascript version all together. We don't need JavaScript to make this behavior.

DeiAsPie:
Finding a solution that works without js anyways and making that the default would be great imo

jonaharagon:
Implemented in #829

PR #752, opened 2022-03-20
Secure PDF Viewer

dngray:


Tvirin:
I wonder if it is good to suggest when they have this attitude

https://gitlab.com/IzzyOnDroid/applists/-/issues/16

dngray:
> I wonder if it is good to suggest when they have this attitude

I think they are concerned about supply chain issues. There is nothing wrong with that.

PR #755, opened 2022-03-20
Fix archinstall's release date

Alonely0:
https://www.reddit.com/r/PrivacyGuides/comments/tiq8e9/incorrect_info_on_new_linux_page/

## Description

Resolves: #


mfwmyfacewhen:
Not sure it even needs a date tbh. "Arch has an [official installer](https://wiki.archlinux.org/title/Archinstall) to make the installation process a little easier." is the only necessary part I think.

dngray:
I'm thinking we'll remove the date [as it won't really matter](https://www.phoronix.com/scan.php?page=news_item&px=Arch-Linux-Does-Archinstall) in the future and it won't be important to know that it is "new".

Issue #730, opened 2022-03-17
Recent Updates to DivestOS


SkewedZeppelin:
## Description
DivestOS now has many additional hardening features thanks to GrapheneOS, including the hardened memory allocator on 18.1.
https://divestos.org/index.php?page=news#2022-03.3

**URL of affected page:** https://www.privacyguides.org/android/

tommytran732:
Are the builds all ready yet? I updated to the March 14 build and the Bromite webview option mentioned in the update notes doesn't actually work (it only allows the user to use the Mulch webview and that's it).

dngray:
Also happy to take PR from you @SkewedZeppelin, and yes I do note you're the author of DivestOS, but that's okay, we will check/commit accuracy.

PR #763, opened 2022-03-22
DivestOS updates (#730)

SkewedZeppelin:
## Description

Resolves: #730

Fixes a typo in UnifiedNlp and notes that the hardened malloc is now available on 18.1 as per recent updates: https://divestos.org/index.php?page=news#2022-03.3

Easiest way to verify the malloc is in use is looking at the VIRT column of top/htop in Termux or Terminal Emulator:
- 100GB+ is GrapheneOS hardened_malloc
- 10GB or so is scudo
- very little is jemalloc

In the case of a 32-bit device:
- 2GB or so is scudo
- very little is jemalloc

Scudo and Jemalloc also both show up in `cat /proc/self/maps`, hardened_malloc only shows up in [debuggable](https://github.com/GrapheneOS/hardened_malloc/blob/main/Android.bp#L80) Android builds


dngray:
> UnifiedNlp

:grin: how nobody ever saw that typo.... 

LGTM

SkewedZeppelin:
Thanks!

PR #767, opened 2022-03-22
Revamping the Encrypted DNS page (#767)

dngray:
## Revamping the DNS page to fix several issues that were discussed with the community.

This is a continuation of https://github.com/privacyguides/privacyguides.org/pull/575 which was a WIP.

## Resolves: Issues considered solved as per my knowledge:
- #572    
- #561  
- #543    
- #321 
- https://github.com/privacyguides/privacyguides.org/pull/757
 
## Recommendations considered:
- #556 
- #593 
- #197 
- #567 
- #241 
- #252 
- #111 

## Changes made:
- Added [when to use dns](https://github.com/dngray/encrypted-dns/blob/main/dns.svg) image by @dngray 
- Removed UncensoredDNS
- Added MullvadDNS
- Removed columns for Qname minimization and DNSSEC
- Added column for ECS
- Removed the warning of Android DNS being ineffective when using VPN
- Added definitions of Qname Minimization, DNSSEC and ECS
- Replaced Nebulo by a better project [RethinkDNS](https://rethinkdns.com/)
- Removed third party clients and suggest using native implementations instead

## Effected page: 
https://privacyguides.org/providers/dns/

PR #757, opened 2022-03-21
Update encrypted DNS information for Apple devices

lx-is:
The inclusion of the 'Secure DNS profile creator' utility could be left out, but I think it serves a useful enough purpose while fitting within *most* threat models. The information that follow its mention allows readers to determine if it's right for them or not. Realistically, unsigned profiles aren't a huge deal — very easily, you can open them in a text editor to verify the [contents](https://user-images.githubusercontent.com/100533118/159229500-5d414e9c-4618-4fd1-8c8c-244ac1401757.png) before installation.

### All changes
- Minor phrasing updates
- Mention the DNS Setting API available to apps
- Mention VPN DNS priority
- Mention Secure DNS profile creator
- Information on signed profiles
- Add link to ControlD signed profiles
- Add macOS instructions (#782)

PR #782, opened 2022-03-25
Clarified Native Operating System section for DNS

gracjankn:
I made the explanation slightly easier to undestand and less wordly. I also added instructions for macOS, as the structure of settings is different than on iOS.


lx-is:
The DNS section is currently being rewritten (#767)

I currently already have #757 open, which is going to be looked at once the main rewrite is complete. I'll update my PR to include macOS instructions. If you want to take a look at my PR and leave a review, feel free :)

gracjankn:
You're right, I haven't noticed that. I'm closing the pull request.

PR #784, opened 2022-03-25
DivestOS's support for the Bromite SystemWebView

tommytran732:


ph00lt0:
@tommytran732 nice, but does this mean that PG recommends this? I think we should be clear on this.

tommytran732:
It's a rather nuanced topic whether to use it or not tbh.

On one hand Bromite ia yet another party to trust. On another hand Bromite has more hardening patches and i have heard that some of the patches Mulch took from Vanadium and Bromite dont go well with each other. Also, when Bromite fell behind in the past, Mulch managed to stay relatively up to date. However, right now, Bromite is actually more up to date that Mulch.

I think this is rather a choice for the user since there is no clear recommendation to be made.

ph00lt0:
> It's a rather nuanced topic whether to use it or not tbh.
> 
> On one hand Bromite ia yet another party to trust. On the other hand Bromite has more hardening patches and i have heard that some of the patches Mulch took from Vanadium and Bromite dont go well with each other. Also, when Bromite fell behind in the past, Mulch managed to stay relatively up to date. However, right now, Bromite is actually more up to date that Mulch.
> 
> I think this is rather a choice for the user since there is no clear recommendation to be made.

I share the same view. I just think it may be confusing to average user reading the website.

PR #790, opened 2022-03-26
Update WireGuard information on VPN provider page

lx-is:
Someone on Matrix mentioned that currently the page states WireGuard is "experimental" and "not currently recommended for production use." I think these statements aren't entirely accurate anymore; WireGuard has proven itself as is demonstrated by almost all reputable VPN companies ([Mullvad](https://mullvad.net/en/help/why-wireguard/), [IVPN](https://www.ivpn.net/wireguard/), [ProtonVPN](https://protonvpn.com/blog/wireguard/), etc) adopting it as a default and by it being incorporated into the Linux kernel.

### Changes
- Update WireGuard description for all providers
- Update minimum to qualify criteria to include WireGuard
- Update criteria alert to include WireGuard and add line break before long link
- Include WireGuard support for ProtonVPN
- State for which apps, per provider, WireGuard is the default
- Mention which providers feature a WireGuard config generator for use with the official WireGuard apps
- Include a legal note as required by [WireGuard's trademark policy](https://www.wireguard.com/trademark-policy/)
- Add more spacing above VPN warning so it doesn't appear as if it's specifically applying to IVPN

### DIFF
--- a/_includes/legacy/sections/vpn.html
+++ b/_includes/legacy/sections/vpn.html
@@ -1,7 +1,7 @@
 <h2 id="vpn" class="anchor"><a href="#vpn"><i class="fas fa-link anchor-icon"></i></a> Recommended VPN Services</h2>
 
 <div class="alert alert-success" role="alert">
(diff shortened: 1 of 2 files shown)


dngray:
Yes this should be re-worded, at the point it was written, I don't think any provider supported it.

PR #797, opened 2022-03-26
Improving accessibility labels

gracjankn:


dngray:
This will get resolved when we do https://github.com/privacyguides/privacyguides.org/issues/802

dngray:
Not needed anymore https://github.com/privacyguides/privacyguides.org/pull/829

PR #804, opened 2022-03-27
Slight correction to firefox card

mfwmyfacewhen:
Added that the search suggestion options are only for the US

### DIFF
--- a/_data/software/browsers-desktop/1_firefox.yml
+++ b/_data/software/browsers-desktop/1_firefox.yml
@@ -17,7 +17,7 @@ description: |
   </ul>
   You can still stay logged into websites by allowing exceptions.
 
-  ##### **Disable Search Suggestions**
+  ##### **Disable Search Suggestions (U.S. Only)**
    <ul style="list-style-type:none;padding-left:0;">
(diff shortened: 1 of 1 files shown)


_data/software/browsers-desktop/1_firefox.yml
```diff
@@ -17,7 +17,7 @@ description: |
   </ul>
   You can still stay logged into websites by allowing exceptions.
 
-  ##### **Disable Search Suggestions**
+  ##### **Disable Search Suggestions (U.S. Only)**
```
lx-is, line 20: It would be better to put just say that these features may not be available depending on one's region, just in case the feature expands to more areas in the future.
```suggestion
  ##### **Disable Search Suggestions**
   *These features may not be available depending on your region.*
```

Lostname777: Yes, as my Firefox do contain Search Suggestions section but doesn't have the options named like you have mentioned. My firefox says:

**Search Suggestions**
Choose how suggestions from search engines appear.
```
Provide search suggestions
Show search suggestions in address bar results
Show search suggestions ahead of browsing history in address bar results
Show search suggestions in Private Windows
```

Issue #806, opened 2022-03-28
Change under centralized networks


Lostname777:
On [this](https://privacyguides.org/real-time-communication/#centralized-networks) page, under disadvantages, in the second point, you have added a link connected to a word "ownership". That link is broken as it links to your previous blog.

PR #808, opened 2022-03-28
Fix link to old blog

elitejake:
## Description

Resolves: #806


dngray:
hmm, weird that I missed that one.

PR #818, opened 2022-03-29
Calendar and Contacts legacy migration

dngray:


dngray:
Would like to write a bit more about Tutanota's calendar @tommytran732

PR #820, opened 2022-03-30
Notebooks legacy migration

dngray:


dngray:
We can remove paperwork.cloud, seems the site is down and the project is dead.

https://github.com/paperwork

> [dead] OpenSource note-taking & archiving alternative to Evernote, Microsoft OneNote & Google Keep

All repositories are archived.

PR #823, opened 2022-03-30
Small grammar correction and link to Postmill.

NylaTheWolf:
Not sure if I'm allowed to make two changes at once but I felt it would be easier to kill two birds with one stone. I fixed up a grammatical error ("A is a public..."). I also decided to link to the GitLab repo for Postmill, as I never heard of Postmill before and I figured others might not know what it was either.

## Description

Resolves: #

### DIFF
--- a/_includes/legacy/sections/social-news-aggregator.html
+++ b/_includes/legacy/sections/social-news-aggregator.html
@@ -26,7 +26,7 @@ web="https://join-lemmy.org/"
 <h3>Worth Mentioning</h3>
 
 <ul>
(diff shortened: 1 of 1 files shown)


dngray:
> Not sure if I'm allowed to make two changes at once but I felt it would be easier to kill two birds with one stone.

Sure no problem, they're to do with the same issue (and minor), what we like to avoid is issues like [this](https://github.com/privacyguides/privacyguides.org/pull/579) as they are a pain to review.

PR #825, opened 2022-03-31
File sharing/sync legacy migration

dngray:


dngray:
At some point when we have a hardware page I wouldn't mind moving the Freedombox there.

PR #830, opened 2022-04-01
Update fonts with Flathub icon

elitejake:
Update all the font files with Flathub icon.


jonaharagon:
Appreciate it, but when we merge #829 (very shortly) the icons are going to be refactored into individual svg icons. It's a more workable solution. Do you have the individual icon file?

See that branch for example: https://github.com/privacyguides/privacyguides.org/tree/pr-mkdocs_md/theme/.icons/pg

PR #832, opened 2022-04-01
Make Session's logo appear next to its recommendation

Beventar:
## Description

Resolves: # 

I fixed so there's a Session logo next to its recommendation on the page instead of Briar's logo


dngray:
Well spotted.

PR #852, opened 2022-04-02
Remove wrong ios app store link for FairEmail

GetBoz:
## Description

Resolves: Removes wrong ios app store link for FairEmail#


dngray:
Well spotted, that shouldn't be there.

PR #876, opened 2022-04-03
Remove ExifCleaner privacy policy

elitejake:
## Description

ExifCleaner's privacy policy was incorrectly linked to Bitwarden's.


dngray:
That must have slipped in there when I copied the Bitwarden entry as a template, and edited the fields.

PR #878, opened 2022-04-04
Minor download icon updates

lx-is:
### Changes
- Use Mac App Store title and icon for MAS apps
- Reorder some download icons for site-wide consistency
- Update Session F-Droid link to use F-Droid icon

### DIFF
--- a/docs/browsers.md
+++ b/docs/browsers.md
@@ -23,7 +23,7 @@ These are our current web browser recommendations and settings. We recommend kee
 
     **Downloads**
     - [:fontawesome-brands-windows: Windows](https://www.torproject.org/download/)
-    - [:fontawesome-brands-apple: macOS](https://www.torproject.org/download/)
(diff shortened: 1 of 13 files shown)


dngray:
> - Use Mac App Store title and icon for MAS apps

Might as well replace it in all files:

```
find . -type f -print0 | xargs -0 sed -i "s/\[\:fontawesome-brands-apple\: macOS\]/\[\:fontawesome-brands-app-store\: Mac App Store\]/g"
```

lx-is:
> Might as well replace it in all files

I went through each page individually and replaced only the Mac App Store links. I think a distinction should be made for MAS apps vs. standard Mac apps due to the sandboxing requirement in the App Store.

dngray:
Ah yes, you're 100% right. think this is fixed https://github.com/privacyguides/privacyguides.org/pull/883

PR #884, opened 2022-04-04
Clarify iOS browser restrictions

lx-is:
> all browsers on the App Store are essentially Safari under the hood

I didn't like how this sounded as it's not entirely accurate as I've recently learned.\
See https://developer.apple.com/forums/thread/88052 & https://news.ycombinator.com/item?id=15872111

This rewrite is also less scientific, as in it's easier to understand. Most people don't know or care what WKWebView is.

I also removed the Safari download button because it's already built into any supported OS, so I don't really see the point of having it when the homepage and privacy policy are already linked.


dngray:
> I didn't like how this sounded as it's not entirely accurate as I've recently learned.
> See https://developer.apple.com/forums/thread/88052 & https://news.ycombinator.com/item?id=15872111

Heh, that's what sometimes happens when you provide too much of a verbose answer.

> This rewrite is also less scientific, as in it's easier to understand. Most people don't know or care what WKWebView is.

Agreed.

Issue #848, opened 2022-04-02
Website Issue | Remove mathjax code and use privacy plugin?


dngray:
Will remove external mathjax resource, can use some local code for now.

Looking at https://squidfunk.github.io/mkdocs-material/setup/ensuring-data-privacy/#built-in-privacy-plugin also.

realguyman:
Do we need to include MathJax from a CDN, or are we strictly removing it?

dngray:
> Do we need to include MathJax from a CDN, or are we strictly removing it?

Removing it, i prefer not have external links, it was added in early prototype, as we use some on the Android page.

PR #891, opened 2022-04-04
Remove remote MathJax JS

dngray:
Closes https://github.com/privacyguides/privacyguides.org/issues/848


elitejake:
We only have a single instance where MathJax is used, right? In my opinion, we should just use inline SVG.

dngray:
> We only have a single instance where MathJax is used, right? In my opinion, we should just use inline SVG.

Nah, because these are bad for screen readers.

PR #904, opened 2022-04-04
Detail the DivestOS hardening

SkewedZeppelin:
## Description

More thoroughly notes what features DivestOS includes.
- https://divestos.org/index.php?page=news
- https://divestos.org/index.php?page=technical_details

Also sadly removes the Bromite WebView option, for some unknown reason it causes the 15.1 branch to not boot.

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -39,11 +39,11 @@ Currently, only [Pixel phones](https://grapheneos.org/faq#device-support) meet i
 
     It has some privacy features on top of AOSP, including [Datura firewall](https://calyxos.org/docs/tech/datura-details), [Signal](https://signal.org) integration in the dialer app, and a built in panic button. CalyxOS also comes with firmware updates and signed builds, so [verified boot](https://source.android.com/security/verifiedboot) is fully supported.
 
-    To accomodate users who need Google Play Services, CalyxOS optionally includes [MicroG](https://microg.org/). With MicroG, CalyxOS also bundles in the [Mozilla](https://location.services.mozilla.com/) and [DejaVu](https://github.com/n76/DejaVu) location services.
-
-    Currently, CalyxOS only supports [Pixel phones](https://calyxos.org/docs/guide/device-support/).
-
     [Visit calyxos.org](https://calyxos.org/){ .md-button .md-button--primary } [Privacy Policy](https://calyxinstitute.org/legal/privacy-policy){ .md-button }
+
+To accomodate users who need Google Play Services, CalyxOS optionally includes [MicroG](https://microg.org/). With MicroG, CalyxOS also bundles in the [Mozilla](https://location.services.mozilla.com/) and [DejaVu](https://github.com/n76/DejaVu) location services.
+
+Currently, CalyxOS only supports [Pixel phones](https://calyxos.org/docs/guide/device-support/).
 
 !!! attention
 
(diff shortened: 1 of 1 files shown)


docs/android.md
```diff
@@ -60,9 +60,9 @@ Currently, only [Pixel phones](https://grapheneos.org/faq#device-support) meet i
...
+DivestOS has automated kernel vulnerability ([CVE](https://en.wikipedia.org/wiki/Common_Vulnerabilities_and_Exposures)) [patching](https://gitlab.com/divested-mobile/cve_checker), fewer proprietary blobs, a custom [hosts](https://divested.dev/index.php?page=dnsbl) file, and [F-Droid](https://www.f-droid.org) as the app store. It includes [UnifiedNlp](https://github.com/microg/UnifiedNlp) for network location. It's hardened WebView, [Mulch](https://gitlab.com/divested-mobile/mulch), enables [CFI](https://en.wikipedia.org/wiki/Control-flow_integrity) for all architectures and includes [network state partitioning](https://developer.mozilla.org/en-US/docs/Web/Privacy/State_Partitioning#network_partitioning). DivestOS also includes kernel patches from GrapheneOS and enables all available kernel security features via [defconfig hardening](https://github.com/Divested-Mobile/DivestOS-Build/blob/master/Scripts/Common/Functions.sh#L758). All kernels newer than version 3.4 include full page sanitization and all ~22 Clang-compiled kernels have `-ftrivial-auto-var-init=zero` enabled.
```
dngray, line 63: > full page sanitization

I would have linked a link there, perhaps https://portswigger.net/daily-swig/google-and-mozilla-unveil-plans-to-bake-html-sanitization-into-their-browsers

https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API seems a bit too "terse" for this kind of reference.

> all ~22 Clang-compiled kernels have `-ftrivial-auto-var-init=zero` enabled

Would be good if we coudl find a source for that, that explains what it is.

docs/android.md
```diff
@@ -60,9 +60,9 @@ Currently, only [Pixel phones](https://grapheneos.org/faq#device-support) meet i
...
 
-DivestOS 16.0, 17.1, and 18.1 implements GrapheneOS's [`INTERNET`](https://developer.android.com/training/basics/network-ops/connecting) permission toggle and [hardened memory allocator](https://github.com/GrapheneOS/hardened_malloc).
+DivestOS 16.0, 17.1, and 18.1 implements GrapheneOS's [`INTERNET`](https://developer.android.com/training/basics/network-ops/connecting) permission toggle, [hardened memory allocator](https://github.com/GrapheneOS/hardened_malloc), [exec-spawning](/android/#additional-hardening), JNI constification, and partial bionic hardening patchsets. 17.1 and 18.1 feature GrapheneOS's per-network full [MAC randomization](https://en.wikipedia.org/wiki/MAC_address#Randomization) option, `ptrace_scope` control, and automatic reboot/Wi-Fi/Bluetooth timeout options. All branches additionally have various miscellaneous patches courtesy of GrapheneOS.
```
dngray, line 65: > [exec-spawning](/android/#additional-hardening), JNI constification, and partial bionic hardening patchsets. 17.1 and 18.1 feature GrapheneOS's per-network full MAC randomization option, ptrace_scope control, and automatic reboot/Wi-Fi/Bluetooth timeout options. All branches additionally have a smattering of various miscellaneous patches courtesy of GrapheneOS.

I'm thinking this is getting a little "too complex" for this description here, maybe we should move them to [additional hardening of DivestOS](https://www.privacyguides.org/android/#additional-hardening-divestos), then describe what they are. This way it won't overwhelm readers of the recommendation, but if they *want to know more* the option is there, without having to do extensive research.

I would like to then see a source explaining:

- JNI constification
- partial bionic hardening patchsets (link to upstream?)
- `ptrace_scope` control (another one that could be moved down to additional hardening.
- automatic reboot/Wi-Fi/Bluetooth timeout options

dngray (changes requested):

docs/android.md
tommytran732: This is outside of the recommendation card, so I think it is fine. In fact, I'd cut off the recommendation card for CalyxOS and make it look more like GrapheneOS or DivestOS. The card can have a brief description of the recommendation, then anything outside of it can be used to explain it further.

It does not make sense to make an "additional hardening of DivestOS" if you are not going to make one for GrapheneOS and Calyx IMO.

tommytran732: I added the link for ptrace and sanitization sanitation. The auto reboot/Wifi-/Bluetooth only has 1 mention on https://grapheneos.org/features, so I don't now if you even wanna link that.

I also split off the CalyxOS's card so it's consistent with GrapheneOS and DivestOS.

I think it's good for merge.

docs/android.md
SkewedZeppelin: page sanitization is zeroing of RAM, not browser

auto-var-init: https://reviews.llvm.org/D54604?id=174471

docs/android.md
```diff
@@ -60,9 +60,9 @@ Currently, only [Pixel phones](https://grapheneos.org/faq#device-support) meet i
...
...estOS has automated kernel vulnerability ([CVE](https://en.wikipedia.org/wiki/Common_Vulnerabilities_and_Exposures)) [patching](https://gitlab.com/divested-mobile/cve_checker), fewer proprietary blobs, a custom [hosts](https://divested.dev/index.php?page=dnsbl) file, and [F-Droid](https://www.f-droid.org) as the app store. It includes [UnifiedNlp](https://github.com/microg/UnifiedNlp) for network location. It's hardened WebView, [Mulch](https://gitlab.com/divested-mobile/mulch), enables [CFI](https://en.wikipedia.org/wiki/Control-flow_integrity) for all architectures and includes [network state partitioning](https://developer.mozilla.org/en-US/docs/Web/Privacy/State_Partitioning#network_partitioning). DivestOS also includes kernel patches from GrapheneOS and enables all available kernel security features via [defconfig hardening](https://github.com/Divested-Mobile/DivestOS-Build/blob/master/Scripts/Common/Functions.sh#L758). All kernels newer than version 3.4 include full page [sanitization](https://portswigger.net/daily-swig/google-and-mozilla-unveil-plans-to-bake-html-sanitization-into-their-browsers) and all ~22 Clang-compiled kernels have `-ftrivial-auto-var-init=zero` enabled.
```
qua3k, line 63: What is this? Zeroing pages on free with init_on_free=1 is unrelated to XSS sanitization. How is a browser feature related to the kernel?

docs/android.md
dngray: > It does not make sense to make an "additional hardening of DivestOS" if you are not going to make one for GrapheneOS and Calyx IMO.

I'm inclined to agree with you.

> This is outside of the recommendation card, so I think it is fine. In fact, I'd cut off the recommendation card for CalyxOS and make it look more like GrapheneOS or DivestOS. The card can have a brief description of the recommendation, then anything outside of it can be used to explain it further.

Fair enough, that works too.



> I also split off the CalyxOS's card so it's consistent with GrapheneOS and DivestOS.

:+1:

docs/android.md
dngray: That was my bad, i added that link oops.

docs/android.md
dngray: So I remember what JNI was from my Java days.

Just wondering if we can get a reference for:

- "constification", discussed within that paragraph.
- automatic reboot/Wi-Fi/Bluetooth timeout options

I don't think it would be clear what these things are.

dngray: I'm guess it has something to do with https://github.com/GrapheneOS/platform_frameworks_base/commit/b3a5c3db7fa158619656871ba3c5e3a73ee73725

dngray: A better reference would be https://en.wikipedia.org/wiki/Const_(computer_programming)

PR #905, opened 2022-04-05
Update BitLocker image

lx-is:
Updated the BitLocker image to match the Windows 11 style and rephrased the beginning of the BitLocker description to make it clear that BitLocker isn't always enabled by default. Not that it was claiming this before, but I can imagine it being accidentally interpreted as such.


dngray:
:+1:

PR #907, opened 2022-04-05
Update all tools page

lx-is:
### Changes
- Added "Email Cloaking Services" and "Self-Hosting Email" under the email section
- Added missing icons
  - Included mini icons for this purpose
- Update ProtonMail logo to match their new color scheme
- Added some color to the StartMail dark mode logo
- Fixed the Secure Camera card


dngray:
> * Included mini icons for this purpose

I'm wondering if this is necessary, considering the ones made mini are all SVG.

lx-is:
> I'm wondering if this is necessary, considering the ones made mini are all SVG.

As in you'd rather just replace all the wide icons with these?

dngray:
> > I'm wondering if this is necessary, considering the ones made mini are all SVG.
> 
> As in you'd rather just replace all the wide icons with these?

No, i think that wouldn't work hehe, yeah wasn't quite sure what you meant by "mini" but now i've checked it makes sense.

PR #909, opened 2022-04-05
Fix the missing DecSync logo in the tools page

Beventar:
## Description

Resolves: # 

Made DecSync logo appear in tools and its own recommendation page.


lx-is (changes requested):
Since the icon is just pure white, a version for light mode is needed too. The icons should be added on [/calendar-contacts/#decsync](https://www.privacyguides.org/calendar-contacts/#decsync) too.

Also I would run the svgs through [SVGOMG](https://jakearchibald.github.io/svgomg/). I'm not sure if it's best practice or not, but I typically enable all the toggles except 'remove xmlns'

jonaharagon (approved):
image

image

dngray:
I did come across this logo, but it didn't really apply to the DecSync project, but rather DecSync CC.

Considering this is under the Contacts/Sync section maybe we should list DecCC instead:

- https://play.google.com/store/apps/details?id=org.decsync.cc
- https://f-droid.org/app/org.decsync.cc

dngray (changes requested):
With SVGs we optimize these https://github.com/privacyguides/privacyguides.org/wiki/Contributing#svg

So I'll do that.

PR #916, opened 2022-04-05
Split DNS page into 2 pages

### LINKED ISSUE #834: Re-arrange some of the DNS page content
## Description

**URL of affected page:** /dns

One of the comments in https://reddit.com/r/PrivacyGuides/comments/tsujrt/new_encrypted_dns_page/i2ubfny/ suggested re-arranging the page so the tutorial stuff is off by itself.

I think this is a good idea, and now we're on mkdocs that should be a lot easier to do.

Might expand it with a bit about testing DNSSEC etc too while I am at it.

jonaharagon:
Based on @D0T1X's suggestion at https://github.com/privacyguides/privacyguides.org/issues/834#issuecomment-1086898506

This splits DNS into two pages, an [informational page](https://deploy-preview-916--privacyguides.netlify.app/technology/dns/) and a [server recommendations page](https://deploy-preview-916--privacyguides.netlify.app/dns/).

If we like this, it should probably be used as a template for a few other pages as well. Basically anything *explaining a concept* should go in the introductory section, and anything we *recommend* **or** anything explaining *how to set up something* on a platform goes on the recommendations page. This is why, for example, the "Native Operating System Support" section remains on the current page.

I don't think this closes the original issue, because the new page being created could probably use some reworking on its own.


jonaharagon:
It occurs to me we should probably have *three* primary website sections, an introductory section, an advanced guides section (where this page probably belongs), and a recommendations section.

freddy-m:
@jonaharagon those sections seem fairly sensible. Perhaps an intermediate one as well?

tommytran732 (approved):
I think this is good for merge. It is better than a giant DNS page. We could split this into 3 pages later in a different PR.

tommytran732:
> @jonaharagon those sections seem fairly sensible. Perhaps an intermediate one as well?

Probably will make it a bit too fragmented.

PR #919, opened 2022-04-05
Integrating Metadata Removal

lx-is:
I started researching how to integrate ExifTool into macOS's context menu for my own purposes, but somewhere along that path I thought it might serve as a cool guide.

I'm not sure if this type of content fits on the site, so feel free to close this PR. If that's the case I can just post it somewhere else, no big deal.

Scripting is a bit out of my comfort zone, so I'm sure there's room for improvement.


tommytran732 (approved):
If you ask me, this is the kind of content we want more of.

elitejake:
What if we make this into a Security Basics guide which is focused on metadata removal.

dngray:
> What if we make this into a Security Basics guide which is focused on metadata removal. An intro into what is metadata and how to remove them (not only on images) would be great.

That would be a good idea.

ph00lt0:
@lx-is why are you not doing the MacOS version in shortcuts? You should be able to achieve the same there, and it is a little bit less complex.

lx-is:
> If you want me to update the PR with the explanation on how to do this let me know!

Yes, please do! That's much better.

docs/security/integrating-metadata-removal.md
```diff
@@ -0,0 +1,191 @@
...
+    brew install exiftool
+    ```
+
+3. [Tag](https://github.com/jdberry/tag/): a tool to manipulate tags on macOS.
+
+    ```bash
+    brew install tag
+    ```
+
+You can check if ExifTool is installed by running `exiftool -ver`. You should see a version number. Next, confirm that ExifTool is installed at `/opt/homebrew/bin/exiftool` by running `which exiftool`.
```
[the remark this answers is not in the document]
lx-is: I changed it to say:

> take note of where ExifTool is installed by running `which exiftool`.

and then later in the guide it says

> You should set your path to output of `which exiftool`.

ph00lt0:
> > @lx-is why are you not doing the MacOS version in shortcuts? You should be able to achieve the same there, and it is a little bit less complex.
> 
> I went ahead and added it. Your solution is much simpler, thank you.

Prefect! You don't need to enable allow running scripts btw, it will prompt you and it will ask if you want to allow it once of any time, same goes for access to documents.

docs/security/integrating-metadata-removal.md
```diff
@@ -0,0 +1,180 @@
...
+    done
+    return $@
+    ```
+
+    You should set your path to output of `which exiftool`.
+
+![macOS metadata removal shortcut](/assets/img/integrating-metadata-removal/shortcut-macos.png)
+
+<small>
+*Worth mentioning: the open source [ImageOptim](https://imageoptim.com/mac) app integrates into Finder's Services menu by default.*
```
ph00lt0, line 79: this quote feels a bit out of place

docs/security/integrating-metadata-removal.md
```diff
@@ -0,0 +1,180 @@
...
+</small>
+
+#### Enabling & using the Shortcut
+
+1. Ensure the shortcut is enabled by going to *System Preferences > Extensions > Finder* and enabling the shortcut you created.
```
ph00lt0, line 84: You do not need to do this tho, you can much simpler just enable it within the shortcuts app :)

I use also the clipboard input if you don't just the quick action in finder will be enough. Saves a few steps opening the settings app as well. 
image

ph00lt0 (changes requested):

ph00lt0 (approved):

ph00lt0:
@lx-is I just noticed in the quick actions menu there is also this convert image option which allows stripping of meta data.
image

Do you know if this is worse then the shortcut?

ph00lt0:
@dngray the screenshot is currently not the same as the code in the snippet. The return line includes `return $@` in the screenshot and `return "$@"` in the snippet. I am not even sure if you need to return at all. My version needed this to support clipboard input, but I don't think it's needed if you don't want to include that.

docs/security/integrating-metadata-removal.md
```diff
@@ -0,0 +1,166 @@
...
+     - Set the input to **Shortcut Input**
+     - Select **as arguments** for the pass input
+     - Leave **Run as administrator** unchecked
+
+6. Use the following as the body of the script:
+
+    ```bash
+    #!/bin/sh
+    PATH=/opt/homebrew/bin
+    exiftoolPath=$(which exiftool)
```
ph00lt0, line 61: This trick does not work and gives me an exemption. The path should be `/usr/local/bin/` using the path in the loop is not helping.

ph00lt0 (changes requested):

docs/security/integrating-metadata-removal.md
```diff
@@ -0,0 +1,166 @@
...
+    for f in "$@"
+    do
+        exiftoolPath -all= "$f";
+    done
+    return "$@"
```
ph00lt0, line 66: @dngray the screenshot is currently not the same as the code in the snippet. The return line includes `return $@` in the screenshot and `return "$@"` in the snippet. I am not even sure if you need to return at all. My version needed this to support clipboard input, but I don't think it's needed if you don't want to include that.

ph00lt0 (changes requested):

docs/security/integrating-metadata-removal.md
ph00lt0: Found a way to fix this: https://github.com/lx-is/privacyguides.org/pull/1/files

docs/security/integrating-metadata-removal.md
ph00lt0: Also removed in https://github.com/lx-is/privacyguides.org/pull/1 seems unnecessary.

lx-is:
> @lx-is I just noticed in the quick actions menu there is also this convert image option which allows stripping of meta data. image
> 
> Do you know if this is worse then the shortcut?

If it's anything like the Preserve Metadata option in Shortcuts then it's not as exhaustive. I haven't tried it myself though.

> Found a way to fix this: https://github.com/lx-is/privacyguides.org/pull/1/files

The script here is resulting in an error for me. Does it work for you? @ph00lt0

ph00lt0:
> The script here is resulting in an error for me. Does it work for you? @ph00lt0


@lx-is do you mean my latest? PR or what dngray wrote? What is the error you are getting?

lx-is:
> @lx-is do you mean my latest? PR or what dngray wrote? What is the error you are getting?

Yours and dng's aren't working for me. His doesn't error out, but it doesn't do anything either. With yours I get a generic "Run Shell Script finished with an error" message.

ph00lt0:
> Yours and dng's aren't working for me. His doesn't error out, but it doesn't do anything either. With yours I get a generic "Run Shell Script finished with an error" message.

If you get you should see a more detailed error in the shortcut app. Else what happens is you run these commands line by line? 

dng's didn't work for me either, because it won't execute from the wrong path. That's why I thought of using dirname to get the actual path, assuming that this would work on any set up.

lx-is:
> @lx-is If you get you should see a more detailed error in the shortcut app

error

ph00lt0:
How odd @lx-is, if you execute $(which exiftool) does it return "/opt/homebrew/bin/exiftool" or something different?

lx-is:
> How odd @lx-is, if you execute `$ which exiftool` does it return `/opt/homebrew/bin/exiftool` or something different?

`/opt/homebrew/bin/exiftool`

ph00lt0:
For me `dirname $(which exiftool) returns `/usr/local/bin` this is the path in which homebrew installs the tool, thus allowing it to work. I am surpised this doesn't for you as you used `/opt/homebrew/bin` also in your first version.

lx-is:
> For me `dirname $(which exiftool)` returns `/usr/local/bin` this is the path in which homebrew installs the tool, thus allowing it to work. I am surpised this doesn't for you as you used `/opt/homebrew/bin` also in your first version.

`dirname $(which exiftool)` returns `/opt/homebrew/bin` for me.

ph00lt0:
> > For me `dirname $(which exiftool)` returns `/usr/local/bin` this is the path in which homebrew installs the tool, thus allowing it to work. I am surpised this doesn't for you as you used `/opt/homebrew/bin` also in your first version.
> 
> `dirname $(which exiftool)` returns `/opt/homebrew/bin` for me.

Yeah that's right.
Could you try this as script maybe instead?

```bash
for f in "$@"
do
    $(which exiftool) -overwrite_original -all= "$f";
done
```

lx-is:
> This works for me and saves a line, this this works for you it would be awesome.

I get the same type of error :/. This is so frustrating haha

So far, this is still what's worked best *for me*:

```BASH
#!/bin/sh
PATH=/opt/homebrew/bin
for f in "$@"
do
	exiftool -all= "$f";
done
return "$@"
```

docs/security/integrating-metadata-removal.md
dngray: Merged into this PR

dngray:
The reason `dirname $(which exiftool)` doesn't work is because that returns the path without the binary, ie `/opt/homebrew/bin/`.

What  error do you get when you have just `$(which exiftool)`, perhaps its because there's more than one result?

dngray:
> ```shell
> for f in "$@"
> do
>     $(which exiftool) -overwrite_original -all= "$f";
> done
> ```

There's no need to evaluate the binary's path each run of the loop.

dngray:
Does this not work?

```bash
#!/bin/sh
absExiftool=$(which exiftool)

for f in "$@"
do
    "$absExiftool" -all= "$f";
done
return "$@"
```

dngray:
@tommytran732 that's not going to work `PATH=dirname $(which exiftool)`.

At very least it would need to be `PATH=$(dirname $(which exiftool))`

dngray:
Another simpler approach is to get rid of `PATH` or `absExiftool` the homebrew path should already be in a user's path.

ph00lt0:
> ```shell
> #!/bin/sh
> PATH=$(dirname $(which exiftool))
> for f in "$@"
> do
>   	exiftool -all= "$f";
> done
> return "$@"
> ```
>
> Do you get an error with either of these? I don't have a Mac so I can't check.

@dngray @tommytran732 this is the only one working for me.

tommytran732:
> > ```shell
> > #!/bin/sh
> > PATH=$(dirname $(which exiftool))
> > for f in "$@"
> > do
> >   	exiftool -all= "$f";
> > done
> > return "$@"
> > ```
> >
> > Do you get an error with either of these? I don't have a Mac so I can't check.
> 
> @dngray @tommytran732 this is the only one working for me. Also looks the safest, it worked for me without the $() but I agree it's better to do it like this. What currently is on the website definitely does not work.

I tested without the PATH and it works fine for me

PR #923, opened 2022-04-06
Add instructions for manually installing APKs

dngray:
Submitted via patch


dngray:
With this one I want to show an example perhaps with Secure Camera.

dngray:
#### From Matrix:

> Another option would be to install the APK directly the developers git repository releases page

should be:

> Another option would be to download the APK from the developer's git repository directly.

> and add it an RSS reader of your choice.

Should be:

> and add it to an RSS reader of your choice.

> You don't need to worry about slow updates that come from F-Droid or hope the developer has their own F-Droid repository.

Should be:

> That way, you don't need to worry updates being slow through F-droid or hope the developer has their own F-Droid repository.

dngray:
Another thing regarding this, we really have to advise people to check signatures ie `apksigner verify --print-certs app.apk` and verify the SHA-256 digest somewhere.

docs/android.en.md
```diff
@@ -328,4 +328,47 @@ To mitigate these problems, we recommend [Neo Store](https://github.com/NeoAppli
...
 
+
+## RSS Notifications
+
+If an app is released on GitHub you may be able to add a RSS feed to your [news aggregator](/news-aggregators) that will help you be aware of new releases. Using [Secure Camera](#secure-camera) as an example, you would navigate to the "releases" page on GitHub and append [`.atom`](https://github.com/GrapheneOS/Camera/releases.atom) to the URL.
+
+<figure markdown>
+  ![RSS Feed](./assets/img/android/gfeeds-light.png#only-light)
+  ![RSS Feed](./assets/img/android/gfeeds-dark.png#only-dark)
+</figure>
```
jonaharagon, line 339: Why is this a figure without a caption?
```suggestion
![RSS Feed](./assets/img/android/gfeeds-light.png#only-light)
![RSS Feed](./assets/img/android/gfeeds-dark.png#only-dark)
```

docs/android.en.md
```diff
@@ -328,4 +328,47 @@ To mitigate these problems, we recommend [Neo Store](https://github.com/NeoAppli
...
+
+### Verify APK fingerprint
+
+It may also be a good idea to verify the APK's signature, with the [`apksigner`](https://developer.android.com/studio/command-line/apksigner) tool which is a part of the Android [build-tools](https://developer.android.com/studio/releases/build-tools).
+
+<div class="annotate" markdown>
```
jonaharagon, line 345: Ahh, HTML tags in Markdown isn't the way to do this https://squidfunk.github.io/mkdocs-material/reference/annotations/

PR #924, opened 2022-04-06
Adjust VPN criteria and reduce focus on bitcoin

realguyman:
## Description

Resolves: #553

Monero is now required instead of Bitcoin for VPN criteria. Focus has been reduced for Bitcoin. All present VPNs meet the new criteria.


dngray:
Merged in https://github.com/privacyguides/privacyguides.org/commit/dc94c77dadc72b4f09e0488ea4240a45bf8d4fee

ph00lt0:
Why on earth would you require crypto trash for payments? Makes absolutely no sense. Most people who buy monero will not know how to do this in a private way anyways only causing you to get attention from certain organizations. Mailing in cash also isn't perfect but probably still better, and what is far better are the gift cards, like Mullvad offers. 

Additionally at no point will a VPN give you anonymity so why bother about the payment? Your ISP will know that you use this VPN service. Consider what you are trying to hide and how.

I guess it's nice you can pay with monero, but a requirement I really do not understand.

realguyman:
@ph00lt0 This can re-evaluated. If you take an issue with this, please _don't_ put your takes on this pull request.

If you read the issue that this pull request solves you'll see it was a temporary fix to reduce Bitcoin encouragement and remove the requirement for it.

As it stands Monero is a much better option to Bitcoin. We've yet to have a section covering cryptocurrency and anonymous methods of payment, which is a complex topic.

realguyman:
@ph00lt0 Sorry for my poor choice of wording. I have no intention of discouraging discourse about the matter. What I intended to encourage was for you (or someone else) to open a GitHub issue about the matter, so we don't clutter this pull request, which was intended to be a quick fix.

ph00lt0:
@realguyman no worries, it does probably makes sense to open an issue about this. But the requirements are not stated anywhere public atm or are they? I just felt it's strange to require this. If there would be a good reliable vpn provider without monero support it would just be weird not to include it. Monero is arguably better then bitcoin of course, but generally speaking we don't think a VPN makes you anonymous so the impact will be quite minimal.

ph00lt0:
@realguyman no worries, it does probably makes sense to open an issue about this. But the requirements are not stated anywhere public atm or are they? I just felt it's strange to require this. If there would be a good reliable vpn provider without monero support it would just be weird not to include it. Monero is arguably better then bitcoin of course, but generally speaking we don't think a VPN makes you anonymous so the impact will be quite minimal.

PR #926, opened 2022-04-06
Encrypted DNS: Should I use encrypted DNS update 

Razac-elda:
I updated the "Should I use encrypted DNS" section because I felt it wasn't very clear.

The flow chart is better suited here, I polished it up a bit and added a new choice. I think security and performance of 3rd party DNS should be mentioned since the table on the recommandation also talks about filters.

### DIFF
--- a/docs/technology/dns.md
+++ b/docs/technology/dns.md
@@ -19,9 +19,9 @@ Below, we discuss and provide a tutorial to prove what an outside observer may s
 
 1. Using [`tshark`](https://www.wireshark.org/docs/man-pages/tshark.html) (part of the [Wireshark](https://en.wikipedia.org/wiki/Wireshark) project) we can monitor and record internet packet flow. This command records packets that meet the rules specified:
 
-   ```bash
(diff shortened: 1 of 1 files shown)


realguyman:
@jonaharagon @dngray Sorry, I think I broke something. lol

dngray:
> I think security and performance of 3rd party DNS should be mentioned since the table on the recommandation also talks about filters

It's not really the correct. Correct how? enumeration of badness? It's not really security. The context of that flow chart is eavesdropping.

Razac-elda:
I would consider optional malwares filters as "increased security" in those cases where it is not feasible to implement a local solution. I do agree that it is slower than unencrypted, but my suggestion was based on the idea of recommending encrypted DNS when possible and if we combine that with filtering it might gain some performance.

> Regardless of whether or not the user is aware of immediate repercussions, I think this is bad advice.

DNS blocking was already described in the sentence below the chart with an approriate warning so I added it in the flow chart.

> We will update the article when something changes

Seems legit.

PR #953, opened 2022-04-08
Remove Social Networks & Social News Aggregators

tommytran732:
This PR removed Social Networks and Social News Aggregators as discussed in https://github.com/privacyguides/privacyguides.org/discussions/949

Legacy 3rd party PNGs and SVGs are also removed. Other legacy PNGs and SVGs are not removed yet, as I don't know if they are still being used or not.


dngray (approved):
Based on what I wrote in that discussion thread, I approve, but I am a bit on the fence on this one.

When we define "privacy" who is the adversary?

In regard to commercial platforms these tightly integrate with adtech empires, while it's possible smaller instances could do that, at present they do not. It'd be a shame to see mastodon go, when this does actually have E2EE on private messages too.

In regard to Lemmy, from what i've used of it, it's a lot faster than Reddit, (more like old.reddit.com). As much as I use Reddit, I really do hate the "new reddit", it's clunky and horrible. I can fully foresee one day the company behind that product ruining it completely.

dngray (changes requested):
I think we should definitely let everyone have an opportunity to speak on this before we merge/close.

tommytran732:
> Based on what I wrote in that discussion thread, I would approve, but I am a bit on the fence on this one.
> 
> When we define "privacy" who is the adversary?
> 
> In regard to commercial platforms these tightly integrate with adtech empires, while it's possible smaller instances could do that, at present they do not. It'd be a shame to see mastodon go, when this does actually have E2EE on private messages too.
> 
> In regard to Lemmy, from what i've used of it, it's a lot faster than Reddit, (more like old.reddit.com). As much as I use Reddit, I really do hate the "new reddit", it's clunky and horrible. I can fully foresee one day the company behind that product ruining it completely.

If you want we can keep mastodon as the only recommendation. Do keep in mind that it is only good because of e2ee messages. The primary functions (tweets, retweets, comments) are not private and can never be private.

dngray:
What I am saying is they do offer some privacy in respect to not having to go through hoops to create an account (you don't need a phone number for example).

Privacy in terms of content? well, it's not, because its publicly broadcast, so I don't think E2EE is really the only factor you can use to evaluate this, that works for platforms where you want to restrict distribution to a few parties, but that's about it.

realguyman:
I don't think we should remove _all_ of the social media platforms. I've written my piece on #949.

dngray:
In an effort to keep this clean, all future posts should be in the discussion thread.

realguyman:
PrivacyGuides has gone a re-write/re-birth, so I see little reason to not purge social media and such. We're not preventing anyone from recommending them again, so I think it's safe to remove them for now.

@dngray I'll remove your "requested changes" so we can merge the pull request, assuming everything is OK to go.

docs/tools.md
```diff
@@ -369,6 +341,14 @@ We [recommend](dns.md#why-should-i-use-encrypted-dns) a number of encrypted DNS
...
 </div>
 
+### Social Media
+
+<div class="grid cards" markdown>
+
+- ![Mastodon logo](/assets/img/social-media/mastodon.svg){ .twemoji } [Mastodon](https://joinmastodon.org/)
+
+</div>
+
```
jonaharagon, line 351: Missing a "learn more" link to the social media recommendations.

bkil:
I don't agree with this change (comments provided in the other issue)

PR #955, opened 2022-04-08
De-indent "Where to get your applications"

dngray:
I reordered some things that were said about F-Droid, and made Droid-ify a subheading.

I think this fixes https://github.com/privacyguides/privacyguides.org/pull/720#issuecomment-1092820636

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -340,15 +340,30 @@ The installation of GrapheneOS on a Pixel phone is easy with their [web installe
 
 A [CalyxOS membership](https://calyxinstitute.org/membership/calyxos) also entitles you to a device preloaded with CalyxOS.
 
-## GrapheneOS's App Store
+## Where to get your applications
+
+### GrapheneOS's App Store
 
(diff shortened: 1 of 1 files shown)


docs/android.md
```diff
@@ -340,15 +340,27 @@ The installation of GrapheneOS on a Pixel phone is easy with their [web installe
...
 
-### Droid-ify
+Sometimes the official F-Droid repository may fall behind on updates. F-Droid maintainers reuse package IDs while signing apps with their own keys, which is not ideal as it does give the F-Droid team ultimate trust. The Google Play version of some apps may contain unwanted telemetry or lack features that are available in the F-Droid version. The Google Play Store requires a Google account to login which is not great for privacy. The [Aurora Store](https://auroraoss.com/download/AuroraStore/) (a Google Play Store proxy) does not always work, though it does most of the time.
```
ph00lt0, line 353: Generally this section looks a lot better now indeed. Still thinking aurora should bet it's own header like F-droid and GrapheneOS's App Store. It doesn't feel logical to write about this under F-droid.

ph00lt0 (changes requested):

docs/android.md
dngray: I gave aurora store it's own heading, we don't really write much about it.

PR #965, opened 2022-04-10
Use dark mode logo when in dark mode (in readme)

lx-is:
Implemented the #gh-light-mode-only/#gh-dark-mode-only tags detailed in GitHub's [Basic formatting syntax](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to).

Unfortunately the theme tags have to be in the *a* tag, which means it'll be in the URL. GitHub did not implement this feature well at all. You could do this in markdown and avoid that, but then the image can't be centered. Or alternatively you could do it in the *img* tag, but you'd have to get rid of the *a* tag.

Preview: https://github.com/lx-is/privacyguides.org/blob/readme-logo-fix/README.md


dngray:
I didn't even know about that, yeah I don't visit github in dark mode much.

PR #987, opened 2022-04-12
Add changelog, deploy on tagged releases

jonaharagon:
From discussion #980 

This changes our deployment so that changes are **not** made on every push to `main`. The website is only updated when a release is published on GitHub. Hopefully it's better for people keeping track of changes to the website, it will also show a version number on the website. When we get translations going, having an easily visible version number/release schedule will make it easy to compare translations the main website.

This also adds a CHANGELOG.md file which ideally should be updated in every PR. We can then easily copy new info from that file to the release description when we deploy a new release.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,20 +1,12 @@
 # Additional Co-Owners are added to the TOP of this file.
 
 # High-traffic pages
-/index.html @JonahAragon @dngray
-/blog.html @JonahAragon
+/docs/index.md @JonahAragon @dngray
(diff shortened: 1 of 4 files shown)


lx-is:
I've seen some projects use automatic changelogs for releases. I'm not sure how it works, but could be worth looking into at least.

It would also be helpful to have some info in the wiki on how the CHANGELOG.md file should be formatted.

jonaharagon:
@lx-is I mentioned that in our team chat yesterday. So far with these last releases I've been liking the automatically generated release notes, so CHANGELOG.md may just be going away.

PR #993, opened 2022-04-12
Include Apple Mail on Privacy Tools page

lx-is:


jonaharagon (approved):
Now I have to hunt down who added Apple Mail without doing this 😠 

lol, thanks

jonaharagon:
Actually, we should maybe decide whether we want to list native operating system tools on this page @privacyguides/team. For example, we don't list Bitlocker/FileVault on this page even though they are recommended for OS encryption. I don't know if this is a similar situation.

PR #1003, opened 2022-04-13
Improvements to Privacy Tools page

elitejake:
## Description

- Make use of footnotes instead of `*`
- Use right arrow on `Learn more...` links


lx-is:
I wonder if it would be possible to implement an icon as a link to the footnote. As they stand right now, the links are hardly visible.

I think a little `i` info icon would look nice, though I'm not sure if it's even possible to link an icon to a footer note.

Maybe this would work?

```md
[:material-information-outline:]([^1)
```

realguyman:
While I do like footnotes, there isn't enough contrast between the two links, which causes it to blend in:

image

image

Ideally links should be underlined for accessibility. This is one of the cases where it would be extremely beneficial to differentiate between separate links.

---

Why did you add arrows to the "Learn more ..." links? I'm not sure if it was needed, or improves anything significantly. Is this just for aesthetics?

lx-is:
> Why did you add arrows to the "Learn more ..." links? I'm not sure if it was needed, or improves anything significantly. Is this just for aesthetics?

I'm a fan of that addition personally. Nothing wrong with spending time on aesthetics 😛

flexagoon:
I think you shouldn't use footnotes. That's some pretty important information, it would probably be better if it was closer to the actual recommendation card. Because when you use footnotes, there's a high chance that the user will not notice it or think that it's something not important

lx-is:
Yeah I'm still mixed on the idea of footnotes for this too. If I'm being honest, I'm a bit mixed on the overview page as a whole. I feel it disregards a lot of important info that people should be seeing but most likely aren't. I'd personally like to see all the tools link to their card on pg, not the tool's homepage.

jonaharagon:
I'm of the same opinion as @flexagoon, I think footnotes are more suited for linking sources and providing additional clarification, but not for important explanatory content (which is why I didn't use them originally). 

I'm mixed on the linking to homepages vs linking to cards thing, I could see utility for either decision. 

I do like the arrows on the learn more links.

docs/tools.md
```diff
@@ -28,13 +28,13 @@ For your convenience, everything we recommend is listed below with a link to the
 
 - ![uBlock Origin logo](/assets/img/browsers/ublock_origin.svg){ .twemoji } [uBlock Origin](https://github.com/gorhill/uBlock)
 - ![AdGuard logo](/assets/img/browsers/adguard.svg){ .twemoji } [AdGuard for Safari](https://adguard.com/en/adguard-safari/overview.html)
-- ![ToS;DR logo](/assets/img/browsers/terms_of_service_didnt_read.svg){ .twemoji } [Terms of Service; Didn't Read\*](https://tosdr.org/)
+- ![ToS;DR logo](/assets/img/browsers/terms_of_service_didnt_read.svg){ .twemoji } [Terms of Service; Didn't Read](https://tosdr.org/)[^1]
 
-</div>
+[^1]: We do not recommend installing ToS;DR as a browser extension. The same information is provided on their website.
```
jonaharagon, line 33: This text should still be outside of the `div` element.

jonaharagon (changes requested):
I'm not sure about this change. I wonder if annotations make more sense than footnotes: https://squidfunk.github.io/mkdocs-material/reference/annotations/

realguyman:
@jonaharagon I checked out the annotations feature and it seems to work wonderfully. Without JS, too. Seems to be an experimental feature, though.

jonaharagon:
I think all "insiders" features are marked experimental, but we make heavy use of them in this project.

jonaharagon:
I'm splitting this into two separate PRs, #1015 with your right arrow enhancement, and #1016 replacing the current notes with annotations; I'm not into the footnotes solution.

Thanks :)

elitejake:
I wasn't into footnotes solution myself, but `*` was really ugly, IMO. Anyway, I completely agree with replacing it with annotations.

Issue #908, opened 2022-04-05
Writing style guidelines


dngray:
To make Privacy Guides a high quality resource I propose that we have some guidelines, that how content should be presented.

I like having advanced articles of the site, but I think needs to be a bit like an onion, in respect to being able to peel off entry level layers and get some usefulness out of it without being bombarded with very in-depth specifics that might turn off new readers.

This would allow for entry level articles to self cite more complex reasoning, for issues discussed.

I described that in https://github.com/privacyguides/privacyguides.org/issues/834#issuecomment-1087354314

We might want to get a guideline, this is a good one I think https://www.gov.uk/guidance/content-design/writing-for-gov-uk.

I think in general this would be an improvement, as it would prevent highly specific information being on the surface https://github.com/privacyguides/privacyguides.org/pull/904/files/7be16078d8d7e6fb8eb0469fb59da210dac7dbc1#r842391988

This issue is one i want to get codified before we begin outreach for translations https://github.com/privacyguides/privacyguides.org/discussions/30

I think it will help translators a lot.

tommytran732:
The page is extremely inconsistent. Should additional information be inside of the cards like the Android or Linux page or by inside of the cards like the rest of the site?

I think its best to just do it like the Android/Linux page, so we can have a brief discussion in the card, but more in depth information can be put outside the card and the user can ignore what's outside of the cards if they are not interested.

freddy-m:
Would this require a style guide?

jonaharagon:
> I think its best to just do it like the Android/Linux page

Yes 👍

jonaharagon:
https://www.plainlanguage.gov/guidelines/ is a similar set of guidelines in American English.

elitejake:
Also check out Wikipedia''s [Manual of Style](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style).

Razac-elda:
After a quick look at those guidelines I think @elitejake found the right one, the UK and US ones are good for general article advice. Wikipedia introduces concepts for more advanced styles and, being a widley popular website, users will find it more familiar to read.

dngray:
Another one that wouldn't hurt is https://docs.microsoft.com/en-us/style-guide/procedures-instructions/writing-step-by-step-instructions

elitejake:
We shouldn't follow these guidelines as-is, because they don't make use of Markdown. 

[...]

PR #1006, opened 2022-04-13
Clarify macOS DNS instructions and reword a sentence

flexagoon:
## Description

Changes the macos dns settings instructions. Before the change, it looked like a single settings path, not like two separate ones.

This PR also rewords a sentence to be more correct

### DIFF
--- a/docs/dns.en.md
+++ b/docs/dns.en.md
@@ -37,7 +37,7 @@ The criteria for the servers listed above are:
 
 ### Android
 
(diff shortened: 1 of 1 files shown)


dngray:
Took a look at this I've made some changes as we've kinda developed a de-facto style guide on https://docs.microsoft.com/en-us/style-guide/procedures-instructions/writing-step-by-step-instructions was discussed in https://github.com/privacyguides/privacyguides.org/issues/908

PR #1011, opened 2022-04-13
Remove CHANGELOG.md

jonaharagon:
This is just additional work, I didn't realize GitHub would automatically generate such changelogs with our new release strategy :)

We just have to make sure going forward that PRs are titled appropriately/clearly since they are used when auto-generating release notes.

See [releases](https://github.com/privacyguides/privacyguides.org/releases) for an example of the type of generated release notes we'd just use instead.


realguyman:
I forgot about this feature. I couldn't seem to get it to work on my personal projects, though. Seems to be working just fine. :)

jonaharagon:
@realguyman I believe it only generates based on PRs, so it works here since we have a PR heavy workflow.

PR #1027, opened 2022-04-14
Support asset localization

jonaharagon:
This uploads every file with `.en.` in the name to Crowdin, not just markdown files.


dngray:
LGTM i checked the images dir and couldn't find any others.

Issue #1023, opened 2022-04-14
openSUSE automated testing


BobIsMyManager:
## Description
openSUSE is tested through [openQA](https://openqa.opensuse.org/), the open build service is used for automated packaging/distribution of software.
**URL of affected page:**
https://www.privacyguides.org/linux-desktop/#opensuse-tumbleweed

BobIsMyManager:
I think there was a misunderstanding, the openSUSE Build Service is **not** involved at all when it comes to testing, all it does is package/distribute software and the snapshots of openSUSE. The actual testing of the snapshots is done by openQA alone.

PR #1040, opened 2022-04-16
Removed OBS as it is not involved with testing

BobIsMyManager:
Resolves:    #1023


BobIsMyManager:
OBS can still be mentioned when talking about reproducable packaging if you still want to mention OBS

dngray:
Ooops.

PR #1052, opened 2022-04-19
Flawed E2EE in Web Applications

realguyman:
There are countless E2EE web applications in today's world despite all of the warning and flaws pointed out in old research, often dismissed as "out of scope". I'm currently working on an article that will discuss and explain why E2EE in web applications is flawed, and why we should move away from them where applicable.

Especially with critical data (e.g. passwords, credit card information, etc.). The article isn't complete yet, but already covers the most critical information. A lot of clean up and other potential information may need to be provided.

I'll be working on this as I have time to. Please leave reviews and discussions in this PR.

### DIFF
--- a/docs/security/web-e2ee.en.md
+++ b/docs/security/web-e2ee.en.md
@@ -1,0 +1,28 @@
+---
+title: "E2EE in Web Applications"
+---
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
@realguyman the past some Ytbers recommended to use progressive web apps on smartphones because it will decrease the number of applications on your deceive and therefore it will decrease the number of applications that could be compromised.

After reading this i think that statement is somewhat flawed but i would like your opinion on it.

bkil:
What are you trying to prove with this post? Web (-Browser based) Apps, especially if FOSS and engineered from the ground up for offline use via Service Workers, are for most intents and purposes isomorphic to non-browser based apps. It's just another containerization technology and runtime. 

All the same kind of attacks would apply to desktop or native mobile apps that support live (or at least daily) updating through its own servers (or its own stores). It's TOFU either way.

At the same time, nothing keeps you from self-hosting the web apps you use either on a trusted server or even on localhost and review all code changes before running them.

dngray:
We don't have permission to push to this branch, you must have disallowed that.

I've merged this with a branch on the main repository so that we can rebase it.

Future developments will be in https://github.com/privacyguides/privacyguides.org/pull/1342

PR #1053, opened 2022-04-19
Split Android page into separate articles and update device recommendations

### LINKED ISSUE #1049: Changing Android recommendations
### Discussed in https://github.com/privacyguides/privacyguides.org/discussions/928

<sup>Originally posted by **tommytran732** April  6, 2022</sup>
Historically, there were a few reasons why one would want to use CalyxOS over GrapheneOS:
- GrapheneOS didn't have support for either the Sandboxed Play Services or MicroG, and CalyxOS supported MicroG. This made Calyx a more attractive option from a usability perspective.
- CalyxOS supported legacy devices like the Pixel 2 series for longer than GrapheneOS
- CalyxOS was pretty much the only good custom OS for the Xiaomi A2

These days, a lot have changed:
- GrapheneOS now has the Sandboxed Play Services. This approach is a lot more secure than MicroG as it doesn't require a security bypass (the signature spoofing patch) or any more privileges than a regular app. MicroG on the other hand still needs to run as a system app and be granted way more privileges than a normal user app for full functionality. From a usability perspective, the Sandboxed Play Services is also a lot better - it supports more APIs and other services like the In-app billing API or Google Play Games.
- CalyxOS no longer supports the Pixel 2 series or the Xiaomi A2

This, combined with the fact that CalyxOS does not have nearly as much hardening done as GrapheneOS and the fact that they did fall behind on security updates for months, leaves no compelling reason to use CalyxOS on a Pixel device.

In the future, Calyx is looking to add support for the Fairphone 4, Oneplus 8T and OnePlus 9. They have already released test builds for those devices. There is quite a lot to unpack here:

The Fairphone 4 is an **extremely** insecure device. On top of the fact that they are using an outdated SOC that will go EOL in < 2 years from now (so the whole sustainability claim really doesn't make much sense because there will be no security update for the firmware), they quite literally use the AVB test keys to do verified boot, making it completely useless, unless a custom key is enrolled https://forum.fairphone.com/t/bootloader-avb-keys-used-in-roms-for-fairphone-3-4/83448/4

They have had this issue in the Fairphone 3. They said they fixed it in the Fairphone 4. Turns out, they didn't!!!

That leaves us with the OnePlus 8T and 9. I don't know of any major deficiencies with these devices yet (someone told me that they do not clear their RAM after a reboot, but I have yet to see any proof of this), so at least they are still better than the Fairphone.  Both of these Oneplus phones and the Fairphone do not have any Secure Element however, so they will be relying on their TEE to store secrets and do rate limiting, which is less secure than what we have on the Pixel.

Here is what I am proposing:
1. Mention that GrapheneOS outclasses CalyxOS in both security and usability and if the user has a Pixel, they should get GrapheneOS. There is no compelling reasons to use any other custom OSes.
2. Mention that the Fairphone is extremely insecure and that the OnePlus phones are still less secure than a Pixel. If the user has those phones (**especially the Fairphone**), they should try to sell them and buy a Pixel phone. If they are unable to, then fall back to use CalyxOS. If the user can trade the Fairphone 4 for a Oneplus 8T or 9, they should absolutely do so.
3. Give similar recommendations for DivestOS phones. If the user has an EOL phone or some other phones that they can sell to get a modern Pixel, they should do it. If they can't then use DivestOS.
4. Note that custom OSes like /e/, despite booting in green state on the Fairphone, have completely useless verified boot and there is no system integrity protection whatsoever. In fact, anything booting in green state on the phone is not to be trusted. The yellow state is still okay as far as I know.
5. And to reiterate how broken this is - if the user has a Fairphone, they should absolutely be installing CalyxOS or any other custom OS with verified boot support. The stock OS cannot be trusted either because verified boot with it is broken as mentioned above. If we don't want to pollute our Android page with this, I suggest an "Insecurities" category just to detail all of this out, or link to relevant blogs discussing it.
We could do it for 
- The fairphone
- F-Droid (link to Wonderfall's blog)
- LineageOS (link to Madaidan's blog)
- /e/ OS (link to Madaidan's blog and @SkewedZeppelin 's comment on our discussion page
- Ungoogled Chromium (link to duck's blog)</div>

jonaharagon:
Started working on #1049


docs/android.en.md
```diff
@@ -72,41 +71,74 @@ DivestOS 16.0, 17.1, and 18.1 implements GrapheneOS's [`INTERNET`](https://devel
...
+
+If you are unable to obtain a Google Pixel, recent OnePlus devices provide a good balance of security with custom operating systems and longevity, with OnePlus 8 and later devices receiving 4 years of security updates. CalyxOS has [experimental support](https://calyxos.org/news/2022/04/01/fairphone4-oneplus8t-oneplus9-test-builds/) for the **OnePlus 8T** and **9**.
+
+DivestOS has support for most OnePlus devices up to the **OnePlus 7T Pro**, with varying levels of support.
+
+#### Fairphone
+
+!!! danger
+
+    Out of the box, Fairphone devices are incredibly insecure. [Fairphone's stock bootloader trusts the public AVB signing key](https://forum.fairphone.com/t/bootloader-avb-keys-used-in-roms-for-fairphone-3-4/83448/10), meaning any system can be installed and the phone will trust it as if it is the stock system. This essentially breaks verified boot on a stock Fairphone device.
```
SkewedZeppelin, line 133: this should probably be comment 11 instead of 10

jonaharagon:
```suggestion
    Out of the box, Fairphone devices are incredibly insecure. [Fairphone's stock bootloader trusts the public AVB signing key](https://forum.fairphone.com/t/bootloader-avb-keys-used-in-roms-for-fairphone-3-4/83448/11), meaning any system can be installed and the phone will trust it as if it is the stock system. This essentially breaks verified boot on a stock Fairphone device.
```

dngray:
I'm thinking we shouldn't mention:

> A [CalyxOS membership](https://calyxinstitute.org/membership/calyxos) also entitles you to a device preloaded with CalyxOS.

Because looking at that link it seems to be $600 for Pixel 4a or Pixel 4a 5G, that seems like bad value to me these days.

PR #1061, opened 2022-04-21
Fix Datura firewall link

efb4f5ff-1298-471a-8973-3d47447115dc:


docs/android.en.md
```diff
@@ -37,7 +37,7 @@ Currently, only [Pixel phones](https://grapheneos.org/faq#device-support) meet i
 
     **CalyxOS** is a decent alternative to GrapheneOS.
 
-    It has some privacy features on top of AOSP, including [Datura firewall](https://calyxos.org/docs/tech/datura-details), [Signal](https://signal.org) integration in the dialer app, and a built in panic button. CalyxOS also comes with firmware updates and signed builds, so [verified boot](https://source.android.com/security/verifiedboot) is fully supported.
+    It has some privacy features on top of AOSP, including [Datura Firewall](https://calyxos.org/docs/tech/datura-details), [Signal](https://signal.org) integration in the dialer app, and a built in panic button. CalyxOS also comes with firmware updates and signed builds, so [verified boot](https://source.android.com/security/verifiedboot) is fully supported.
```
jonaharagon, line 40: I usually see this just referred to as Datura, is "Firewall" part of the proper name anywhere in the app, maybe this change would be more appropriate:
```suggestion
    It has some privacy features on top of AOSP, including [Datura](https://calyxos.org/docs/tech/datura-details) firewall, [Signal](https://signal.org) integration in the dialer app, and a built in panic button. CalyxOS also comes with firmware updates and signed builds, so [verified boot](https://source.android.com/security/verifiedboot) is fully supported.
```

efb4f5ff-1298-471a-8973-3d47447115dc: yeah that makes more sense!

PR #1067, opened 2022-04-21
Recommend a custom Android OS

jonaharagon:
Recommend a custom OS on the Android overview page and link to our OS recommendations.

### DIFF
--- a/docs/android/overview.en.md
+++ b/docs/android/overview.en.md
@@ -2,9 +2,19 @@
 title: Android Overview
 icon: material/cellphone-check
 ---
-Android is a secure operating system that has strong [app sandboxing](https://source.android.com/security/app-sandbox), [Verified Boot](https://source.android.com/security/verifiedboot), and a robust [permission](https://developer.android.com/guide/topics/permissions/overview) control system.
+Android is a secure operating system that has strong [app sandboxing](https://source.android.com/security/app-sandbox), [Verified Boot](https://source.android.com/security/verifiedboot) (AVB), and a robust [permission](https://developer.android.com/guide/topics/permissions/overview) control system.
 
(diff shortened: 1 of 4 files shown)


tommytran732:
I disagree with the reasoning. The problem here is not about auditing or tweaks by the OEM. The bloatware are not a problem either if they are not treated as system applications with invasive integration.

The primary problem with the stock OSes is that they give the Play Services an excessive amount of privileges and the user have no option to opt out of them. This leads to a larger attack surface, and the fact that Play Services cannot be disabled and that they have access to hardware identifiers (especially when they don't need them) is worrisome.

dngray:
> This is the primary reason why we recommend custom OSes without such invasive integration in the first place.

I also don't want to be recommending replacing an OEM OS with one that breaks verified boot.

Better off [disabling apps](https://www.digitaltrends.com/mobile/disable-apps-android/) than trying to remove them with adb, debloating tools etc. The reason is you can't modify the system image really that way and the apps will reappear as soon as a new system image is pushed to the device.

Some OEMs don't let you disable apps, I honestly think in that scenario users should just steer clear. There is only so much that can be done and if an OEM is that hostile to user control, they shouldn't be trusted.

jonaharagon:
> I also don't want to be recommending replacing an OEM OS with one that breaks verified boot. 

We don't though? *confusion*

tommytran732:
> > I also don't want to be recommending replacing an OEM OS with one that breaks verified boot.
> 
> We don't though? _confusion_

The problem is the language being used isn't clear and the reasoning is flawed. Who audits the changes made by GrapheneOS, CalyxOS or DivestOS?

tommytran732:
> In addition to the privacy issues with Google Play Services, the "stock" operating system that comes pre-installed on your phone often has "bloatware" such as sponsored games, social media apps, and carrier-installed applications which in some cases cannot be entirely removed. OEMs often make additional modifications to the system ranging from minor theme tweaks to major operating system changes, which are nearly all proprietary modifications that go un-audited.

> If your phone supports it, the best way to get around all of these issues is to install a reputable custom Android-based operating system which replaces the stock system entirely. The custom operating systems we recommend are fully open-source and support AVB just like the stock installation, so your phone's security is maintained.

PR #1079, opened 2022-04-22
Use "set up", instead of "setup"

taivlam:
Use "set up" as a verb, instead of "setup" as a noun

### DIFF
--- a/docs/linux-desktop.en.md
+++ b/docs/linux-desktop.en.md
@@ -48,7 +48,7 @@ Tumbleweed follows a rolling release model where each update is released as a sn
 
 Arch Linux has a rolling release cycle. There is no fixed release schedule and packages are updated very frequently.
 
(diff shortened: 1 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
what about the other pages that uses the word setup? U can typ it in the searchbar on the site and will see all pages that includes this word

jonaharagon (approved):

taivlam:
I only noticed this in one of the subheadings but didn't go through the rest of the page.

I'd only wholesale replace the word across the whole site **with** double checking, because this requires looking at the context for correct usage - but at least for this page it looks like Jonah already fixed all occurrences.

jonaharagon:
I checked the whole site and combined it with your PR, so AFAIK we should be good :)

PR #1081, opened 2022-04-23
Fix capitalization of KeePassXC and KeePassDX

elitejake:
I thought this was fixed at #920.


jonaharagon:
Hm I guess only on the tools page, I better check the other pages "fixed" in that PR quick lol

jonaharagon (approved):
thx

PR #1092, opened 2022-04-23
Reduce the number of admonitions across the site

jonaharagon:
We don't want to over-use them, because they should specifically call out important information. We have a number of (especially "notes" and "info") admonitions that could just as easily be plain-text, because their content isn't particularly more important than other content on the page.

### DIFF
--- a/docs/browsers.en.md
+++ b/docs/browsers.en.md
@@ -8,8 +8,6 @@ These are our current web browser recommendations and settings. We recommend kee
 
 ### Tor Browser
 
-!!! anonyimity "This product provides anonymity"
-
 !!! recommendation
 
     ![Tor Browser logo](assets/img/browsers/tor.svg){ align=right }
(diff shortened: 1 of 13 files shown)


lx-is (commented):
What's the rationale behind moving some of the call-outs above the recommendation card? (e.g. Apple Mail and Tahoe LFS)

jonaharagon:
Basically I have it so that notes about _choosing whether to use_ the recommendation are above the card (i.e. you don't probably want to use Apple Mail on iOS, you probably don't want to use Tahoe-LAFS unless you know what you're doing), and notes about actually _using_ the recommendation are below the card (i.e. [a note about JS E2EE](http://localhost:8000/productivity/#cryptpad), or a [warning about potential IP leaks](http://localhost:8000/video-streaming/#freetube)).

I don't know if that makes sense, does it? Maybe it just does to me :)

lx-is:
Hm, yeah that makes sense. I don't really like how it pushes the card away from the header, but I can't think of a better way it could be done. Good change.

Issue #1094, opened 2022-04-23
Replace the "Visit website.example" buttons with "Visit Homepage" or "Visit Repository"


jonaharagon:
Kinda related but kinda not — I was thinking what if we replaced the "Visit website.example" purple buttons with "Visit Homepage" and "Visit Repository" instead. 

This would make the buttons more visually consistent with the privacy policy buttons, as well as solve the issue when a URL is really long or unappealing (e.g. the buttons on the [Magic Wormhole](https://www.privacyguides.org/file-sharing/#magic-wormhole) and [git-annex](https://www.privacyguides.org/file-sharing/#git-annex) cards).

_Originally posted by @lx-is in https://github.com/privacyguides/privacyguides.org/issues/1093#issuecomment-1107663538_

jonaharagon:
Yes, this is a change I wanted to make after switching to mkdocs.

I wonder if `Visit Homepage` or `Homepage` or `Project Homepage` is better.

lx-is:
> I wonder if `Visit Homepage` or `Homepage` or `Project Homepage` is better.

My initial vote would go towards just "Homepage" but that might make the button look too short. I guess I'd have to see the options visually.

flexagoon:
This will also make translation much easier 😉

dngray:
I think this one is more a less finished if you want to review.

PR #1095, opened 2022-04-23
Update primary button text

jonaharagon:
Resolves: #1094 

Preview: https://github.com/privacyguides/privacyguides.org/pull/1095#issuecomment-1107909848

### DIFF
--- a/docs/android.en.md
+++ b/docs/android.en.md
@@ -27,7 +27,7 @@ We recommend installing one of these custom Android operating systems on your de
 
     GrapheneOS provides additional [security hardening](https://en.wikipedia.org/wiki/Hardening_(computing)) and privacy improvements. It has a [hardened memory allocator](https://github.com/GrapheneOS/hardened_malloc), network and sensor permissions, and various other [security features](https://grapheneos.org/features). GrapheneOS also comes with full firmware updates and signed builds, so verified boot is fully supported.
 
-    [Visit grapheneos.org](https://grapheneos.org/){ .md-button .md-button--primary } [Privacy Policy](https://grapheneos.org/faq#privacy-policy){ .md-button }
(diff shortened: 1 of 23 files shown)


lx-is (changes requested):
I think having "Repository" instead of "Source Code" would be better.

The latter might sound too technical/irrelevant for some people, thus discouraging them from clicking the link. More often than not, repositories house important non-code related information such as wikis and a faq, with the readme sometimes even serving as the project's homepage.

jonaharagon:
@lx-is I actually figured the opposite, "repository" is dev-speak, while more people are probably familiar with the idea of "source code". 

I wonder if we should change the links to the project's README specifically if they are only on a Git repo, and say "View README" or "View Info" instead.

lx-is:
Huh, interesting. I thought repository was a pretty mainstream term nowadays.

I just feel "source code" implies only the raw code, meant for auditing and such. Maybe just me.

I think "___ Info" could work, with the blank being replaced with project, product, software, or whatever.

elitejake:
Or the button text should say `Visit Homepage` instead of `Source Code`. Because it is technically the project's home page.
<hr>

If we _are_ sticking to `Source Code`, I'd say `View Source` would be better,

dngray:
I've gone ahead and changed the others, merge when you want @jonaharagon

elitejake:
@dngray what's the difference between `Visit Website` and `Visit Homepage`?

lx-is:
Should "Visit" even be used at all? Having a verb isn't consistent with the other buttons

dngray:
> @dngray what's the difference between `Visit Website` and `Visit Homepage`?

I used Homepage when the product was a project site, I used Website, when it was a service

PR #1104, opened 2022-04-26
remove CTemplar

namazso:
https://ctemplar.com/ctemplar-is-shutting-down/


dngray:
Hmm does seem to be the case.

PR #1107, opened 2022-04-26
Add link to PrivateBin instances list

elitejake:


jonaharagon:
Does this change add much, since the directory is a click away from the homepage anyways?

PR #1115, opened 2022-04-27
Remove icon from external source

elitejake:
The icon is not necessary and it is loaded from an external source. So, I removed it.

### DIFF
--- a/docs/browsers.en.md
+++ b/docs/browsers.en.md
@@ -90,7 +90,7 @@ The [Firefox Sync](https://hacks.mozilla.org/2018/11/firefox-sync-privacy/) serv
 
 #### Extensions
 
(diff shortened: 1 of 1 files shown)


jonaharagon:
Yikes, kind of annoyed that any of these are external actually, since the other icons are not. *Investigating*

jonaharagon:
I would rather wait and see if https://github.com/squidfunk/mkdocs-material/issues/3851 can be resolved, since we use more emojis than just this one.

elitejake:
You're right. It totally slipped my mind that this was a bug in the Privacy plugin.

jonaharagon:
Closing this. Best case the Privacy plugin is updated. Worst case scenario I'd rather put together a PR that manually adds all the emojis we use the same way as our other custom icons, instead of this. Which would be annoying, but I'd rather not go without emojis :)

Issue #1109, opened 2022-04-26
Disroot no longer offers email aliases


pm4rcin:
## Description
On the site where should be a form to apply there is information that it was disabled and checking on web archive it is in that state for at least one year https://web.archive.org/web/20210116063329/https://disroot.org/en/forms/alias-request-form
Here's the live version: https://disroot.org/en/forms/alias-request-form
**URL of affected page: https://www.privacyguides.org/email/#disroot**

Issue #1075, opened 2022-04-22
Disroot no longer supports 2fa on Email, only Cloud


Dyrimon:
## Description
Disroot has disabled their 2fa on email when they switched from rainloop to SnappyMail

**URL of affected page:**
https://www.privacyguides.org/email/#disroot

The affected page declares Disroot supports TOTP 2fa on email, but it's no longer true. As a disroot user myself, I've noticed this when I bypassed 2fa on login. The [FAQ page](https://disroot.org/en/faq) on 2fa says that now it only supports 2fa on cloud. https://howto.disroot.org/en/tutorials/cloud/settings#two-factor-authentication

And this change does not meet the minimum to qualify criteria for Email category according to the PG criteria.

jonaharagon:
@dngray can you check this?

darhma:
The 2fa on email with IMAP/SMTP does not make much sense as explained here: https://github.com/the-djmaze/snappymail/issues/84#issuecomment-817852617

dngray:
Ill check with them to see if this is temporary, or if they're planning to support that.

jonaharagon:
> The 2fa on email with IMAP/SMTP does not make much sense

The benefit of it is that webmail often has account management functionality in addition to just email capabilities. If you can do things like change your password via webmail, for example, 2FA does still provide benefit.

muppeth:
We have decided to remove 2FA on webmail during the switch to snappymail yesterday because it was causing confusion for people (verification code field was ever present during login even for those not using 2FA which caused ton of questions). This was also good opportunity to remove it because it only created illusion of security for people since you could just use IMAP to circumvent it. We have enabled it because of the pressure of some users in the past and I think it was a mistake to do it. 
That said, we want to actually work on service wide 2FA solution to provide truly two factor solution for the entire platform. 

https://disroot.org/en/blog/2022-04-snappymail

darhma:
> The benefit of it is that webmail often has account management functionality in addition to just email capabilities. If you can do things like change your password via webmail, for example, 2FA does still provide benefit.

I use disroot as my mail provider and in the new webmail interface there doesn't seem to be an option to change the password.

Dyrimon:
> I use disroot as my mail provider and in the new webmail interface there doesn't seem to be an option to change the password.

Indeed, I just noticed that the only security functions in the "Security" section of the webmail is PGP related only.

Dyrimon:
[...]

PR #1117, opened 2022-04-27
Remove Disroot email recommendation

jonaharagon:
Unfortunately, Disroot does not meet our minimum criteria for recommendation:

- Protection of webmail with 2FA, such as TOTP. (#1075)

I considered removing our 2FA requirement instead of this PR, as it was pointed out that webmail 2FA provides little additional security when IMAP/SMTP does not allow for two-factor authentication anyways. However, Disroot also does not meet a number of our best practices for recommendation, including:

- Encrypts email account data at rest.
- Support for hardware authentication, ie U2F and [WebAuthn](https://en.wikipedia.org/wiki/WebAuthn).
- Offering email aliases. (#1109)
- Accepting Monero for purchases.

Additionally, removing Disroot allows us to increase our minimum security requirement for email storage from "encrypted" to "encrypted with zero-access encryption", a change which has also been made in this PR. Every other recommended provider utilizes (or in the case of mailbox.org, *can optionally utilize*) zero-knowledge encryption for email storage.

In light of all of this, I think the best course of action is to not recommend their email services. Ultimately I don't think Disroot provides a _bad_ service, I just believe our four other recommendations are all objectively _better_ from a privacy and security standpoint.

Closes: #1109 Closes: #1075


freddy-m (approved):
LGTM but I'll let others check over this before merging

dngray (dismissed):
> - Encrypts account data at rest.

This was always filesystem/full disk encryption, and not specifically on a per email/inbox basis. This was the reason why Mailfence was removed: https://github.com/privacytools/privacytools.io/issues/1908 I emailed them on the 9 Mar 2022 for an update and they said it still wasn't ready yet.

> - Support for hardware authentication, ie U2F and [WebAuthn](https://en.wikipedia.org/wiki/WebAuthn).

@muppeth [mentioned](https://github.com/privacyguides/privacyguides.org/issues/1075#issuecomment-1106814174) they switched to SnappyMail rather quickly because of a security issue with Rainloop. There is a plugin for 2FA with SnappyMail (apparently according to this page https://snappymail.eu/comparison.html), so I wanted to know if there was any plan to use that.

We do allow providers who allow you to disable IMAP/SMTP and use webmail only as with the situation of Mailbox.org mentioned here: https://github.com/privacyguides/privacyguides.org/issues/1075#issuecomment-1106835446 StartMail uses a similar approach, to Mailbox.org with a "[trusted IMAP password](https://support.startmail.com/hc/en-us/articles/360006536217-Configure-a-trusted-device-IMAP-password)".

I'm overly filled confidence from the author of SnappyMail describing 2FA+IMAP as "fake security", https://github.com/the-djmaze/snappymail/issues/84#issuecomment-817852617. 

There are ways to provide 2FA and also IMAP support, SASL XOAUTH2 etc. It would be nice to see more providers support this method as many clients have implemented it, because the large providers, (Google Outlook, Yahoo), already do use it.

One point that I have considered is perhaps we should adjust the 2FA requirement to be not necessary. The reason is currently all providers that provide 2FA either require you to use their app, (tutanota) bridge, (protonmail, and you must give them an encrypted copy of your PGP key) or webmail (mailbox.org).

We do know these approaches do come at the cost of *some security* as discussed in https://github.com/privacyguides/privacyguides.org/pull/1052, so by requiring MFA we may actually be recommending less secure approaches.

In other words you're trading authentication security for private key security.

> - Offering email aliases. (#1109)

This was always a "best case" scenario https://www.privacyguides.org/email/#technology and never something we required. It depends on a person's threat model, because you can combine your service with a [cloaking service](https://www.privacyguides.org/email/#email-cloaking-services) to achieve privacy.

> Accepting Monero for purchases.

They do accept Monero https://disroot.org/en/cryptocurrency.

jonaharagon:
I think I addressed all the points you mentioned in the original post. As far as Monero:

CleanShot 2022-04-28 at 08 03 49@2x

https://disroot.org/en/forms/domain-linking-form

> [...] removing Disroot allows us to increase our minimum security requirement for email storage from "encrypted" to "encrypted with zero-access encryption", a change which has also been made in this PR.

I think this is the most important change here, this should probably be a mandatory feature given the number of providers that support this functionality.

dngray:
> > [...] removing Disroot allows us to increase our minimum security requirement for email storage from "encrypted" to "encrypted with zero-access encryption", a change which has also been made in this PR.
> 
> I think this is the most important change here, this should probably be a mandatory feature given the number of providers that support this functionality.

That I am inclined to agree with. I did check with them and apparently they are looking at site-wide MFA of some kind, so that would be cool. Obviously they said that will take some time.

I think for now we might remove it, and then re-add it when they've done that.

docs/email.en.md
```diff
@@ -351,9 +308,8 @@ Email servers deal with a lot of very sensitive data. We expect that providers w
 - A server suite preference of TLS 1.2 or later and a plan for [Deprecating TLSv1.0 and TLSv1.1](https://datatracker.ietf.org/doc/draft-ietf-tls-oldversions-deprecate/).
 - [SMTPS](https://en.wikipedia.org/wiki/SMTPS) submission, assuming SMTP is used.
 - Website security standards such as:
-
-- [HTTP Strict Transport Security](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security)
-- [Subresource Integrity](https://en.wikipedia.org/wiki/Subresource_Integrity) if loading things from external domains.
+    - [HTTP Strict Transport Security](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security)
+    - [Subresource Integrity](https://en.wikipedia.org/wiki/Subresource_Integrity) if loading things from external domains.
```
dngray, line 312: I did spot these linter errors:

docs/email.en.md:311 [markdownlint]  Warning MD007/ul-indent: Unordered list indentation [Expected: 2; Actual: 4]
docs/email.en.md:312 [markdownlint]  Warning MD007/ul-indent: Unordered list indentation [Expected: 2; Actual: 4]

docs/email.en.md
```diff
@@ -363,9 +319,8 @@ Email servers deal with a lot of very sensitive data. We expect that providers w
 - Implementation of [Authenticated Received Chain (ARC)](https://en.wikipedia.org/wiki/Authenticated_Received_Chain), this is useful for people who post to mailing lists [RFC8617](https://tools.ietf.org/html/rfc8617).
 - Bug-bounty programs and/or a coordinated vulnerability-disclosure process.
 - Website security standards such as:
-
-- [Content Security Policy (CSP)](https://en.wikipedia.org/wiki/Content_Security_Policy)
-- [Expect-CT](https://datatracker.ietf.org/doc/draft-ietf-httpbis-expect-ct)
+    - [Content Security Policy (CSP)](https://en.wikipedia.org/wiki/Content_Security_Policy)
+    - [Expect-CT](https://datatracker.ietf.org/doc/draft-ietf-httpbis-expect-ct)
```
dngray, line 323: I spotted a couple more:

docs/email.en.md:322 [markdownlint]  Warning MD007/ul-indent: Unordered list indentation [Expected: 2; Actual: 4]
docs/email.en.md:323 [markdownlint]  Warning MD007/ul-indent: Unordered list indentation [Expected: 2; Actual: 4]

Weirdly doesn't seem to work if you have any less spaces.

jonaharagon: > Weirdly doesn't seem to work if you have any less spaces.

Oh yeah I noticed this as well, I'm not sure why this is the case, I need to look into this.

PR #1118, opened 2022-04-27
Session messenger update

lx-is:
I was researching Session so I took the time to update the recommendation with some more polished info.

I replaced the download links with non-direct download links. I think linking to a direct download is a bad practice; visitors should ensure their download comes from the actual site, which is harder to determine with a direct link. I also removed the Android download link since it was a duplicate of the F-Droid link.

### DIFF
--- a/includes/abbreviations.en.md
+++ b/includes/abbreviations.en.md
@@ -46,7 +46,6 @@
 *[OTPs]: One-Time Passwords
 *[OpenPGP]: Open-source implementation of Pretty Good Privacy (PGP)
 *[P2P]: Peer-to-Peer
-*[PFS]: Perfect Forward Secrecy
 *[PGP]: Pretty Good Privacy (see OpenPGP)
 *[PII]: Personally Identifiable Information
 *[QNAME]: Qualified Name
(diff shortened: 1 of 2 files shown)


includes/abbreviations.en.md
```diff
@@ -46,6 +46,7 @@
 *[OTPs]: One-Time Passwords
 *[OpenPGP]: Open-source implementation of Pretty Good Privacy (PGP)
 *[P2P]: Peer-to-Peer
+*[PFS]: Perfect Forward Secrecy
```
jonaharagon, line 49: We don't actually use this acronym anywhere on the site.

includes/abbreviations.en.md
```diff
@@ -73,3 +74,4 @@
 *[cgroups]: Control Groups
 *[fork]: In software development, a fork is created when developers take a copy of source code from one software package and start independent development on it, creating a distinct and separate piece of software.
 *[rolling release]: An update release cycle in which updates are released very frequently, instead of at set intervals.
+*[perfect forward secrecy]: When a piece of an encryption system automatically and frequently changes the keys it uses to encrypt and decrypt information, such that if the latest key is compromised, it exposes only a small portion of the user’s sensitive data.
```
jonaharagon, line 77: Also (leaving a comment here for future reference) I want to avoid putting privacy/security-related terms in the abbreviations section. A term like "fork" isn't particularly relevant to our readers, it's just background information. A term like "perfect forward secrecy" actually has direct impact on what we are discussing, so I'd rather this be on the page. Something like:

> Session does [not](https://getsession.org/blog/session-protocol-technical-information) support _perfect forward secrecy_, which is when an encryption system automatically and frequently changes the keys it uses to encrypt and decrypt information, such that if the latest key is compromised it exposes only a small portion of your sensitive data.

jonaharagon (approved):

includes/abbreviations.en.md
dngray: > perfect forward secrecy

iirc its just "forward secrecy" in most documents I've seen, there isn't really a 'perfect' there.

PR #1121, opened 2022-04-27
Remove WKD

jonaharagon:
All keys are outdated, and I no longer use PGP. @dngray or @tommytran732 if you still want to keep your keys on WKD, update them so they actually have your @privacyguides.org email and add yours back to this PR. Otherwise they can all just be removed.


dngray (changes requested):
Will update with my key.

dngray (approved):

dngray:
I've updated mine, not sure if @tommytran732 is gonna do his. Super easy:

```
gpg --with-wkd-hash --fingerprint  0x{{ your key }}
gpg --no-armor --export 0x{{ your key }} > {{ wkd hash from above }}
```

PR #1123, opened 2022-04-27
Move email cloaking services and self-hosted email under provider recommendations

lx-is:
Moves email cloaking services and self-hosted email solutions higher up on the email providers page (below the email services recommendations).

I think they are better off higher on the page as they're important, especially email cloaking services.

After this gets merged/closed I plan on adding more info to the email cloaking services section; it is quite sparse of information.


jonaharagon (approved):
LGTM. Each provider could definitely use additional information, and probably full/regular recommendation cards (in a future PR).

PR #1150, opened 2022-05-01
Fix OnlyOffice description

elitejake:
It is web-based, right?

If it isn't, still fix the original sentence.

### DIFF
--- a/docs/productivity.en.md
+++ b/docs/productivity.en.md
@@ -35,7 +35,7 @@ Get working and collaborating without sharing your documents with a middleman or
 
     ![OnlyOffice logo](assets/img/productivity/onlyoffice.svg){ align=right }
 
-    **OnlyOffice** is alternative, it is free and open-source office suite with extensive functionality.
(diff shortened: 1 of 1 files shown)


jonaharagon:
Well, OnlyOffice also has iOS and Android apps. And it has a desktop app, which might itself be web-based, I'm not sure, but I don't know if I would call a standalone app "web-based" just because it isn't native code.

docs/productivity.en.md
```diff
@@ -35,7 +35,7 @@ Get working and collaborating without sharing your documents with a middleman or
 
     ![OnlyOffice logo](assets/img/productivity/onlyoffice.svg){ align=right }
 
-    **OnlyOffice** is alternative, it is free and open-source office suite with extensive functionality.
+    **OnlyOffice** is a web-based free and open-source office suite with extensive functionality.
```
jonaharagon, line 38:
```suggestion
    **OnlyOffice** is a cloud-based free and open-source office suite with extensive functionality, including integration with Nextcloud.
```

PR #1152, opened 2022-05-02
Show backers on donation page

jonaharagon:
Do we have an issue with adding this? https://privacyguides-privacyguides-org-preview-pr-1152.surge.sh/about/donate/#backers

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -327,7 +327,7 @@
         },
         "mkdocs-material": {
             "path": "./mkdocs-material",
-            "version": "==8.2.11+insiders.4.13.1"
+            "version": "==8.2.12+insiders.4.13.2"
         },
         "mkdocs-material-extensions": {
             "hashes": [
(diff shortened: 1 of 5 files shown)


Dyrimon:
That page is connecting to `cloudflareinsights.com` and `sentry.io` trackers. Is it possible to remove those?

freddy-m:
> Do we have an issue with adding this? 

Not really, though I don't mind the page as it is. Just hope it doesn't encourage the sort of spam we used to get with PTio...

dngray:
> That page is connecting to `cloudflareinsights.com` and `sentry.io` trackers. Is it possible to remove those?

Seems that OpenCollective uses https://www.cloudflare.com/web-analytics, so linking to this does link to that.

Looking through it though, I don't think it is particularly privacy invasive. It only provides some fairly minimal data to OpenCollective, such as location, etc.

If we just linked to the opencollective org https://opencollective.com/privacyguides then it would be the same thing.

jonaharagon:
> Just hope it doesn't encourage the sort of spam we used to get with PTio...

This is less of a sponsorship-type change and more just a statement of fact. Worth pointing out this only links to OpenCollective profiles and not custom links (although of course the profiles could have their own links specified). If we wanted to explore a partner program again I'd probably want to model it after https://www.qubes-os.org/partners/, but we're not making that change now or adding any sponsor-type tier back to Open Collective at the moment.

dngray (approved):

PR #1153, opened 2022-05-02
Remove ".onion" check from email providers

jonaharagon:
This moves ".onion" address information from its own banner to the additional information category. I think currently we are taking a too negative stance against providers who do not offer this functionality, which is really only tangential to providing an email service.

Discussion: https://github.com/privacyguides/privacyguides.org/discussions/1149


dngray:
> I think currently we are taking a too negative stance against providers who do not offer this functionality, which is really only tangential to providing an email service.

Agreed.

dngray (approved):

PR #1155, opened 2022-05-02
Link to Mastodon instead of Twitter

jonaharagon:
Since we have Mastodon now, and it's somewhat mainstream as of late, it makes more sense to start promoting this primarily as our social media platform. We'll still keep Twitter in the README and website footer.


freddy-m (approved):
Good idea!

dngray (approved):

PR #1176, opened 2022-05-04
Re-recommend Snowflake (#1176)

jonaharagon:
I think delisting Snowflake was a mistake. I believe the original reasoning was that, like ToS;DR, it doesn't directly impact the installing user's privacy. However, unlike ToS;DR, it *does* directly impact user privacy: The privacy of users of Snowflake proxies.

I think alongside a section on what Snowflake is, who it impacts, and a clear warning that it does nothing to help you yourself, we should recommend it because it is a net benefit to privacy overall if people in more privileged situations install it. I don't think that was super clear when we previously recommended it, but I think that is fixed in this PR.

https://privacyguides-privacyguides-org-preview-pr-1176.surge.sh/browsers/#snowflake


dngray:
I do like what is written there. I think this card system provides a better ability to actually explain why you might use it and what some of the pitfalls are.

blacklight447 (approved):
I Agree with this change. Also, helping Tor users is in everyones best interest.

PR #1177, opened 2022-05-04
Add context to browser configuration options

jonaharagon:


docs/browsers.en.md
```diff
@@ -57,43 +54,55 @@ These are our current web browser recommendations and settings. We recommend kee
...
 
-- Select **Enable HTTPS-Only Mode in all windows**
+- [x] Select **Enable HTTPS-Only Mode in all windows**
+
+This prevents you from unintentionally connecting to a website in plain-text HTTP. The HTTP protocol is extremely uncommon nowadays, so this should have little impact on your day to day browsing.
 
 #### Sync
 
-The [Firefox Sync](https://hacks.mozilla.org/2018/11/firefox-sync-privacy/) service uses E2EE.
+[Firefox Sync](https://hacks.mozilla.org/2018/11/firefox-sync-privacy/) allows your browsing data (history, bookmarks, etc.) to be accessible on all your devices. It is fully E2EE, so it should be safe to use.
```
lx-is, line 101: It's the first time the abbreviation is used on this page.

```suggestion
[Firefox Sync](https://hacks.mozilla.org/2018/11/firefox-sync-privacy/) allows your browsing data (history, bookmarks, etc.) to be accessible on all your devices. It is fully end-to-end encrypted (E2EE), so it should be safe to use.
```

docs/browsers.en.md
```diff
@@ -126,13 +135,15 @@ These options can be found in :material-menu: → :gear: **Settings** → **Priv
 
 ##### HTTPS-Only Mode
 
-- Select **Always use secure connections**
+- [x] Select **Always use secure connections**
+
+This prevents you from unintentionally connecting to a website in plain-text HTTP. The HTTP protocol is extremely uncommon nowadays, so this should have little impact on your day to day browsing.
```
lx-is, line 140:
```suggestion
This prevents you from unintentionally connecting to a website in plain-text HTTP. The HTTP protocol is extremely uncommon nowadays, so this should have little to no impact on your day to day browsing.
```

lx-is (changes requested):

lx-is:
I'm going to do the Safari section later tn

docs/browsers.en.md
```diff
@@ -57,43 +54,55 @@ These are our current web browser recommendations and settings. We recommend kee
...
+This protects you by blocking social media trackers, fingerprinting scripts (note that this does not protect you from *all* fingerprinting), cryptominers, cross-site tracking cookies, and some other tracking content. It protects from many common threats, but does not completely block all tracking avenues, because it is designed to never have any impact on site functionality which limits its filters.
 
 ##### Sanitize on Close
 
 If you want to stay logged in to particular sites, you can allow exceptions in **Cookies and Site Data** → **Manage Exceptions...**
 
-- Select **Delete cookies and site data when Firefox is closed**
+- [x] Check **Delete cookies and site data when Firefox is closed**
+
+This protects you from persistent cookies, but does not protect you against cookies acquired during any one browsing session. When this is enabled you know you can easily clear your browser of all cookies by simply closing or restarting Firefox.
```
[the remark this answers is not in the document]
jonaharagon: so:
```suggestion
This protects you from persistent cookies, but does not protect you against cookies acquired during any one browsing session. When this is enabled, it becomes possible to easily cleanse your browser cookies by simply restarting Firefox. You can set exceptions on a per-site basis, if you wish to stay logged in to a particular site you visit often.
```

docs/browsers.en.md
jonaharagon: Gonna think about this... I am leaning towards leaving it as-is and changing the rule for very common abbreviations like E2EE, because the reader is *probably* familiar with it after reading enough of the site, and the glossary functionality includes the full text on hover regardless.

lx-is: I actually thought about that too. I figured the browser page is one of the highest trafficked pages, and so the highest chance of someone not being confident with the term is higher. There is the contextual definition, but I've found it doesn't work well on mobile, which is the platform statistically most readers view the site from. For example, if the definition is near the screen edge, it won't wrap and will instead be cut off. So, I've been avoiding relying on the definition popups in my PRs.

docs/browsers.en.md
```diff
@@ -164,19 +175,19 @@ Privacy Report is accessible through the "**Aa**" icon in the URL bar.
 
 This is WebKit's own [implementation](https://webkit.org/blog/8943/privacy-preserving-ad-click-attribution-for-the-web/) of privacy preserving ad click attribution. If you do not wish to participate, you can disable this feature.
 
-- Select **Privacy Preserving Ad Measurement**
+- [x] Select **Privacy Preserving Ad Measurement**
```
lx-is:
```suggestion
- [ ] Uncheck **Privacy Preserving Ad Measurement**

Ad click measurement has traditionally used tracking technology that infringes on user privacy. [Private Click Measurement](https://webkit.org/blog/11529/introducing-private-click-measurement-pcm/) is a WebKit feature and proposed web standard aimed towards allowing advertisers to measure the effectiveness of web campaigns without compromising on user privacy.

The feature has little privacy concerns on its own, so while you can choose to leave it on, we consider the fact that it's automatically disabled in Private Browsing to be an indicator for disabling the feature.
```

docs/browsers.en.md
```diff
@@ -164,19 +175,19 @@ Privacy Report is accessible through the "**Aa**" icon in the URL bar.
 
 This is WebKit's own [implementation](https://webkit.org/blog/8943/privacy-preserving-ad-click-attribution-for-the-web/) of privacy preserving ad click attribution. If you do not wish to participate, you can disable this feature.
```
lx-is, line 176:
```suggestion
```

lx-is (changes requested):

docs/browsers.en.md
```diff
@@ -164,19 +175,19 @@ Privacy Report is accessible through the "**Aa**" icon in the URL bar.
...
 
-- Select **Privacy Preserving Ad Measurement**
+- [ ] Uncheck **Privacy Preserving Ad Measurement**
 
 ##### Apple Pay
 
 If you do not use Apple Pay, you can toggle off the ability for websites to check for it.
 
-- Select **Check for Apple Pay**
+- [ ] Uncheck **Allow websites to check for Apple Pay and Apple Card**
```
lx-is, line 184: I wrote an email to Apple asking whether or not this should be enabled or not. My fear is that, by disabling it, it's creating another "do not track" setting scenario where it just makes it easier to identify a browser.

> Hi, I'm reaching out to see if your team could answer a question I had about the "Check for Apple Pay" feature in Safari.
> 
> As the Safari specific privacy policy states, "When you are in an app or on a website that uses Apple Pay, ... they can check if you have Apple Pay enabled on that device." The takeaway being: sites are able to probe the browser for whether or not the "Check for Apple Pay" feature is enabled.
> 
> My concern is that, by *disabling* this feature, it can harm privacy by potentially making the browser more identifiable as most Safari users likely do not disable the feature.
> 
> This scenario would be akin to the infamous "ask website to not track" setting, that is considered a flop as it created yet another data point to identify browsers.
> 
> I get that fingerprinting is pretty inevitable unless you're using the Tor browser, but the goal isn't to make fingerprinting easier either.
> 
> So my question is — should I keep the "Check for Apple Pay" setting enabled in an effort to blend in with other Safari users, or is that a misguided effort?
> 
> Thanks for your time.

jonaharagon:
@lx-is I think I'm going to merge this as-is, because I believe the information about Apple Pay is accurate. If you receive something to the contrary from Apple we can open another PR.

PR #1192, opened 2022-05-10
Update Syncthing description

elitejake:
The original description isn't helpful, it's like an advertisement.


mfwmyfacewhen (approved):
Looks good to me

PR #1195, opened 2022-05-11
Minor grammar fix in the first paragraph in Multi-factor Authentication

subatomic-honda-civic:
Resolves: # 

Changed "receive from an SMS or app." to "receive from SMS or an app." as **S**MS starts with an vowel but **a***pp* is. Basically minor a grammar change in a nutshell.


mfwmyfacewhen (approved):
Looks good

Issue #1196, opened 2022-05-11
Strange Wording for Firefox Sync


mfwmyfacewhen:
## Description
"It is fully E2EE, so it should be safe to use." - this is very strange wording. Either it is safe to use or it isn't, this just makes it sound like we're unsure. Is there a reason it's worded like this or can it just be shortened to "It is fully E2EE"?
**URL of affected page:** https://www.privacyguides.org/browsers/#sync

PR #1198, opened 2022-05-11
Fix wording on Firefox sync

mfwmyfacewhen:
Resolves: #1196 

"it should be safe to use" makes it sound like we're unsure about recommending it. Removed this part so it is shorter and more confident.

### DIFF
--- a/docs/browsers.en.md
+++ b/docs/browsers.en.md
@@ -98,7 +98,7 @@ This prevents you from unintentionally connecting to a website in plain-text HTT
 
 #### Sync
 
-[Firefox Sync](https://hacks.mozilla.org/2018/11/firefox-sync-privacy/) allows your browsing data (history, bookmarks, etc.) to be accessible on all your devices. It is fully E2EE, so it should be safe to use.
+[Firefox Sync](https://hacks.mozilla.org/2018/11/firefox-sync-privacy/) allows your browsing data (history, bookmarks, etc.) to be accessible on all your devices. It is fully E2EE.
 
 #### Extensions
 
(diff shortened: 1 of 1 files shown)


docs/browsers.en.md
```diff
@@ -98,7 +98,7 @@ This prevents you from unintentionally connecting to a website in plain-text HTT
 
 #### Sync
 
-[Firefox Sync](https://hacks.mozilla.org/2018/11/firefox-sync-privacy/) allows your browsing data (history, bookmarks, etc.) to be accessible on all your devices. It is fully E2EE, so it should be safe to use.
+[Firefox Sync](https://hacks.mozilla.org/2018/11/firefox-sync-privacy/) allows your browsing data (history, bookmarks, etc.) to be accessible on all your devices. It is fully E2EE.
```
subatomic-honda-civic, line 101: "It is fully E2EE." to me, sounds a bit *weird*, maybe somehow "integrate" that into the previous sentence? 

Something like

> ... to be accessible on all your devices while using E2EE.

subatomic-honda-civic (changes requested):

docs/browsers.en.md
mfwmyfacewhen: Yeah I agree, maybe "Firefox Sync allows your browsing data (history, bookmarks, etc.) to be accessible on all your devices and protects it with E2EE."

subatomic-honda-civic (approved):
LGTM, Honda Civic Approved :+1:

freddy-m (approved):
LGTM

PR #1205, opened 2022-05-12
Fix wording in cloud storage page

mfwmyfacewhen:
E2EE eliminates trust in your provider, so I updated the description to reflect that.


blacklight447 (approved):
Looks good to me.

PR #1207, opened 2022-05-12
Make headings consistent and remove redundant descriptions

mfwmyfacewhen:
The title and subtitle are redundant here. Also the blurb about browsing the "dark net" is unnecessary as the title is self-explanatory.

### DIFF
--- a/docs/cloud.en.md
+++ b/docs/cloud.en.md
@@ -6,7 +6,7 @@ If you are currently using a Cloud Storage Service like Dropbox, Google Drive, M
 
 Trust your provider by using an alternative below that supports E2EE.
 
-### Nextcloud
+## Nextcloud
 
 !!! recommendation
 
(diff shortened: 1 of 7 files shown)


lx-is (changes requested):
~~Change each network's heading to h2 instead of h3~~

wait nvm, that wouldn't be consistent with other pages

lx-is (approved):

blacklight447:
Are we sure we don't just wanna replace it rather then just remove? It seems rather emoty without it, and it wouldn't be consistent with the rest of the site which provides small describtions at the top of the page.

mfwmyfacewhen:
> Are we sure we don't just wanna replace it rather then just remove? It seems rather emoty without it, and it wouldn't be consistent with the rest of the site which provides small describtions at the top of the page.

Sure, what do you think should go there? Maybe something like "These networks aren't accessible through a normal browser, you need specific software to use them."

jonaharagon (changes requested):
> wait nvm, that wouldn't be consistent with other pages

probably an indication the other pages should be changed.

jonaharagon (approved):

PR #1210, opened 2022-05-12
Detail Safari Private Browsing

lx-is:
Detail some of the benefits of Safari Private Browsing, added macOS instructions, and refined the iOS instructions based on [Apple's instructions](https://support.apple.com/en-us/HT203036).


mfwmyfacewhen (approved):
LGTM

PR #1215, opened 2022-05-13
Add F-Droid repository link for Bromite

subatomic-honda-civic:
Added the F-Droid repository link for Bromite, which is located under the Download drop-down

Resolves: #1170  

<sup>Originally posted by **d4rklynk** May  3, 2022</sup>
We should add the repo link (or at least this link https://www.bromite.org/fdroid) in the recommended configuration of Bromite.
- Add repo link for F-droid
- Check box in Droidi-fy</div>


freddy-m (commented):
Surely we don't need both and Android link and an FDroid link in this scenario? cc @privacyguides/team

mfwmyfacewhen:
> Surely we don't need both and Android link and an FDroid link in this scenario? cc @privacyguides/team

No I think just pick one or the other.

freddy-m:
If that's the case then we'll just go for FDroid.

freddy-m (approved):
This is better IMO

dngray:
The Android one normally refers to the download not being in F-Droid. This isn't an official F-Droid repository.

That also said we don't recommend F-Droid, we recommend [Neo Store formerly Droid-Ify](https://www.privacyguides.org/android/#neo-store) which makes it a whole lot easier, you just toggle the repository on, as it's already there.

Screenshot_20220513-232256

dngray (approved):

PR #1216, opened 2022-05-13
Made the "Extended Description" for ToS;DR go more smoothy.

subatomic-honda-civic:
Resolves: # 

Made the extended sentence for ToS;DR more smoothy and less choppy by using transition words.


freddy-m (commented):
Not so sure this is an improvement - seems a little clunky to me. Thoughts @privacyguides/team?

mfwmyfacewhen:
> Not so sure this is an improvement - seems a little clunky to me. Thoughts @privacyguides/team?

I think it works better either like this or with a semicolon. "We do not recommend installing ToS;DR as a browser extension; the same information is provided on their website."

freddy-m:
>  with a semicolon

This is the best option.

dngray (approved):

Issue #1221, opened 2022-05-13
ControlD does not support DNSSEC


prestr:
## Description
While using ControlD did a test on https://cmdns.dev.dns-oarc.net/ and https://cmdns.dev.dns-oarc.net/ and it said DNSSEC was failing. 
I reached out to ControlD via email and they advised "Control D DNS does not support DNSSEC.". I also got the same response on reddit. 

**URL of affected page:** https://www.privacyguides.org/dns/

mfwmyfacewhen:
That's one of the requirements for the recommendations so I think it should be removed. Did the test myself and got the same thing.

Dyrimon:
In a [reddit comment](https://www.reddit.com/r/ControlD/comments/okb7g6/comment/h56tmqz/?utm_source=share&utm_medium=web2x&context=3) @yegors says DNSSEC is enforced.

prestr:
> In a [reddit comment](https://www.reddit.com/r/ControlD/comments/okb7g6/comment/h56tmqz/?utm_source=share&utm_medium=web2x&context=3) @yegors says DNSSEC is enforced.

I guess I need to stop deleting my reddit posts. He said a few weeks ago its not used.

PR #1222, opened 2022-05-13
Remove ControlD DNS

mfwmyfacewhen:
ControlD does not support DNSSEC. This is one of the requirements for providers, so it should be removed.

Resolves: #1221


lx-is:
If this gets approved, do you think it should be removed from the [signed profiles](https://www.privacyguides.org/dns/#signed-profiles) selection as well?

mfwmyfacewhen:
> If this gets approved, do you think it should be removed from the [signed profiles](https://www.privacyguides.org/dns/#signed-profiles) selection as well?

Yeah I didn't think of that.

PR #1224, opened 2022-05-14
Add account deletion guide

mfwmyfacewhen:
Adding a guide on deleting old accounts.

Resolves: #1203


docs/security/account-deletion.en.md
```diff
@@ -0,0 +1,7 @@
+---
+title: "Account Deletion"
+icon: 'material/form-textbox-password'
+---
+It's an unfortunate fact that over time we will accumulate online accounts, many of which we may no longer use. Most people simply forget about them and leave them to rot; however, deleting your old accounts can keep threat actors from accessing them and any data that you may have left there.
```
lx-is, line 5: Here's my take on the beginning except. You can decide if you like it better or not

```suggestion
It can prove easy to accumulate a number of online accounts, but it can prove difficult to eventually delete any forgotten accounts. While deleting past accounts is an important step in reclaiming your privacy, it is often made difficult by [deceptive design](https://www.deceptive.design/) patterns.
```

lx-is (changes requested):

docs/security/account-deletion.en.md
mfwmyfacewhen: I like "It can prove easy to accumulate a number of online accounts" bc I think keeping a positive tone is a good thing. I don't really want to say it's going to be difficult right at the start of the guide because I don't want people to get demoralized before they even start.

lx-is: Excellent point. I edited it to make it sound more welcoming, if you'd like to take another look.

mfwmyfacewhen: How about "It can prove easy to accumulate a number of online accounts,  many of which we may no longer use. Deleting unused accounts is an important step in reclaiming your privacy, as these often contain personal information. While [deceptive design](https://www.deceptive.design/) patterns can often introduce inconveniences along the way, this guide aims to navigate you through the process."

lx-is: Each iteration keeps getting better lol.

How about:

> It can prove easy to accumulate a number of online accounts, many of which may no longer be useful. Deleting those past accounts is an important step in reclaiming your privacy, as dormant accounts are especially vulnerable. While [deceptive design](https://www.deceptive.design/) patterns can often introduce inconveniences along the way, this guide aims to navigate you through the account deletion process.

mfwmyfacewhen: Nice

docs/security/account-deletion.en.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Account Deletion"
+icon: 'material/account-remove'
+---
+It can be easy to accumulate a number of online accounts over time, many of which you may no longer use. Deleting these unused accounts is an important step in reclaiming your privacy, as dormant accounts are especially vulnerable to having their credentials and/or data leaked in a [data breach](https://www.malwarebytes.com/data-breach). While [deceptive design](https://www.deceptive.design/) patterns can often introduce inconveniences along the way, this guide aims to navigate you through the account deletion process.
```
dngray, line 5: I'd say we can elaborate here without using third party links to describe these things.

I'd elaborate and say data breaches do happen regularly and use https://haveibeenpwned.com/PwnedWebsites a reference to *how often* it occurs.

docs/security/account-deletion.en.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Find Old Accounts
+
+### Password Manager
+
+If you have a password manager that you've used for your entire digital life, this part will be very easy. You might not think you've had one, but many people used the password manager in their browser before they even knew what a password manager was.
```
dngray, line 11: Might mention tools like keepassxc 2.6 allow for [comprehensive reports](https://keepassxc.org/blog/2020-07-07-2.6.0-released/) to be checked with Have I Been Pwned [as does Bitwarden](https://bitwarden.com/help/reports/)

docs/security/account-deletion.en.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Replace Real Info with Fake Info
+
+Once you've made sure you can log in, change all the information in your account to fake information. You can use a [fake name generator](https://www.fakenamegenerator.com/) to accomplish this. The reason you want to do this is many sites will retain information you previously had even after account deletion, which opens it up to being exposed in a data breach. The hope is that they will overwrite the previous information with the new data, but of course there is no guarantee of this.
+
+For the account email, either create a new burner email account via your provider of choice or create an alias using an [email aliasing service](/email/#email-aliasing-services). Do not use temporary email providers like GuerillaMail, as many sites will give a period of time in which your account can be reactivated. Delete your burner email account after this period expires.
```
dngray, line 29: I think we can avoid mentioning specific products here.

dngray (changes requested):

docs/security/account-deletion.en.md
mfwmyfacewhen: Even the google password manager and the apple one have something like this so I might just mention it generally without talking about specific products, watchu think?

dngray: Yeah that might be a better way to go about it, maybe show the Bitwarden one off because it has good documentation.

blacklight447 (approved):

dngray (approved):

PR #1230, opened 2022-05-15
Fix youtube embeds

mfwmyfacewhen:
The YouTube embeds are awful right now I want to try to fix the height.


lx-is:
You should just create a YouTube embedded video class in the stylesheet and then have it be used by these iframes. Then you wouldn't have to define the styling for every embed

mfwmyfacewhen:
> You should just create a YouTube embedded video class in the stylesheet and then have it be used by these iframes. Then you wouldn't have to define the styling for every embed

I do not know what I'm doing lol

dngray:
I think the plan was at some point  a self-hosted invidious instance for these videos, that way it wouldn't require js to play them

PR #1231, opened 2022-05-15
Maintain aspect ratio of YouTube embeds

lx-is:
Resolves the issue where the embeds become visually glitchy when resizing the window. It now maintains a 16:9 aspect ratio

follow up to #1230


mfwmyfacewhen (approved):
Nice looks much better

PR #1233, opened 2022-05-15
Add Brave

dngray:
@tommytran732 and I have been working on this one for a while, which I think is now ready.


tommytran732 (approved):
The new format looks good. Nothing to complain about.

PR #1238, opened 2022-05-15
Edit ubo to not recommend installing all filter lists

mfwmyfacewhen:
Recommend to only use UBO in its default configuration.

Resolves: #1237

### DIFF
--- a/docs/browsers.en.md
+++ b/docs/browsers.en.md
@@ -138,7 +138,9 @@ We recommend setting *Aggressive* which can be found in :material-menu: **Settin
 
 We also suggest changing the fingerprinting blocker to *Strict* in :material-menu: **Settings** → **Shields** → **Fingerprint blocking**. You can always downgrade it if you need to on a per-site basis.
 
-!!! danger Brave allows users to select additional adblock filters by visiting brave://adblock. We recommend that you do not use this feature and stick to the default settings provided by the Brave Shield to not stand out from other braves users and to not increase your attack surface. If there is a vulnerability in the Shield, third party filters can provide malicious rules to exploit it.
+!!! danger "Do not use brave://adblock!" 
+    
+    Brave allows users to select additional adblock filters by visiting brave://adblock. We recommend that you do not use this feature and stick to the default settings provided by the Brave Shield to not stand out from other braves users and to not increase your attack surface. If there is a vulnerability in the Shield, third party filters can provide malicious rules to exploit it.
 
 ##### Social media blocking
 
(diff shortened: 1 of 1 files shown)


docs/browsers.en.md
```diff
@@ -262,7 +266,7 @@ Do note that Private Browsing does not save cookies and website data, so it won'
 
 ##### iCloud Sync
 
-While synchronization of Safari History, Tab Groups, and iCloud Tabs uses E2EE, bookmarks sync does [not](https://support.apple.com/en-us/HT202303); they are only encrypted in transit and stored in an encrypted format on Apple's servers. Apple may be able to decrypt and access them.
+While synchronization of Safari History, Tab Groups, and iCloud Tabs uses E2EE, bookmarks sync does [not](https://support.apple.com/en-us/HT202303); they are only encrypted in transit and stored in an encrypted format on Apple's servers. Apple can be able to decrypt and access them.
```
lx-is, line 269:
```suggestion
Synchronization of Safari History, Tab Groups, iCloud Tabs, and saved passwords are E2EE. However, bookmarks are [not](https://support.apple.com/en-us/HT202303). Apple can decrypt and access them in accordance with their [privacy policy](https://www.apple.com/legal/privacy/en-ww/).
```

docs/browsers.en.md
```diff
@@ -292,11 +296,9 @@ We generally do not recommend installing [any extensions](https://www.sentinelon
...
-
-Additional filter lists do slow things down and may increase your attack surface, so only apply what you need.
-
-uBlock Origin also has different [blocking modes](https://github.com/gorhill/uBlock/wiki/Blocking-mode). The easy mode [might not](https://www.ranum.com/security/computer_security/editorials/dumb/) necessarily keep you safe from every tracker out there, whereas the more advanced modes let you control exactly what needs to run.
+!!! danger "Stick to the default filter lists"
+    
+    Additional filter lists do slow things down and may increase your attack surface, so only apply what you need. If there is a vulnerability in uBlockOrigin, third party filters can provide malicious rules to exploit it.
```
lx-is, line 301:
```suggestion
    Additional filter lists do slow things down and may increase your attack surface, so only apply what you need. If there is a vulnerability in uBlock Origin, third party filters can provide malicious rules to exploit it.
```

lx-is (changes requested):

docs/browsers.en.md
```diff
@@ -280,7 +284,7 @@ We generally do not recommend installing [any extensions](https://www.sentinelon
 
     **uBlock Origin** is a popular content blocker that could help you block ads, trackers, and fingerprinting scripts.
 
-    We suggest enabling all of the [filter lists](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists) under the "Ads," "Privacy," and "Malware domains". The "Annoyances" and "Multipurpose" lists can also be enabled, but they may break some social media functions. The *AdGuard URL Tracking Protection* filter list makes extensions like CleanURLs and NeatURLs redundant.
+    We suggest leaving the extension in its default configuration, as extra filter lists can add additional [attack surface](https://portswigger.net/research/ublock-i-exfiltrate-exploiting-ad-blockers-with-css). Only enable the AdGuard URL Tracking Protection filter list, as it makes extensions like CleanURLs and NeatURLs redundant.
```
lx-is, line 287: Would it be good to mention the uBO's [blocking modes](https://github.com/gorhill/uBlock/wiki/Blocking-mode)?

mfwmyfacewhen: There's no mention of this in the actual ui so I don't see a point, it will probably just confuse people.

Lostname777:
What do you think about enabling "uBlock filters – Annoyances" under annoyances column as it is an ublock filter but not enabled by default.

blacklight447 (approved):

PR #1240, opened 2022-05-15
Update Safari Privacy Report instruction

lx-is:
Yes, this change is absolutely necessary. It will revolutionize the site.

@mfwmyfacewhen this is what you get for overwriting Aa in #1234 😆 (also lucky you got to open pr number 1..2..3..4)

## Before
before

## After
after


mfwmyfacewhen (approved):
LGTM

PR #1242, opened 2022-05-15
Change "Attention" to "Warning" in Warning Box Message

d4rklynk:
You wrote "warning" on other Warning Boxes AFAIK, I think it would be better to stick to it.
It's confusing to have too many terms to explain a danger IMO


mfwmyfacewhen (approved):
Yeah I think that's better

PR #1245, opened 2022-05-16
Rename "email cloaking" to "email aliasing" on overview page

lx-is:
It was renamed with #1133

I also reordered the two services in the way they appear on the main page


mfwmyfacewhen (approved):
LGTM

PR #1246, opened 2022-05-16
Rename "MullvadDNS" to just "Mullvad"

lx-is:
"MullvadDNS" isn't the official name. Plus, table column is just for the provider name, in which just "Mullvad" would make sense in the same way that Cloudflare is just "Cloudflare"


mfwmyfacewhen (approved):
Yep

PR #1261, opened 2022-05-17
Change Warning to Danger for Tor Browser Warning Box

d4rklynk:
Resolves: #


d4rklynk:
I think we should have a red box for this one, because it's really not advised at all to install any add-ons to TB.
And if you install addon to TB, you defeat the **ALL** purpose of Tor Browser

mfwmyfacewhen (approved):
Yeah agree

dngray (approved):

PR #1269, opened 2022-05-18
Add warning to avoid third-party extensions on Brave

d4rklynk:
I think adding UBO on Brave is close to useless because it would be redundant with **Shield** feature of the browser.

I'm not sure about my syntax of my sentences though as english isn't my native language.
Please reconsider my grammar, vocabulary and the syntax if you'll agree to merge this.

### DIFF
--- a/docs/browsers.en.md
+++ b/docs/browsers.en.md
@@ -131,10 +131,6 @@ The [Arkenfox project](https://github.com/arkenfox/user.js) provides a set of ca
 
     1. We advise against using the Flatpak version of Brave as it is believed to feature a weaker sandboxing system. As well, the package is **not** maintained by Brave Software, Inc.
 
-## Don't install uBlock Origin
-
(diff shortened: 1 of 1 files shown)


lx-is (commented):
uBlock is only explicitly mentioned under the extensions heading for Firefox, so I think it's already pretty clear. That said, it wouldn't hurt to have a small notice; I do think the use of an admonition here is a bit too much.

Maybe an annotation under the shields heading would be a better fit.

d4rklynk:
I think we should let the small notice under the Brave Box, because if someone is used to install UBO on Chrome or FF, and check just PG for alternative, he wuld see directly that he shouldn't use UBO.
IMHO

d4rklynk:
But I agree that the admonition is probably too much

d4rklynk:
I changed it to a title instead of a box

lx-is:
I put the notice under the extensions category. This way, it applies to all third-party extensions and not uBO specifically.

lx-is (approved):

PR #1286, opened 2022-05-21
Default to dark mode

jonaharagon:
This is for Tommy, I only use light mode lol


lx-is:
sorry tommy but

https://kevq.uk/is-dark-mode-such-a-good-idea/

PR #1291, opened 2022-05-22
Change browser configuration guides to be closer to the ui of the browser

mfwmyfacewhen:

### DIFF
--- a/docs/browsers.en.md
+++ b/docs/browsers.en.md
@@ -269,11 +269,11 @@ This prevents you from unintentionally connecting to a website in plain-text HTT
 
 #### Recommended Configuration
 
-These options can be found in :gear: **Settings** → **Safari** → **Privacy and Security**.
+These options can be found in :gear: **Settings** → :material-apple-safari: **Safari** → **Privacy and Security**
 
 ##### Cross-Site Tracking Prevention
 
-- [x] Enable **Prevent Cross-Site Tracking**
+Prevent Cross-Site Tracking :material-toggle-switch:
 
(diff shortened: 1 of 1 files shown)


docs/browsers.en.md
```diff
@@ -278,7 +278,7 @@ The feature has little privacy concerns on its own, so while you can choose to l
 
 If you do not use Apple Pay, you can toggle off the ability for websites to check for it.
 
-- [ ] Disable **Allow websites to check for Apple Pay and Apple Card**
+Check for Apple Pay :material-toggle-switch-off-outline:
```
dngray, line 281: Probably should keep the Disable/Enable or Check/Uncheck in there as they show in https://docs.microsoft.com/en-us/style-guide/procedures-instructions/describing-interactions-with-ui

The reason is otherwise screen readers won't be able to likely tell the user what is going on.

For some of these it should be. *Switch, turn on, turn off*

docs/browsers.en.md
```diff
@@ -278,13 +278,18 @@ The feature has little privacy concerns on its own, so while you can choose to l
...
 
 Open Safari and tap the Tabs button, located in the bottom right. Then, expand the Tab Groups list.
 
-- [x] Select **Private**
+| Edit | Tab Groups |  Done  |
+| :--- | :--------- | -----: |
+| :material-cellphone:   Start Page |       |         |
+| :material-hand-back-left:   Private |          | :material-check: |
+|      |        |       |
+| :material-plus:   New Empty Tab Group |       |       |
```
jonaharagon, line 292: I'm torn about this, it looks fine but I also wonder if it should just be a screenshot. We can add alt text to that which is probably more accessibility-friendly.

mfwmyfacewhen: maybe

PR #1292, opened 2022-05-23
Clarify Neo Store rebranding

lx-is:


jonaharagon (approved):
Is there no way to do `<small>` tags with Markdown, maybe with a material-mkdocs/pymdownx extension?

lx-is:
> Is there no way to do `<small>` tags with Markdown, maybe with a material-mkdocs/pymdownx extension?

Not that I'm aware of

dngray (approved):

PR #1303, opened 2022-05-23
Add Feeder

dngray:
Closes: https://github.com/privacyguides/privacyguides.org/discussions/982

I think we might replace Handy News Reader with this, added benefit is it supports JSON feeds, and seems to have more community contributors, CI tests etc.

It now supports folders (it didn't used to), and the ability to pin articles for later, so not really a whole lot of reason to list both now.


dngray:
Replaced with high rez graphic from https://gitlab.com/spacecowboy/Feeder/-/tree/master/graphics

Author does not have SVG images.

dngray:
> This [page](https://privacyguides-privacyguides-org-preview-pr-1303.surge.sh/news-aggregators/) is already quite clunky... could we cut down on some of the other providers as well?

Thinking of removing Handy News Reader.

The issue is the others aren't multi platform, and tend to be native.

Issue #1258, opened 2022-05-17
microG wrongly represented


mar-v-in:
## Description

**URL of affected page:** https://www.privacyguides.org/android/grapheneos-vs-calyxos/

On the page comparing GrapheneOS and CalyxOS, there are a few statements made about microG, which may be true for microG on CalyxOS but do not apply to microG in general:

----

> it needs to be updated every time Android has a major version update (or the Android API changes)

The Android API is mostly backwards compatible - otherwise, regular apps would also break with major system updates. Previous major updates did not require specific updates in microG. For example, microG was not adjusted for Android 12 yet and thus does not make use of certain new features, yet it works perfectly fine on Android 12 systems.

> It also needs to run in the highly privileged `system_app` SELinux domain like regular Google Play Services

This is not true. While microG has a handful of convenience features that only work when run privileged (e.g. accounting the battery use from geolocation to the app actually requesting it), those features are entirely optional and microG works perfectly fine in the default `untrusted_app` domain. For example, [ProtonAOSP suggests installing microG as a regular app](https://protonaosp.org/gapps#install-microg) resulting in exactly such a setup. I myself am also running microG as regular unprivileged app.

> This is less secure than Sandboxed Google Play's approach, which does not need access to sensitive system APIs.

While technically true, I think this statement is misleading. A lot of code on Android runs in the privileged mode and one should only allow trusted code to run in this mode, but as microG is entirely open-source, reviewing it is possible, whereas Google Play Services code is proprietary and thus must be trusted blindly if used.

> With microG, you have the option to either not use a network location backend at all, shift trust to another location backend like Mozilla, or use [DejaVu](https://github.com/n76/DejaVu), a location backend that locally collects and saves RF-based location data to an offline database which can be used when GPS is not available.

microG uses a modular system that provides multiple options for location resolution. A module that reroutes requests to GPS is entirely possible. Also there is the possibility to do the location only based on cell towers and not wifi and prefetch a database of cell towers in a whole country (much smaller than a database of wifis would be), allowing for local location resolution without leaking any data (and [this is already developed](https://f-droid.org/packages/org.fitchfamily.android.gsmlocation/)).

> Choosing a network location like Mozilla to use with microG provides little to no privacy benefit over Google because you are still submitting the same data and trusting them to not profile you.

[...]

PR #1306, opened 2022-05-23
Update microG on Calyx information

jonaharagon:
Resolves: #1258

### DIFF
--- a/docs/android.en.md
+++ b/docs/android.en.md
@@ -51,9 +51,9 @@ Google Pixel phones are the only devices that currently meet GrapheneOS's [hardw
     [:octicons-code-16:](https://github.com/CalyxOS){ .card-link title="Source Code" }
     [:octicons-heart-16:](https://members.calyxinstitute.org/donate){ .card-link title=Contribute }
 
-For people who need Google Play Services, CalyxOS optionally includes [microG](https://microg.org/). CalyxOS also includes alternate location services, [Mozilla](https://location.services.mozilla.com/) and [DejaVu](https://github.com/n76/DejaVu).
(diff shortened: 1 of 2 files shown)


tommytran732:
Probably should add a note that MicroG is **partially** open source as components like Droidguard are not. We should also add a mention that, on CalyxOS, they add the privileged esim activation app (outside of MicroG) that cannot be disabled. Effectively, Google still has your hardware identifiers on Calyx anyways.

tommytran732 (dismissed):
Let's put this on hold as I want to discuss a few things in https://github.com/privacyguides/privacyguides.org/issues/1258

tommytran732:
Okay, I made some changes:
1. Note that microG is only partially open source as components like DroidGuard is still proprietary. Push notifications still go through Google's server as usual.
2. Note that microG can be used without signature spoofing and special privileges, however that will greatly reduce its compatibility and usefulness
3. Revert the claim about microG making it harder to track user with a network location service. That is not even true, because you can absolutely be trivially profiled even without cookies. Imagine if you have your smartwatch or wireless headphones around the phone when you use the network location service. Their MAC address would be submitted to the server. The next time the server sees the same MAC address from your smartwatch or headphones being submitted, it knows that its you again. This is not to mention, microG cannot actually enforce the no cookie policy as Micay explained, that is entirely up to the location backend.
4. Added mentions about the privileged eSIM app on CalyxOS. This nullifies one of the selling point of MicroG in the first place, because if you don't trust the privileged Play Services to not collect your hardware ID, why would you add a cannot-be-opted-out privileged Google app?
5. Make it absolutely clear that the recommendation for Calyx is harm reduction and should only be used for the OnePlus and Fairphone (and especiall the Fairphone). Let's be honest here, under which scenario/threat model would it make sense to use Calyx over GrapheneOS at all?

blacklight447 (approved):

dngray (approved):

dngray (approved):

Issue #947, opened 2022-04-08
Building a threat model


dngray:
I haven't decided whether this should be a part of the threat modeling page, but I'm thinking maybe not.

I'm thinking we should include https://github.com/privacyguides/privacyguides.org/discussions/468 under a "building a threat" page.

I wouldn't mind covering jurisdiction at the same time such as what is mentioned in https://github.com/privacyguides/privacyguides.org/pull/937

The TLDR of that is, consider where *you* are, consider the privacy policy of the service you intend to use and the laws of the country where that service resides.

Rather than a "good list of countries" or "bad list" which changes regularly.

Sirwhofromwhere:
In my opinion 'building a threat' should be part of the threat modelling page. As the Knowledge Base expands with further sections, it will be important to consolidate as much as possible in to directly relevant categories. That is unless you are able to implement a sub-category feature where clicking on the main category in the list presents a drop down, and then hyperlinking & referencing at the bottom of the page to it for further learning on that subject.

Approaching topics with overlap with a first priority of finding ways to incorporate them under the same category, whether that requires re-writing of the original page to capture all the information efficiently or simply expanding the page to a greater number of words, is the best long term strategy as it will always be easier to simply create another section. The cost of which is an unapproachable category list that discourages a snack approach to reading.

dngray:
So in this issue I'm thinking we should address:

  - https://github.com/privacyguides/privacyguides.org/discussions/468

This leads into:

- https://github.com/privacytools/privacytools.io/issues/1437#issuecomment-546845665
  - https://github.com/privacyguides/privacyguides.org/pull/937
    - Email
    - VPN (Also looking at adding Windscribe in this PR as discussion 
      https://github.com/privacyguides/privacyguides.org/discussions/372#discussioncomment-1691195
- https://github.com/privacyguides/privacyguides.org/issues/1020 (this one we can start minimally, and people can expand via PR), it needs to mostly be a stub for translators.

Few reasons for this:

- Eyes is not the only agreement, geographical location of the user is relevant to determine immediate threats in threat model
- Privacy is obtained through using strong E2EE, and not spraying data about yourself all over the internet, which is the reason for: https://github.com/privacyguides/privacyguides.org/discussions/949

Might possibly look at something that addresses https://github.com/privacyguides/privacyguides.org/discussions/888, I'm not quite sure what that would look like.

PR #1310, opened 2022-05-24
TM Example

dngray:
A few things I noticed, that I didn't have a chance to review, and a threat model example.

Closes: https://github.com/privacyguides/privacyguides.org/issues/947


dngray:
One thing I did notice is that acronyms like `TPM` don't seem to be showing the abbreviation.

lx-is:
> One thing I did notice is that acronyms like `TPM` don't seem to be showing the abbreviation.

I think I recall Jonah saying he wants to avoid putting privacy or security terms in the abbreviations file, opting to describe them directly on the page instead. I can't remember where he said this though since it's not in the style guide.

mfwmyfacewhen (approved):

docs/basics/common-threats.en.md
```diff
@@ -28,7 +28,7 @@ Whistleblowers and journalists, for example, can have a much more extreme threat
...
+Security and privacy are often conflated, because you need security to obtain any semblance of privacy: Using tools which appear private is futile if they could easily be exploited by attackers to release your data later. However, the inverse is not necessarily true; the most secure service in the world *isn't necessarily* private. The best example of this is trusting data to Google, who, given their scale, have had minimal security incidents by employing industry-leading security experts to secure their infrastructure. Even though Google provides a very secure service, very few would consider their data private in Google's free consumer products (Gmail, YouTube etc).
```
jonaharagon, line 31: I wouldn't make this change, I'm not sure what it adds to the article outside of unnecessary over-explanation.
```suggestion
Security and privacy are often conflated, because you need security to obtain any semblance of privacy: Using tools which appear private is futile if they could easily be exploited by attackers to release your data later. However, the inverse is not necessarily true; the most secure service in the world *isn't necessarily* private. The best example of this is trusting data to Google, who, given their scale, have had minimal security incidents by employing industry-leading security experts to secure their infrastructure. Even though Google provides a very secure service, very few would consider their data private in their hands.
```

blacklight447 (approved):

docs/basics/common-threats.en.md
```diff
@@ -48,7 +48,7 @@ Targeted attacks against a specific user are more problematic to deal with. Comm
...
+If you are concerned about **physical attacks** you should use an operating system with a secure verified boot implementation, such as Android, iOS, macOS, [Windows (with TPM)](https://docs.microsoft.com/en-us/windows/security/information-protection/secure-the-windows-10-boot-process). You should also make sure that your drive is encrypted, and that the operating system uses a TPM or Secure [Enclave](https://support.apple.com/guide/security/secure-enclave-sec59b0b31ff/1/web/1) or [Element](https://developers.google.com/android/security/android-ready-se) for rate limiting attempts to enter the encryption passphrase. You should avoid sharing your computer with people you don't trust, because most desktop operating systems do not encrypt data separately per-user.
```
jonaharagon, line 51: Need to check the accuracy of this, AFAIK Windows does not have a secure verified boot implementation (for a number of reasons, probably most importantly because Microsoft has signed shims that allow the loading of any operating system).

docs/basics/common-threats.en.md
```diff
@@ -157,6 +157,35 @@ Focusing solely on the privacy policies and marketing of a tool or provider can
 
 The privacy policies and business practices of a provider you choose are very important, but should be considered secondary to technical guarantees of your privacy: Don't elect to merely shift trust to another provider when trusting a provider isn't a requirement at all.
 
+:material-numeric-4-circle: **Complicated is better**
+
+We often see people describing privacy threat models that are overly complex. Often these solutions include problems like many different email accounts or complicated setups with a lot of moving parts and conditions. The replies are usually answers to, "What is the best way to do X?".
+
+Best doesn't always necessarily mean a person is after an infallible solution with dozens of conditions, these solutions are often difficult to work with realistically, for the reason we said above, security often comes at the cost of convenience. Below we provide some tips:
```
jonaharagon, line 164:
```suggestion
Finding the "best" solution for yourself doesn't always necessarily mean you are after an infallible solution with dozens of conditions, these solutions are often difficult to work with realistically. As we discussed previously, security often comes at the cost of convenience. Below we provide some tips:
```

docs/basics/common-threats.en.md
```diff
@@ -157,6 +157,35 @@ Focusing solely on the privacy policies and marketing of a tool or provider can
...
+So how might this look?
+
+One of the clearest threat models is one where people *know who you are* and one where they do not. There will always be situations where you must declare your legal name and places where you can get away without doing so.
+
+1. **Known identity** - A known identity is used for things where you must declare your name. There are many such legal documents and contracts where a legal identity is required. This could range from opening a bank account, signing a property lease, obtaining a passport, a customs declaration when importing an item or otherwise dealing with your Government. These things will usually always lead back credentials such as credit cards, credit rating checks, account numbers and possibly physical addresses.
+
+We don't suggest using a VPN or Tor for any of these things as your identity is already known through other means.
+
+``` tip
+    If buying an item the use of a [parcel locker](https://en.wikipedia.org/wiki/Parcel_locker) may help keep your physical address private.
```
jonaharagon, line 179:
```suggestion
1. **Known identity** - A known identity is used for things where you must declare your name. There are many such legal documents and contracts where a legal identity is required. This could range from opening a bank account, signing a property lease, obtaining a passport, a customs declaration when importing an item or otherwise dealing with your Government. These things will usually always lead back credentials such as credit cards, credit rating checks, account numbers and possibly physical addresses.

   We don't suggest using a VPN or Tor for any of these things as your identity is already known through other means.

   !!! tip
       If buying an item the use of a [parcel locker](https://en.wikipedia.org/wiki/Parcel_locker) may help keep your physical address private.
```

docs/basics/common-threats.en.md
```diff
@@ -157,6 +157,35 @@ Focusing solely on the privacy policies and marketing of a tool or provider can
...
+
+2. **Unknown identity** - An unknown identity could be a stable pseudonym that you regularly use. It is not anonymous because it doesn't change. If you're a part of an online community you may wish to retain persona that others know. The reason this is not anonymous is because if monitored over a period of time details about the owner may reveal further information, such as the way they write (lingustics), general knowledge about topics of interest etc.
+
+You may wish to use a VPN for this to mask your IP address. Financial transactions are more difficult and for this we'd suggest using anonymous cryptocurrencies such as Monero. Employing alt-coin shifting may also help disguise where your currency originated. Typically exchanges require KYC (know your customer) to be completed before they will allow you to exchange fiat currency into any kind of cryptocurrency. Local meet-up options may also be a solution, however those often are more expensive and sometimes also require KYC.
```
jonaharagon, line 183:
```suggestion
2. **Unknown identity** - An unknown identity could be a stable pseudonym that you regularly use. It is not anonymous because it doesn't change. If you're a part of an online community you may wish to retain persona that others know. The reason this is not anonymous is because if monitored over a period of time details about the owner may reveal further information, such as the way they write (lingustics), general knowledge about topics of interest etc.

   You may wish to use a VPN for this to mask your IP address. Financial transactions are more difficult and for this we'd suggest using anonymous cryptocurrencies such as Monero. Employing alt-coin shifting may also help disguise where your currency originated. Typically exchanges require KYC (know your customer) to be completed before they will allow you to exchange fiat currency into any kind of cryptocurrency. Local meet-up options may also be a solution, however those often are more expensive and sometimes also require KYC.
```

docs/basics/common-threats.en.md
```diff
@@ -157,6 +157,35 @@ Focusing solely on the privacy policies and marketing of a tool or provider can
...
+
+3. **Anonymous identity** - Anonymous identities are difficult to maintain over long periods of time for even the most experienced. They should be short-term and short lived identities which are rotated regularly.
+
+Using Tor can help with this, it's also worth noting greater anonymity is possible through asynchronous (not real time communication). Real time communication is vulnerable to typing analysis patterns more than a slab of text distributed on a forum, email) etc that you've had time to think about, maybe even put through a translator and back again.
```
jonaharagon, line 187:
```suggestion
3. **Anonymous identity** - Anonymous identities are difficult to maintain over long periods of time for even the most experienced. They should be short-term and short lived identities which are rotated regularly.

   Using Tor can help with this, it's also worth noting greater anonymity is possible through asynchronous (not real time communication). Real time communication is vulnerable to typing analysis patterns more than a slab of text distributed on a forum, email) etc that you've had time to think about, maybe even put through a translator and back again.
```

jonaharagon (changes requested):
smh @ people approving PRs without reviewing what they actually look like.

lx-is (commented):
I think this "Complicated is better" misconception should be the first one listed, given it's probably the most important.

docs/basics/common-threats.en.md
dngray: > I wouldn't make this change, I'm not sure what it adds to the article outside of unnecessary over-explanation.

The point that I am stressing there is that not all Google products are equal and they have different Privacy Policies.

jonaharagon (approved):

PR #1315, opened 2022-05-24
Add AdGuard Home

27lx:
AdGuard home was suggested in the Pi-hole discussion (https://github.com/privacyguides/privacyguides.org/discussions/1098#discussioncomment-2627467). I looked into it and it seems to offer [some advantages](https://github.com/AdguardTeam/AdGuardHome#comparison-pi-hole) over Pi-hole, though overall they're quite similar pieces of software. I think it's best if we list both options as it opens up readers to both good options.

Relevant: https://github.com/privacytools/privacytools.io/issues/1298


docs/dns.en.md
```diff
@@ -132,6 +132,23 @@ Encrypted DNS proxy software provides a local proxy for the [unencrypted DNS](ba
 
 ## Self-hosted Solutions
 
+A self-hosted DNS solution is useful for providing filtering on controlled platforms, such as Smart TVs and other IOT devices, as no client-side software is needed.
+
+### AdGuard Home
+
+!!! recommendation
+
+    ![AdGuard Home logo](assets/img/dns/adguard-home.svg){ align=right }
```
jonaharagon, line 141: Can we not just reuse the adguard SVG we already have?

27lx: I thought about that, but I reasoned that they should just be treated as separate products despite sharing the same logo. It's only an extra 500B.

jonaharagon: Slippery slope, Lexi!

jonaharagon (approved):

PR #1331, opened 2022-05-25
Fixes for the Linux Page

tommytran732:
I don't like polarbears. 

On a serious note though, the without the last sentence, the GRSecurity part sounds weird. As for the LKRG kernel module, the point here is not that it is available on KickSecure, the point here is that if you use Debian you should just add the KickSecure repo and install it since it is fairly easy anyways. Same for Arch and possibly Fedora in the future.

I also updated the mountpoint hardening section since `noexec` on /home and /root breaks a lot of stuff.

### DIFF
--- a/docs/linux-desktop/hardening.en.md
+++ b/docs/linux-desktop/hardening.en.md
@@ -32,15 +32,17 @@ Do **not** disable unprivileged user namespaces if you use software that relies 
 
 ## Linux-Hardened
 
-Some distributions like Arch Linux have the [linux-hardened](https://github.com/anthraxx/linux-hardened), kernel package. It includes [hardening patches](https://wiki.archlinux.org/title/security#Kernel_hardening) and more security-conscious defaults. Linux-Hardened has `kernel.unprivileged_userns_clone=0` disabled by default. See the [warning above](#kernel-hardening) about how this might impact you.
+Some distributions like Arch Linux have the [linux-hardened](https://github.com/anthraxx/linux-hardened), kernel package. It includes [hardening patches](https://wiki.archlinux.org/title/security#Kernel_hardening) and more security-conscious defaults. Linux-Hardened has `kernel.unprivileged_userns_clone=0` disabled by default. See the [note above](#kernel-hardening) about how this might impact you.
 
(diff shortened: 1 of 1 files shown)


blacklight447 (approved):

docs/linux-desktop/hardening.en.md
```diff
@@ -32,15 +32,17 @@ Do **not** disable unprivileged user namespaces if you use software that relies
...
+
+If you can get LKRG and maintain module updates it provides a worthwhile improvement to security. Debian based distributions can get the LKRG DKMS from KickSecure's secure repository and the [KickSecure documentation](https://www.kicksecure.com/wiki/Linux_Kernel_Runtime_Guard_LKRG) has instructions on how this can be achieved. There is no LKRG package for Fedora yet, however the Qubes OS project has a COPR repository which [may become]((https://github.com/QubesOS/qubes-issues/issues/5461) part of the main distribution in the future. Archlinux based systems provide LKRG DKMS modules via an [AUR package](https://aur.archlinux.org/packages/lkrg-dkms).
 
 ## GRSecurity
 
-GRSecurity is a set of kernel patches that attempt to improve security of the Linux kernel. Open source  access, however, requires [subscription to a paid model](https://grsecurity.net/passing_the_baton).
+GRSecurity is a set of kernel patches that attempt to improve security of the Linux kernel.  It requires [payment to access]((https://github.com/QubesOS/qubes-issues/issues/5461) the code. If you have access to GRSecurity, it is recommended that you use it.
```
dngray, line 45: > If you have access to GRSecurity, it is recommended that you use it.

I kinda want to avoid writing "if you can do this do it", everywhere because it is superfluous. It is implied by the fact it is on our hardening page.

The hardening guide was never intended to be a "you must do all these things or else" kind of language, the reason is because the vast majority won't, and that's fine, hardening systems is often a lot of work and requires continued maintenance and effort.

dngray (changes requested):

hook9 (changes requested):
GRSecurity should be in a separate discussion/PR in my opinion.

dngray (approved):

PR #1342, opened 2022-05-28
Flawed E2EE in Web Applications

dngray:
Superseeds: https://github.com/privacyguides/privacyguides.org/pull/1052

### DIFF
--- a/docs/basics/web-e2ee.en.md
+++ b/docs/basics/web-e2ee.en.md
@@ -1,0 +1,28 @@
+---
+title: "E2EE in Web Applications"
+---
+[E2EE](https://en.wikipedia.org/wiki/End-to-end_encryption) (End to End Encryption) mathematically ensures the security of exchanged data in transit (common security properties include integrity, authenticity, confidentiality, deniability, and forward secrecy). The most common method of what people perceive as E2EE is done between users' web browsers and websites (e.g. [`https://google.com`](https://google.com)) using [HTTPS](https://en.wikipedia.org/wiki/HTTPS). Although, the definition of E2EE is loosely defined and varies definitions, prompting [a research paper by cryptographers questioning the "end"ness](https://eprint.iacr.org/2022/449).
+
+Web applications (a type of client) are continuously served from servers, composed of HTML, CSS, and JS to provide structure, styles, and interactivity to the web.
+
+## Problematic Design Limitations
+
+[Nadim Kobeissi](https://en.wikipedia.org/wiki/Nadim_Kobeissi) released [An Analysis of the ProtonMail Cryptographic Architecture](https://eprint.iacr.org/2018/1121) paper, covering ProtonMail specifically. However, the same principles can be applied to other web applications (e.g. Tutanota, BitWarden, Threema, etc.) which attempt to provide E2EE through web browsers.
+
+<iframe width="100%" style="height:50vh"
+  src="https://www.youtube-nocookie.com/embed/DM1tPmxGY7Y"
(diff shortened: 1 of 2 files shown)


bkil:
Okay, let me copy the concerns here because as I see it is still unaddressed in the post.

Web (-Browser based) Apps, especially if FOSS and engineered from the ground up for offline use via Service Workers, are for most intents and purposes equivalent to non-browser based apps. It's just another containerization technology and runtime.

All the same kind of attacks would apply to desktop or native mobile apps that support live (or at least daily) updating through its own servers (or plugins through its own store). It's TOFU either way.

At the same time, nothing keeps you from self-hosting the web apps you use either on a trusted server or even on localhost and review all code changes before running them.

samuel-lucas6:
> Web (-Browser based) Apps, especially if FOSS and engineered from the ground up for offline use via Service Workers, are for most intents and purposes equivalent to non-browser based apps. It's just another containerization technology and runtime.

I agree.

> All the same kind of attacks would apply to desktop or native mobile apps that support live (or at least daily) updating through its own servers (or plugins through its own store). It's TOFU either way.

That's also true.

> At the same time, nothing keeps you from self-hosting the web apps you use either on a trusted server or even on localhost and review all code changes before running them.

Few people self-host and can/want to spend their time reviewing changes.

> What are you trying to prove with this post?

I think [this](https://blog.bren2010.io/blog/criticism-of-javascript-crypto) blog post reflects your points well. If this blog post goes ahead, I think caution needs to be paid to the tone. I believe the point is to warn people about this issue in general and/or suggest people install programs and manually update them.

dngray:
> I think [this](https://blog.bren2010.io/blog/criticism-of-javascript-crypto) blog post reflects your points well. If this blog post goes ahead, I think caution needs to be paid to the tone. I believe the point is to warn people about this issue in general and/or suggest people install programs and manually update them.

Agreed, in any case this one was @realguyman's project, I just moved it to the main repo so that we can rebase it every now and again without having to fix annoying merge conflicts.

bkil:
I checked the PDF and its information is very dated.

The following assumption is invalid in section 5.1.1:

> this untrusted server P must serve an authentic OpenPGP implementation J every time A logs into ProtonMail or, in some cases, multiple times in between A same single ProtonMail session. Since P is untrusted and since no authentication mechanism is implemented to check for the correctness of J, P can arbitrarily and untraceably compromise any information that A sends as part of her ProtonMail session"

A webapp can be implemented in a way so as that the offline part (for example, within a Service Worker) is only ran once, and only pulls in additional resources using SRI or after checking their cryptographic signature (in case of updates). This places trust equivalent to installing a desktop software from the website of a vendor. The same could have been implemented in various other (more twisted) ways even decades ago, but I better not go into that here.

Examine the following statement under 2.1 "Security Assumptions":

> Untrusted Server. We assume that P is untrusted and could act with
the intent to recover encrypted communications between clients A, B, S

It is equivalent to say that the software vendor of the mobile app or desktop app is untrusted in the exactly same way, along with its update mechanism.

bkil:
From all of this FUD, I think one single point is unavoidable and definitely true: emails are routed basically clear text around the world and can be also assumed to be stored as such on most nodes.

It does not give any privacy guarantees that ProtonMail promises to encrypt them before storing them. They have confessed to not doing that multiple times as well.

So, all in all, a big fat warning is in order for so called encrypted mailbox providers in general.

However, it **is** actually feasible to implement E2EE properly in web apps regardless of how ProtonMail is implemented right now, so at a minimum, the article will need to be renamed.

PR #1351, opened 2022-05-28
Minify HTML & CSS

### LINKED ISSUE #1343: Minify HTML & CSS
Taking a look at the HTML output served, it's quite a mess. Hundreds of lines of whitespace and really strange indentation. Solution to this would probably be to use [mkdocs-minify-plugin](https://github.com/byrnereese/mkdocs-minify-plugin), if possible. The base CSS files are already minified, however extra.css isn't.

https://user-images.githubusercontent.com/100533118/170815351-020bede8-2387-4df4-b12d-8c2fa48be5aa.mp4

jonaharagon:
Closes #1343


mkdocs.yml
```diff
@@ -77,6 +77,13 @@ watch:
   - includes
 
 plugins:
+  - minify:
+      minify_html: true
+      htmlmin_opts:
+          remove_comments: true
+      minify_css: true
+      css_files:
+          - assets/stylesheets/extra.css
```
27lx, line 86:
```suggestion
          - assets/stylesheets/extra.css
          - overrides/home.css
```

jonaharagon: Minifying this file breaks the homepage for some reason.

dngray (changes requested):
I spotted a bug with the downloads button.

Before:



After:

jonaharagon:
Oh, that is the same issue we had with minifying the homepage CSS. I don't think we'll be able to minify CSS in this PR then, only HTML. I'll push a change later.

dngray (approved):

PR #1349, opened 2022-05-28
add grid layout to download links

noClaps:
Updated download links

Resolves: # 

- Added text to download links
- Now uses grid layout

PR #1372, opened 2022-06-01
Add grid layout for download links

noClaps:
Superseeds: #1349

Updated download links

Resolves: # 

- Added text to all download links
- Now uses grid layout
- Removed `{.card-link}` from all download links

### DIFF
--- a/docs/android.en.md
+++ b/docs/android.en.md
@@ -164,8 +164,8 @@ Fairphone markets their devices as receiving 6 years of support. However, the So
 
     ??? downloads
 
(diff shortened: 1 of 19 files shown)


27lx (dismissed):
Couple things
- The font should be made smaller (maybe 1em). Currently it is bigger than the download title itself.
- A solution should be found for the annotations that are occasionally used in the downloads menu to give additional context. Example in the Brave card: it is giving additional context to the Linux download, but with this it's separated quite far apart from it https://privacyguides-privacyguides-org-preview-pr-1372.surge.sh/browsers/#brave The annotation box also gets separated from the + icon

27lx (approved):

PR #1377, opened 2022-06-02
List recommendations in alphabetical order

jonaharagon:


27lx:
I don't see the need for alphabetical ordering. Higher positions on the page inherently hold more value, and so it doesn't really make sense to just leave it up to the name of a product or service.

docs/multi-factor-authentication.en.md
```diff
@@ -61,6 +38,29 @@ For the models which support HOTP and TOTP, there are 3 slots for HOTP and 15 fo
 
     The Nitrokey app, while compatible with Librem Keys, requires `libnitrokey` version 3.6 or above to recognize them. Currently, the package is outdated on Windows, macOS, and most Linux distributions' repository, so you will likely have to compile the Nitrokey app yourself to get it working with the Librem Key. On Linux, you can obtain an up-to-date version from [Flathub](https://flathub.org/apps/details/com.nitrokey.nitrokey-app).
 
+### YubiKey
```
dngray, line 41: I think Yubikey, should be #1. It still does a lot more than competing products, at least until the Nitrokey 3 is available.

docs/self-contained-networks.en.md
```diff
@@ -4,31 +4,27 @@ icon: material/security-network
 ---
 These networks are designed to keep your traffic anonymous.
 
-## Tor
+## The Freenet Project
```
dngray, line 7: Not really sure how I feel about this, in most cases I think Tor is what people want.

Also not sure why this starts with "The". I think this should just be "Freenet".

docs/vpn.en.md
```diff
@@ -27,6 +27,59 @@ Find a no-logging VPN operator who isn’t out to sell or read your web traffic.
 
     Our recommended providers use encryption, accept Monero, support WireGuard & OpenVPN, and have a no logging policy. Read our [full list of criteria](#our-criteria) for more information.
 
+### IVPN
```
dngray, line 30: Regarding this one, not opposed to IVPN being number 1.

The reason for that is because they have regular audits, and they cover full network penetration testing not just their apps.

blacklight447 (approved):
I don't think we should change the order based on which service is better. This will cause bickering every moment a new service gets added.

Being listed on PG should just be a big gaint checkmark that its a good service, and its up to you to decide which one is better for your personal usecase.

The changes in here look good to me, and are to my opinion ready for merging.

PR #1391, opened 2022-06-05
Use bigger icons for certain services

27lx:
before

after


blacklight447:
Why remove the mailbox.org text though?

dngray:
> Why remove the mailbox.org text though?

Think it's a bit more consistent with the others.

PR #1392, opened 2022-06-06
Move snowflake embed into a collapsable admonition

27lx:
The design of the Snowflake embed doesn't really fit in with the site design and it also takes up unnecessary screen real estate for those who don't make use of it, and so it's probably best to put it behind a collapsable admonition.


blacklight447 (approved):
Looks great!

PR #1399, opened 2022-06-06
Revert "List recommendations in alphabetical order (#1377)"

dngray:
This reverts commit 42dd2c57a4750c9cecce5fa72736d1fe90458373.

I don't think we should do this. Discussion https://github.com/privacyguides/privacyguides.org/discussions/1396

This was merged without addressing the feedback from @27lx https://github.com/privacyguides/privacyguides.org/pull/1377#issuecomment-1144436702 and me https://github.com/privacyguides/privacyguides.org/pull/1377#pullrequestreview-992992055.

### DIFF
--- a/docs/calendar-contacts.en.md
+++ b/docs/calendar-contacts.en.md
@@ -8,24 +8,6 @@ Calendaring and contacts are some of the most sensitive data posess. Use only pr
 
 These products are included with an subscription with their respective [email providers](email.md).
 
-### Proton Calendar
-
-!!! recommendation
-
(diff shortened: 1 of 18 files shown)


tommytran732:
> As already explained, non-alphabetical ordering is unexpected behavior in most cases and will only confuse users. The problem you see is a symptom of much larger issues, will discuss this at https://github.com/orgs/privacyguides/discussions/1396#discussioncomment-2890811. This should not be reverted :)

We are reverting it by slowly fixing the pages. The alphabetical order is not okay and the reasoning for the order is already explained in several of the pages you changed before you even made them alphabetical.

PR #1400, opened 2022-06-06
Add Markor

dngray:


blacklight447:
Any link to the discussion?

dngray:
> Any link to the discussion?

There is not, as there isn't much to discuss. It doesn't make any network connections, and has been maintained/developed for years. I've also used it for years, and for Android if you're looking for something like org-mode it's a good idea.

In particular it ticks a number of boxes, works well when used in conjunction with other synchronization tools like Syncthing etc, and is a non-proprietary format, easily parsable in any text editor.

The main issue with things like [opentasks](https://github.com/dmfs/opentasks) they use CalDAV which means no E2EE.

You can of course use Etesync, but that requires hosting it somewhere.

A very portable-but secure and simple way of synchronizing a to-do list would be a file in a syncthing share, E2EE can be handled by your FDE option and as far as network connections go, there's nothing outside your LAN.

matchboxbananasynergy:
This app is quite nice and useful, however, there is one thing that makes me not want it recommended on Privacy Guides.

The app has been rejected from Google Play because it doesn't meet the criteria for needing the MANAGE_EXTERNAL_STORAGE permission.

The developer doesn't seem interested in changing the way the app works, which can have security implications, as apps like these now have ways of accessing files that doesn't require storage access.

More information on this can be found here:
https://github.com/gsantner/markor/pull/1521

I think that it makes sense for our Android app recommendations to need to at least meet the minimum criteria to be listed on the Google Play Store, with notable exceptions that are not due to what permission the app requests, but something else (such as NewPipe or App store applications). This is probably off-topic here though, and I should open a discussion about general criteria for Android app recommendations.

dngray:
I did wonder why it wasn't in Google Play, that's unfortunate, maybe it's better if we don't list it then.

dngray:
I've added this to the criteria https://github.com/privacyguides/privacyguides.org/wiki/Page-Criteria#mobile-apps

gsantner:
> The app has been rejected from Google Play because it doesn't meet the criteria for needing the MANAGE_EXTERNAL_STORAGE permission.

@dngray @matchboxbananasynergy 

I tried to appeal and was ignored. Markor is also a file manager and has file management capabilities to manage any files on the device. So thats the reason why MANAGE_EXTERNAL_STORAGE is wanted & used here. Sad to hear that also open source developers shed a bad light on this, like it is stealing all the peoples data and invades privacy - which Markor does not, I put a strong focus on having no online feature, no tracking, no data collection or anything alike.

matchboxbananasynergy:
@gsantner 

Thank you for the hard work that you've put in Markor. I hope that you do not take our decision to not go ahead with including your app on Privacy Guides as an attack.

I've read the issue on your repository (the one mentioned in my comment above) and fully understand your reasons for wanting to do it the way you have.

On the other hand, I, as a non-developer and perhaps naively, don't really see why this permission is needed, and why Markor needs to act as a file manager when it could just use the file picker itself, making the need for an extra permission null.

I wish you the best going forward.

P.S. Thank you for pointing out the SSL certificate issue. It was fixed soon after it occurred.

Zanthed:
@gsantner Perhaps there are far more concerns with your app that I don't think you can really excuse here beyond the unnecessarily invasive MANAGE_EXTERNAL_STORAGE permission that may have contributed to you being rejected on Play Store and scream serious supply chain concerns, privacy, and security issues in the future.

- You have a *terribly* outdated [Android Gradle version](https://github.com/gsantner/markor/blob/master/build.gradle#L16) (3.6.3)
- A [very old Gradle Wrapper (5.6.4) version that isn't pinning the SHA256 hash and is not verifying the Gradle Wrapper jar file in GitHub Actions](https://github.com/gsantner/markor/blob/master/gradle/wrapper/gradle-wrapper.properties#L3)
- You are only [targeting SDK 29](https://github.com/gsantner/markor/blob/master/build.gradle#L20) (your compile SDK variable is for some reason used the same as target SDK when they aren't the same thing) which is actually no longer allowed by Play Store and [you *have* to target SDK 31 to get on the Play Store or release updates to the Play Store](https://developer.android.com/google/play/requirements/target-sdk) which may have contributed to being rejected on Play Store
- You are [not using AndroidX](https://github.com/gsantner/markor/blob/master/gradle.properties) (android.useAndroidX) which are far superior, easier to use APIs with full built-in backwards compatibility and is [phasing out the legacy Android Support libraries](https://developer.android.com/jetpack/androidx#:~:text=AndroidX%20is%20a%20major%20improvement%20to%20the%20original%20Android%20Support%20Library%2C%20which%20is%20no%20longer%20maintained.) (despite the version variable `version_library_appcompat` commented with "AndroidX", it is not being used for AndroidX [here](https://github.com/gsantner/markor/blob/master/build.gradle#L40) and is using the legacy support libraries)
- Your [gradlew scripts are not up to date to the latest ones](https://github.com/gsantner/markor/blob/master/gradlew#L4) as noted by the LICENCE notice (you can see an example of the most up to date one here https://github.com/GrapheneOS/Camera/blob/main/gradlew which was actually updated with gradle wrapper 7.5 recently)
- The [jcenter repo still in use ](https://github.com/gsantner/markor/blob/master/build.gradle#L53)despite the fact that [it's legacy and abandoned entirely](https://blog.gradle.org/jcenter-shutdown)
- You are [pulling in an archived, outdated, and abandoned dependency](https://github.com/gsantner/markor/blob/master/build.gradle#L40) (https://github.com/dcendents/android-maven-gradle-plugin)
- You [explicitly disabled minifying builds](https://github.com/gsantner/markor/blob/master/app/build.gradle#L70)  
- You have far more concerning permissions being requested like [REQUEST_INSTALL_PACKAGES](https://github.com/gsantner/markor/blob/9f54d48c18339f741e73eef2e0986d3e76ff86f5/app/src/main/AndroidManifest.xml#L25), [RECORD_AUDIO](https://github.com/gsantner/markor/blob/9f54d48c18339f741e73eef2e0986d3e76ff86f5/app/src/main/AndroidManifest.xml#L24) (a file manager and text editor recording audio?), not only [READ_EXTERNAL_STORAGE but FULL_EXTERNAL_STORAGE too](https://github.com/gsantner/markor/blob/9f54d48c18339f741e73eef2e0986d3e76ff86f5/app/src/main/AndroidManifest.xml#L19-L20) (unsure where MANAGE_EXTERNAL_STORAGE came from as that was [added in API 30](https://developer.android.com/reference/android/Manifest.permission#MANAGE_EXTERNAL_STORAGE) when you're targeting SDK 29), and [INTERNET](https://github.com/gsantner/markor/blob/9f54d48c18339f741e73eef2e0986d3e76ff86f5/app/src/main/AndroidManifest.xml#L21)
- Your [manifest calls for a very legacy `requestLegacyExternalStorage`](https://github.com/gsantner/markor/blob/master/app/src/main/AndroidManifest.xml#L47) and [`usesCleartextTraffic` allowing unencrypted, plaintext HTTP traffic](https://github.com/gsantner/markor/blob/master/app/src/main/AndroidManifest.xml#L51)
- You have activities in your manifest that seem to be [requesting the linter to use legacy Android versions like Lollipop (Android 5.x)](https://github.com/gsantner/markor/blob/master/app/src/main/AndroidManifest.xml#L175) and [Android JellyBeam (4.3)](https://github.com/gsantner/markor/blob/master/app/src/main/AndroidManifest.xml#L17)

And I haven't even decided to go through the code here.

This is very clearly a legacy app and is way out of scope within PrivacyGuides.

cc @dngray @matchboxbananasynergy

gsantner:
Have you checked the androidsdk-update branch even? Thats whats submitted to gplay, with sdk31-33.

Permissions, have you read the readme? Like that theres a feature for voicenotes? Like theres a feature to install APK files? 

Anyway, I see theres no need to discuss you have different view of privacy. A lot pointd are rather mixed up with security.

PR #1404, opened 2022-06-07
Remove Write.as

tommytran732:
https://github.com/privacyguides/privacyguides.org/discussions/1395


dngray:
That productivity page has office kind of tools on it.

I agree write.as doesn't inherently have anything privacy friendly over it when compared to [other platforms](https://github.com/orgs/privacyguides/discussions/1395#discussioncomment-2889417) although one could argue you might be better off using some write.as instance over something like Medium.

There of course is the issue of the instance going down, and you being left with nothing. I think if you're hosting a blog you really either want to ensure the domain is yours, or it's not going anywhere.

dngray:
Had some more thoughts about this.

I think move the office suites to a page about office suites, and maybe say something about these being local options because"

1*el4nHcOQdfHxzzcTC99jDw.

Secondly I think I would have a section on there for office suites that are E2EE, if we can find any. I was sure one was recommended in the past.

Microsoft Office you may want to mention on the Windows page [because of MDAG](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-application-guard/md-app-guard-overview).

> For Microsoft Office, Application Guard helps prevents untrusted Word, PowerPoint and Excel files from accessing trusted resources. Application Guard opens untrusted files in an isolated Hyper-V-enabled container. The isolated Hyper-V container is separate from the host operating system. This container isolation means that if the untrusted site or file turns out to be malicious, the host device is protected, and the attacker can't get to your enterprise data. For example, this approach makes the isolated container anonymous, so an attacker can't get to your employee's enterprise credentials.

This somewhat crosses over on the purpose of things like Qubes-OS.

tommytran732:
> Had some more thoughts about this.
> 
> I think move the office suites to a page about office suites, and maybe say something about these being local options because:
> 
> 1*el4nHcOQdfHxzzcTC99jDw.
> 
> Secondly I think I would have a section on there for office suites that are E2EE, if we can find any. I was sure one was recommended in the past.
> 
> Microsoft Office you may want to mention on the Windows page [because of MDAG](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-application-guard/md-app-guard-overview).
> 
> > For Microsoft Office, Application Guard helps prevents untrusted Word, PowerPoint and Excel files from accessing trusted resources. Application Guard opens untrusted files in an isolated Hyper-V-enabled container. The isolated Hyper-V container is separate from the host operating system. This container isolation means that if the untrusted site or file turns out to be malicious, the host device is protected, and the attacker can't get to your enterprise data. For example, this approach makes the isolated container anonymous, so an attacker can't get to your employee's enterprise credentials.
> 
> This somewhat crosses over on the purpose of things like Qubes-OS.
> 
> The pads section maybe could be moved to their own page.

Sure but say this in the discussion not the PR lol

27lx (approved):

PR #1405, opened 2022-06-07
Update Search Engine page

tommytran732:
DDG > Brave > SearcXNG > Startpage.

The reasoning for this is literally already listed for most of them. DDG supports no JS mode, making it the best of the bunch. Brave has anonymous analytics enabled by default so it's worse than DDG.

The only thing that's not mentioned is that StartPage sometimes block VPN or Tor IPs, so it is not as friendly and therefore is put at the bottom. This PR added that.

### DIFF
--- a/docs/search-engines.en.md
+++ b/docs/search-engines.en.md
@@ -8,6 +8,25 @@ The recommendations here are based on the merits of each service's privacy polic
 
 Consider using a [VPN](vpn.md) or [Tor](https://www.torproject.org/) if your threat model requires hiding your IP address from the search provider.
 
+## DuckDuckGo
+
+!!! recommendation
+
+    ![DuckDuckGo logo](assets/img/search-engines/duckduckgo.svg){ align=right }
+
+    **DuckDuckGo** is one of the more mainstream private search engine options. Notable DuckDuckGo search features include [bangs](https://duckduckgo.com/bang) and many [instant answers](https://help.duckduckgo.com/duckduckgo-help-pages/features/instant-answers-and-other-features/). The search engine relies on a commercial Bing API to serve most results, but it does use numerous [other sources](https://help.duckduckgo.com/results/sources/) for instant answers and other non-primary results.
+
+    DuckDuckGo is the default search engine for the Tor Browser and is one of the few available options on Apple’s Safari browser.
+
+    [:octicons-home-16: Homepage](https://duckduckgo.com){ .md-button .md-button--primary }
+    [:pg-tor:](https://duckduckgogg42xjoc72x3sjasowoarfbgcmvfimaftt6twagswzczad.onion){ .card-link title=Onion }
+    [:octicons-eye-16:](https://duckduckgo.com/privacy){ .card-link title="Privacy Policy" }
+    [:octicons-info-16:](https://help.duckduckgo.com/){ .card-link title=Documentation}
+
+DuckDuckGo is based in the :flag_us: United States. Their [privacy policy](https://duckduckgo.com/privacy) states they **do** log your searches for product improvement purposes, but not your IP address or any other personally identifying information.
+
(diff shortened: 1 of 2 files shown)


docs/search-engines.en.md
```diff
@@ -85,4 +104,7 @@ Startpage is based in the :flag_nl: Netherlands. According to their [privacy pol
 
 Startpage's majority shareholder is System1 who is an adtech company. We don't believe that to be an issue as they have an distinctly separate [privacy policy](https://system1.com/terms/privacy-policy). The Privacy Guides team reached out to Startpage [back in 2020](https://web.archive.org/web/20210118031008/https://blog.privacytools.io/relisting-startpage/) to clear up any concerns with System1's sizeable investment into the service. We were satisfied with the answers we received.
 
+!!! information
```
27lx, line 107: > The only thing that's not mentioned is that StartPage sometimes block VPN or Tor IPs, so it is not as friendly and therefore is put at the bottom. This PR added that.

This was/is already mentioned in the Startpage card.

27lx (changes requested):
look at the preview lol. the page is all messed up

duckduckgo is on there twice and the startpage card is improperly formatted

docs/search-engines.en.md
```diff
@@ -73,9 +73,9 @@ When you are using a SearXNG instance, be sure to go read their privacy policy.
...
+    **Startpage** is a private search engine known for serving Google search results. Startpage's flagship feature is [Anonymous View](https://www.startpage.com/en/anonymous-view/), which puts forth efforts to standardize user activity to make it more difficult to be uniquely identified. Unlike the name suggests, the feature should not be relied upon for anonymity. If you are looking for anonymity, use the Tor Browser instead. The feature can be useful for hiding some network and browser properties—see the [technical document](https://support.startpage.com/hc/en-us/articles/4455540212116-The-Anonymous-View-Proxy-technical-details).
 
-    Startpage has been known to refuse access to those using a VPN service or Tor, so your mileage may vary.
```
dngray, line 78: I've not known them to refuse outright, though they do sometimes require a captcha.

I think this can be integrated with the note below.

dngray (changes requested):

docs/search-engines.en.md
dngray: Hah same as below https://github.com/privacyguides/privacyguides.org/pull/1405#pullrequestreview-999143798

27lx (approved):

dngray (approved):

hook9:
I honestly think what DuckDuckGo saves is worse than Brave's usage metrics.

[DuckDuckGo](https://duckduckgo.com/privacy#s4):
> We also save searches, but again, not in a personally identifiable way, as we do not store IP addresses or unique User agent strings. We use aggregate, non-personal search data to improve things like misspellings.

[Brave](https://search.brave.com/help/usage-metrics):
> Number of daily/weekly/monthly visits
> Number of returning visits
> Number of search queries per day
> How long you’ve been using Brave Search
> Average query length
> How many users have chosen to leave feedback about Brave Search
> The operating systems people use when they visit (e.g. macOS, Windows, etc)
> The browser you’ve visited from (e.g. Brave, Chrome, Safari, etc)

For search queries that only works for some queries, and it can get really murky, for example if the search query contains personal information or anything that can tie it back to the user.

27lx:
> I honestly think what DuckDuckGo saves is worse than Brave's usage metrics.

I do agree with this.

dngray:
Oh, damn I didn't see this before I merged.

Hmm, I think with Brave/DDG though it might be pretty close because they both have .onion, Brave Search does have [anonymous usage metrics](https://search.brave.com/help/usage-metrics) that you have to disable.

27lx:
> Brave Search does have [anonymous usage metrics](https://search.brave.com/help/usage-metrics) that you have to disable.

The usage metrics are still better, I'd argue, than saving every search. And at least the usage metrics can be disabled, whereas Duck's search logs are mandatory.

dngray:
Point taken, I put in a PR https://github.com/privacyguides/privacyguides.org/pull/1405.

Opt out is better than no opt out.

evil-user:
Heard duckduckgo favors microsoft's trackers, https://www.theregister.com/2022/06/17/brave_duckduckgo/

PR #1406, opened 2022-06-07
Fix 2FA ordering

tommytran732:
The NitroKey should never be listed before the Yubikey. The reasoning has always been provided since day 1. This PR simply reverts the alphabetical ordering.

I added 1 single sentence in the warning in case it is not clear enough.


dngray:
I do agree with this, especially as the Yubikey 5 is very versatile.

The difference may become a little less obvious with the Nitrokey 3, but that's not out yet.

dngray (approved):

PR #1409, opened 2022-06-07
Update Real-Time Communication page ordering and categorization

tommytran732:
This page used to be fine how it was. We list centralized messengers first, then federated networks, then messengers with anonymous routing. The current page is a mess.

We literally explains the type of network right now below, why are we not following that for our recommendations?

Also, remember that generally, centralized messengers are the easiest to use, because you just sign up and use it. That is followed by the federated messengers, because you have to grasp the concept of the homeservers and what not. We list anonymously routed messengers last because they are generally harder to use.

### DIFF
--- a/docs/real-time-communication.en.md
+++ b/docs/real-time-communication.en.md
@@ -2,7 +2,41 @@
 title: "Real-Time Communication"
 icon: material/chat-processing
 ---
-## Cross-Platform Messengers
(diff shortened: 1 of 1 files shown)


27lx (approved):
I prefer this categorization, especially given the explanations of the types at the bottom of the page.

PR #1417, opened 2022-06-08
Refine Tor Overview article

dngray:


blacklight447 (changes requested):
We need to restore the explicit warning that you should not force https unless you know exactly what you are doing, asit contributes to fingerprinting, just having that information in a link is not enough as not everyone is gonna read.

27lx:
> We need to restore the explicit warning that you should not force https unless you know exactly what you are doing, asit contributes to fingerprinting, just having that information in a link is not enough as not everyone is gonna read.

I think the way it's written now does a good job at leading readers to the browser page where information on fingerprinting is better suited.

> Also, the videos must be listed under their own further reading header, else its inconsistent with the rest of the knowledge base. Also, the thumbnails lool just bad.

yup lol

27lx (approved):

27lx:
> We need to restore the explicit warning that you should not force https unless you know exactly what you are doing, asit contributes to fingerprinting, just having that information in a link is not enough as not everyone is gonna read.

I've gone ahead and added the warning to the browser page, which is quite visibly linked on this page.

> Also, the videos must be listed under their own further reading header, else its inconsistent with the rest of the knowledge base. Also, the thumbnails lool just bad.

The videos are now in an "Additional Resources" section and are not embedded.

blacklight447 (approved):
Looks good

PR #1421, opened 2022-06-08
Update Real Time Communication page ordering

dngray:
Signal should be in position 1.

It is simpler to use than Element, and in most cases a substitute from SMS is what people are looking for. It also produces the least amount of metadata.

### DIFF
--- a/docs/real-time-communication.en.md
+++ b/docs/real-time-communication.en.md
@@ -4,6 +4,38 @@ icon: material/chat-processing
 ---
 ## Cross-Platform Messengers
 
+### Signal
+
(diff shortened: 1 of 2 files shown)


tommytran732:
SIgnal is already in #1 in my previous PR. I think we should stick to my PR's ordering.

tommytran732:
PG needs a new release

hook9 (approved):

dngray:
> SIgnal is already in #1 in my previous PR. I think we should stick to my PR's ordering.

Link to PR? this resets things.

PR #1422, opened 2022-06-08
Update Email Providers page ordering

dngray:
Proton Mail should be #1.

The reason for this is that they have the bridge, which ensures that email is encrypted zero knowledge before uploaded to Proton Mail's servers.

Mail with Mailbox.org's zero knowledge depends on a sieve filter that will only work if the mail comes in the inbox, or gets saved in Sent.

There is quite a risk of unencrypted drafts being saved.

All options should be placed above Tutanota, as there's no way to send an encrypted email to a Tutanota user unless you are either on Tutanota, or use a "temporary inbox". Their encryption is centralized and depends on you trusting Tutanota's servers.

### DIFF
--- a/docs/email.en.md
+++ b/docs/email.en.md
@@ -16,50 +16,6 @@ For everything else, we recommend a variety of email providers based on sustaina
 
 ## Recommended Email Providers
 
-### Mailbox.org
-
-!!! recommendation
-
-    ![Mailbox.org logo](assets/img/email/mailboxorg.svg){ align=right }
-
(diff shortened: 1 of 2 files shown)


tommytran732:
Does startmail and mailbox have automatic e2ee for intradomain? If not Tuta should be above them, idk

dngray:
> Does startmail and mailbox have automatic e2ee for intradomain? If not Tuta should be above them, idk

No it does not.

dngray:
> startmail and mailbox have automatic e2ee for intradomain

mailbox.org does, if you have guard set up, and i'm pretty sure startmail does too.

dngray:
> startmail and mailbox have automatic e2ee for intradomain

Mailbox.org [supports WKD](https://mailbox.org/en/security), so any client which supports that can get the key easily.

> When you send/receive a PGP encrypted email to or from another StartMail user, the PGP public key exchange will happen automatically if you change your Settings. 

https://support.startmail.com/hc/en-us/articles/360007509378

PR #1425, opened 2022-06-08
Update VPN Providers page ordering

dngray:
Re-ordering the VPN Providers as discussion on Matrix:

1. Proton: having the most countries, using username + password + 2FA, owns 3 giant DCs and their servers and having their own network. Superior compatibility with Linux. Major deficiency: no permanent killswitch on macOS.
2. IVPN: Has 2FA, has Linux repos. Not so good stuff: not owning or operating any ASN.
3. Mullvad: Own some servers. Need to research if they own their network. Not so good stuff: no 2FA, no Linux repos

We should also look at the level of the audits. Some of these providers have only audited their apps and not full network.

### DIFF
--- a/docs/vpn.en.md
+++ b/docs/vpn.en.md
@@ -27,6 +27,62 @@ Find a no-logging VPN operator who isn’t out to sell or read your web traffic.
 
     Our recommended providers use encryption, accept Monero, support WireGuard & OpenVPN, and have a no logging policy. Read our [full list of criteria](#our-criteria) for more information.
 
+### Proton VPN
+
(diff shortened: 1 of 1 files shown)


thanharrison:
>  3. Mullvad: Own some servers. Need to research if they own their network. Not so good stuff: no 2FA, no Linux repos


They own less than 20% of their [servers](https://mullvad.net/en/servers/). They currently own 163 out of 840 servers which they operate. They only own 677 out of 840 of these servers.

tommytran732 (approved):

PR #1434, opened 2022-06-09
Reorder DivestOS and CalyxOS

d4rklynk:
I think CalyxOS should be only used for the OnePlus 8/9 and the FP4 for those who need Google Play Services.
Because if they don't, they should stick with DivestOS.
Moreover DivestOS is more hardened than CalyxOS.
That's why I think DivestOS should be recommended after GrapheneOS.

I also put this sentence :
> We only recommend CalyxOS as a harm reduction measure for the OnePlus 8T, OnePlus 9, and especially the Fairphone 4 if you need microG.

... in the box presentation (idk how you call that) of CalyxOS, bc I think it's important notice.


matchboxbananasynergy:
I think this is a good idea.

tommytran732 (approved):

PR #1472, opened 2022-06-22
Remove motto Margic Wormhole

d4rklynk:


matchboxbananasynergy:
To be honest, I'd remove that last "motto" sentence altogether, it sounds like a marketing pitch. We should just stick to a brief description of what the software does.

d4rklynk:
Yeah I agree

dngray (approved):

PR #1477, opened 2022-06-22
Reorder Brave and Firefox

d4rklynk:
Tracker: #1470 

As said in the discussion, Brave should be over Firefox as it is easier to configure for most people (ant it is a bit more secure).
I wrote a little notice above Brave to say that.

I didn't change the **Software** title to **Softwares** in the left menu as I don't know how to do it.

_Don't mind the several commits I made. I did shit with my branches_

### DIFF
--- a/docs/android/grapheneos-vs-calyxos.en.md
+++ b/docs/android/grapheneos-vs-calyxos.en.md
@@ -4,7 +4,7 @@ icon: 'material/cellphone-cog'
 ---
 ## Profiles
 
-CalyxOS includes a device controller app so there is no need to install a third-party app like Shelter.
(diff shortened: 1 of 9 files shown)


TrashPandaCodingGarbage:
What about the monopoly of chromium? For the general good of world wide web we should recommend Firefox, as its gecko-based.

d4rklynk:
> What about the monopoly of chromium? For the general good of world wide web we should recommend Firefox, as its gecko-based.

I know, but this is another problem. It is not related to privacy or security. And we still suggest it anyway.

dngray:
For #1 #2, it doesn't really make much of a difference. I don't think Brave needs any help with marketing to be honest.

PR #1478, opened 2022-06-22
Tools Page reorder and fix

d4rklynk:
- Changing introduction
- Added bitwarden-send
- Reorder DivestOS and CalyxOS
- Reorder Brave and DDG
- Rename **Encryption Tools** to **Encryption Software** (as it is written like that in the left)

**Again, don't mind the multiple commits, I messed with my branches**


matchboxbananasynergy:
There is no plural for "software", to my knowledge. Leave that as is please.

PR #1490, opened 2022-06-24
Email Client Overhaul

matchboxbananasynergy:
This PR is for the email client overhaul.

So far, I've removed Kontact and GNOME Evolution


dngray:
Do you want me to wait on merging this one until we do the more complete guide?

matchboxbananasynergy:
> Do you want me to wait on merging this one until we do the more complete guide?

Up to you. We can either get these out of the way now, or wait for Tommy's guides etc and then merge it. I'm fine with either.

mfwmyfacewhen (changes requested):
Would like the line "Canary Mail is closed source. We recommend it due to the few choices there are for email clients on iOS that support PGP E2EE." to be removed.

tommytran732:
I want to remove Mailvelope. The user really should be using a proper email client, not installing a privileged extension in their browser. If one really insists on using PGP in the browser, they can just use ProtonMail instead.

PR #1518, opened 2022-07-08
Remove CalyxOS

tommytran732:
Broader device support than GrapheneOS was the one reason Calyx was still recommended. However, even that doesn't seem to hold up anymore: https://calyxos.org/news/2022/07/06/oneplus-android-12-relock-issue/

At it stands, Calyx being outclassed by GrapheneOS on the Pixel, verified boot being broken on the OnePlus devices, and the Fairphone 4 being very broken itself, I see no reason to keep the CalyxOS recommendation.

To be clear here, even in yellow state, the Fairphone 4 doesn't really have "verified boot". It's more like the same level of security as the old "locked bootloader" thing way back. If I just get the access to the phone for a brief moment and I am only allowed to use adb, and assuming that the recovery or bootloader doesn't have any more serious issues than the test key, sure, I cannot tamper with it. But if I write something directly to the disk and sign it with the test key I can just compromise the device. Alternatively, if the recovery or OS has a vulnerability that I can exploit to write system images, I can just sign whatever I am writing with the test key and verified boot won't stop me from doing so.


mfwmyfacewhen (approved):

matchboxbananasynergy:
Agreed. These latest developments, along with major historical delays when it comes to updates (which are *paramount* to the security of the device), I do not think it is a sane choice for Privacy Guides at all anymore.

d4rklynk:
Agreed to. I don't see the point of using CalyxOS. Just stick to the stock OS for these smartphones if people want to have Google Play Services. Or they should switch to DivestOS.

d4rklynk:
Remove image of CalyxOS @tommytran732 btw.
And shouldn't we remove GrapheneOS vs CalyxOS too ?

matchboxbananasynergy:
> Remove image of CalyxOS @tommytran732 btw. And shouldn't we remove GrapheneOS vs CalyxOS too ?

People commonly compare the two, so if people don't see CalyxOS as a recommendation on the website, they'll start wondering why it's not there. I think that it makes sense to contrast the privacy and security features of GrapheneOS against CalyxOS (which again, is usually compared to GrapheneOS with no good reason) in order for people to have more context as to *why* we recommend GrapheneOS.

d4rklynk:
> > Remove image of CalyxOS @tommytran732 btw. And shouldn't we remove GrapheneOS vs CalyxOS too ?
> 
> People commonly compare the two, so if people don't see CalyxOS as a recommendation on the website, they'll start wondering why it's not there. I think that it makes sense to contrast the privacy and security features of GrapheneOS against CalyxOS (which again, is usually compared to GrapheneOS with no good reason) in order for people to have more context as to _why_ we recommend GrapheneOS.

Yeah ok I understand. Then yeah the title should be something like "Why is GrapheneOS over CalyxOS ?".

dngray:
To be listed, we currently require:

- Regular security updates
- Lockable bootloader

Having a look at the [Calyx feature page](https://calyxos.org/features/) and assuming one has a Pixel (there is no compelling reason to buy a OnePlus or Fairphone), I don't see a whole lot of reasons to choose this over GrapheneOS.

There's the Datura Firewall, but on GrapheneOS you've got the `INTERNET` permission you can revoke, which seems more robust than a stateful firewall. There is [microG](https://calyxos.org/docs/guide/microg/), but I think [sandboxed play services](https://grapheneos.org/usage#sandboxed-google-play) is a better option if you're afraid about Google seeing your unique identifiers (IMSI/IMEI etc).

I suppose there is the shifting trust from Google location services, to Mozilla and DejaVu. There is no reason you can't change that on GrapheneOS though if you wanted. As for the other features, about the only thing I see there is the panic button - [Ripple](https://guardianproject.info/apps/info.guardianproject.ripple/), and there's no reason you couldn't install that as you can with Signal and Whatsapp on GrapheneOS.

Are there any really stand out privacy features on Calyx that are not just bundled apps?

ph00lt0:
Fully agree with all above, but probably good to put this sum up of reasons why CalyxOS is inferior to GrapheneOS on the website. Still i see many people in the privacy world picking / recommending it over grapheneOS for no good reason and I think this can be educated.

dngray (approved):

dngray (changes requested):
Will approve for merging, once https://github.com/privacyguides/privacyguides.org/pull/1518#issuecomment-1178671593 is sorted.

dngray (approved):

PR #1519, opened 2022-07-09
Remove Bromite

matchboxbananasynergy:
Remove Bromite and recommend Brave for Android instead, as Bromite has a trend of being behind in updates.

https://github.com/privacyguides/privacyguides.org/discussions/1512


d4rklynk:
Where's your Brave recommendation ?

matchboxbananasynergy:
> Where's your Brave recommendation ?

added

dngray:
> > Where's your Brave recommendation ?

> added

I asked for this to be moved to a separate https://github.com/privacyguides/privacyguides.org/pull/1520

It does look like we will have to do this. A requirement we have for all web browsers is that they not fall out of sync with upstream in regard to security updates. It's one of the reasons we've not recommended many of the Firefox forks.

Unfortunately for Bromite, this isn't the first time https://github.com/privacyguides/privacyguides.org/discussions/59#discussioncomment-2099847. While we understand that a lot of work goes into a project like this we do have to put our reader's interests first.

dngray (approved):

flexagoon:
@Mohamad20ZX yes it does

PR #1520, opened 2022-07-09
Add Brave Android in Mobile browser recommendations

matchboxbananasynergy:
This PR adds Brave as a recommendation for Android


flexagoon:
Btw, what about brave rewards? Will enabling them compromise my privacy?

dngray (approved):

dngray:
> Btw, what about brave rewards? Will enabling them compromise my privacy?

No, but, you don't need to enable it either.

PR #1541, opened 2022-07-12
Shorten Notification Privacy

donefool:
Android and iOS use the exact same text for notification privacy (https://www.privacyguides.org/advanced/signal-configuration-hardening/#notification-privacy) so it can be shortened for conciseness and consistency.


d4rklynk:
They're not the same
One is "No name or Content"
The other is "No name or message"

donefool:
> They're not the same One is "No name or Content" The other is "No name or message"

Wow I'm blind. Sorry bout that

d4rklynk:
No problem mate

PR #1578, opened 2022-07-27
Add separate desktop and mobile browser pages

mfwmyfacewhen:
Resolves: #1228 

Changed the browser page into a mobile and desktop page for better organization/less confusion.

### DIFF
--- a/docs/desktop-browsers.en.md
+++ b/docs/desktop-browsers.en.md
@@ -1,10 +1,8 @@
 ---
-title: "Web Browsers"
+title: "Desktop Browsers"
 icon: octicons/browser-16
 ---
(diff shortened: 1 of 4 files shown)


docs/desktop-browsers.en.md
```diff
@@ -217,133 +211,6 @@ Under the *System* menu
 
 ## Mobile Recommendations
```
SkewedZeppelin, line 212: dangling header?

mfwmyfacewhen: Woops, good catch!

Lostname777:
Make a separate page for extensions too, that would be even better

dngray (approved):

PR #1588, opened 2022-07-30
Remove magic wormhole

mfwmyfacewhen:
Resolves: #1573 

Removed Magic Wormhole as it has not been updated in years.


dngray:
Approved. Glad to remove this, tbh I don't use any of these things to move files around my network, just plain old [`sftp`](https://en.wikipedia.org/wiki/SSH_File_Transfer_Protocol).

dngray (approved):

dngray (approved):

PR #1604, opened 2022-08-07
Added Passky

zigazajc007:
I have added Passky as recommendation for privacy focused password manager.

Link: https://passky.org
Source Code: https://github.com/Rabbit-Company

## Why is Passky one of the most privacy focused cloud-based password manager?

1. Passky's main website uses self-hosted Plausible Analytics tool, which does not link any data to the user unlike Google Analytics. So this analytics solution collect annonimouse data and is a lot more permisive. https://plausible.io/privacy-focused-web-analytics

2. Passky Clients (Website, Browser Extension, Desktop Application, Mobile application) does not contain any analytic tools.

3. When user opens Passky Client he can choose on what Passky Server data will be stored. If he chose his own Passky Server, it won't be connected in any way with us.

4. If he chose our Passky Server, the server will collect his IP (but IP will be hashed (SHA512) with a salt and stored for less then a day) as this is used to prevent brute force attacks on the server.

5. All submited data by the user (websites, passwords, messages) will be encrypted with AES on the client side and only encrypted data will be stored / send to the server. So server won't be able to decrypt submited data from users (Zero-knowledge architecture).

6. Email provided by the user for the registration also does not need to be valid, as email will only be used if you forgot your username and for 2FA to be reset. Note that if you don't provide your valid email address you won't be able to reset 2FA and get forgoten username, but your account will still function normally without a valid email address.

If users does not want to use their email for registration, but still wants to enter their valid email, they can create one for free using self-hosted Mailu instance which we host and does not contain any modifications: https://mail1.rabbit-hosting.com/admin/user/signup . In order to create free email account you don't need to enter any personal data.

7. Passky does not offer an option for TOTP tokens to be stored as we don't recommend passwords and 2FA to be located in the same basket.


mfwmyfacewhen:
This hasn't been approved. You should put all this in your discussion and wait until your discussion has the "approved" label to make a PR adding new software please.

zigazajc007:
https://github.com/privacyguides/privacyguides.org/discussions/230

It has been requested almost a year ago.

mfwmyfacewhen:
Yes but it wasn't approved to be added to the site.

zigazajc007:
What steps or information needs to be provided in order for the request to be inspected?

mfwmyfacewhen:
It'll get added if the team likes it and it offers something over the current recommendations.

PR #1660, opened 2022-08-19
Fix sentence App Store client -> on the App Store

ph00lt0:
App Store client makes it sound as it if is for the app store, but it is available in the app store instead.

### DIFF
--- a/docs/vpn.en.md
+++ b/docs/vpn.en.md
@@ -87,7 +87,7 @@ Find a no-logging VPN operator who isn’t out to sell or read your web traffic.
 
 ??? check "Mobile Clients"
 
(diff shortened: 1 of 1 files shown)


jonaharagon (approved):
Thanks (I wonder if they copy/pasted it from someplace that said "on iOS" lol)

jonaharagon:
For future reference, for suggesting changes to an open PR, making [line comments](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request) is preferable to creating another PR :)

PR #1662, opened 2022-08-20
Fix Neo Store listing

Guru-25:

### DIFF
--- a/docs/android.en.md
+++ b/docs/android.en.md
@@ -296,8 +296,6 @@ Evaluate whether the additional features in the F-Droid build are worth the slow
 
 #### Neo Store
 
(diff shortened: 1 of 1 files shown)


jonaharagon (approved):
@privacyguides/team two things we should look into: it looks like Droid-ify is now being developed again separately by the original maintainer, suggesting Neo Store might not be a mere rebrand, we should look into the differences; and we should make a decision on https://github.com/privacyguides/team/discussions/17.

dngray:
> it looks like Droid-ify is now being developed again separately by the original maintainer, suggesting Neo Store might not be a mere rebrand

Correct it's not. I was using both of them and found Neo Store was a little less annoying when updating applications,  in regard to having to okay them in the system tray.

dngray (approved):

Issue #1564, opened 2022-07-19
Remove product logos from secondary products


ph00lt0:
The card around the Mollie app is missing.

![mollie](https://user-images.githubusercontent.com/15004290/179796290-adad9780-81b7-4834-8537-8da03835c148.png)

d4rklynk:
That's completely normal.
We don't want to make a box for Molly as the main recommendation is still Signal.

ph00lt0:
in the way it is now it looks very off, i am sorry but it's very inconsistent. Besides that worth mentioning listings were removed and everything else was upgraded to a card. Now you are reintroducing it, I don't think this is a good thing. On all other pages the less preferred tools were also cards. Think about when CalyxOS was still there.

d4rklynk:
Yes but it still secondary. People should not chose between Signal and Molly, they should Signal, and depending of their threat model, their situation their preference they can take Molly.
In this page we recommend Signal, Element, Session and Briar. And optionally you can use Molly, we just say it exists.

d4rklynk:
We did the same for [vaultwarden](https://www.privacyguides.org/passwords/#bitwarden) btw

ph00lt0:
If it's not recommended, don't list it imho. I see now this change is new to me, it still looks like a mistake/broken.

dngray:
This was an intentional design as Molly is a "sub recommendation", its not that we don't recommend it, it's that it's an optional extra.

ph00lt0:
I understand you want to distinguish the other option, but I think the design is confusing and does not represent what you are trying to communicate.

ph00lt0:
My suggestion would be to also make it a card, but shirk f.x. the icon and make the text color less prominent. Currently the icon is even bigger than the one of Signal itself...

jonaharagon:
Looking at this again, I think the logo-type images should be removed from non-recommendations 👍 

The rest of the format looks good as-is without the logo, so they don't need to be converted to cards. I think we should remove the logos from Vaultwarden, Mollie, and NPxSB.

PR #1669, opened 2022-08-21
Adjust Vaultwarden and Molly Listings

jonaharagon:
Resolves: #1564 

- Removes logo from Vaultwarden listing
- Removes Vaultwarden and Molly from tools page
- Moves Molly information from signal hardening page to RTC recommendations page


jonaharagon:
I realized I didn't know what to do, because actually software configuration and recommendations shouldn't be in the _Knowledge Base_ section at all, all of that information would ideally be on the RTC page under _Recommendations_, similar to how we have configuration recommendations for browsers on the browser pages. But I also see why it was done this way, since there's a significant amount of information...

I made this PR more consistent so it should be good for now, and overall look better, but I want to think about how the Signal Configuration page is arranged and maybe rearrange things in the future...

jonaharagon:
I just don't want to have a recommendation like Molly on that page (or the tools page) without all the accompanying information that's on the hardening page (especially since it's Android only anyways).

mfwmyfacewhen (approved):

Beventar (approved):

PR #1681, opened 2022-08-23
Fix Bitlocker for Home edition guide

### LINKED ISSUE #1587: "powershell Get-Disk 0 | findstr GPT && echo This is a GPT system disk!" doesn't work on my windows 10 laptop
## Description

so executing "powershell Get-Disk 0 | findstr GPT && echo This is a GPT system disk!" on powershell on my laptop gives me an error, that the token "&&" isn't valid in this powershell version.

However, when executing "powershell Get-Disk 0" I get a result and I can read "GPT" under "Partition Style". 

> Number Friendly Name Serial Number                    HealthStatus         OperationalStatus      Total Size Partition Style
>                                                                                                              Style
> ------ ------------- -------------                    ------------         -----------------      ---------- ----------
> 0      WDC PC SN5... xxxxxx                                             Healthy              Online                  238.47 GB GPT

So maybe this should be the used method.

Edit:
I can't edit the title anymore, but the instructions are really unclear. For the second step:

> Check TPM version. The value returned must be "3 True". The spec must be 1.2 or above.

 What value are we talking about here? If it's all about the "SpecVersion", then we should mention that.
That's the result of the commands on my laptop:
```
PS C:\WINDOWS\system32> powershell Get-WmiObject -Namespace "root/cimv2/security/microsofttpm" -Class WIN32_tpm | findstr "IsActivated IsEnabled IsOwned SpecVersion"
IsActivated_InitialValue    : True
IsEnabled_InitialValue      : True
IsOwned_InitialValue        : True
SpecVersion                 : 2.0, 0, 1.16
```

**URL of affected page:**

https://www.privacyguides.org/encryption/#bitlocker

jonaharagon:
Fixes #1587

### DIFF
--- a/docs/encryption.en.md
+++ b/docs/encryption.en.md
@@ -102,34 +102,39 @@ BitLocker is [only supported](https://support.microsoft.com/en-us/windows/turn-o
 
     To enable BitLocker on "Home" editions of Windows, you must have partitions formatted with a [GUID Partition Table](https://en.wikipedia.org/wiki/GUID_Partition_Table) and have a dedicated TPM (v1.2, 2.0+) module.
 
-    1. Open Windows [PowerShell](https://en.wikipedia.org/wiki/PowerShell).
+    1. Open a command prompt and check your drive's partition table format with the following command. You should see "**GPT**" listed under "Partition Style":
 
-    2. Check to see partition table format:
         ```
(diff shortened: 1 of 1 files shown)


blacklight447 (approved):
Looks good to me, tested it in a vm and it works alright.

mfwmyfacewhen (approved):

PR #1688, opened 2022-08-23
Update GrapheneOS

d4rklynk:
- Added "notification forwarding"
- Added some links

### DIFF
--- a/docs/android.en.md
+++ b/docs/android.en.md
@@ -33,7 +33,9 @@ We recommend installing one of these custom Android operating systems on your de
     [:octicons-code-16:](https://grapheneos.org/source){ .card-link title="Source Code" }
     [:octicons-heart-16:](https://grapheneos.org/donate){ .card-link title=Contribute }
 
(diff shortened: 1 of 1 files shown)


d4rklynk:
> We mention notification forwarding in the "GrapheneOS vs CalyxOS" page. While it is a great usability improvement, I'm not sure that it's significant enough to be singled out as one of the features mentioned in the main page.

tbh, I thought the same afterward

d4rklynk:
But I know some people can really love that feature, so yeah, idk.

jonaharagon:
@matchboxbananasynergy I think these are my thoughts as well, I don't want to add a lot of extraneous information to our recommendations just for the sake of having a complete product description. I think as a rule we should only call out basic core functionality, notable privacy features, and privacy/security settings we want people to look at/change.

Plus, as you mentioned this is already noted in a knowledge base article.

Thanks @d4rklynk for the PR, but I think I'm just going to close this one, we can reopen it if someone strongly disagrees with me, but I think we should err on the side of keeping our recommendations simple :)

PR #1471, opened 2022-06-21
Remove Imagepipe and Scrambled Exif/Add ExifEraser

matchboxbananasynergy:
Missing ExifEraser SVG, Unsure how to do that part. Sorry!

Resolves: https://github.com/orgs/privacyguides/discussions/1458 and https://github.com/orgs/privacyguides/discussions/1457

PR #1702, opened 2022-08-28
Small fixes in Data Redaction Page

matchboxbananasynergy:
This PR has two minor fixes:

- Removes mention of Metapho being closed source. Open source is not a requirement, and we also have a source code button for open source recommendations, so readers know what software is open and what is closed. Mentioning it gives it a negative connotation which is not fair.
- Removes mention of Imagepipe under the PrivacyBlur recommendation that was recently moved from the Android page. Imagepipe is not available on the Play Store, while Pocket Paint is (as well as F-droid, if people want that) and it seems a lot more popular and polished (over a million downloads on Play Store). We also[ removed Imagepipe for requesting unnecessary permissions](https://github.com/privacyguides/privacyguides.org/pull/1471), so it doesn't make sense to recommend it elsewhere on the site.

### DIFF
--- a/docs/data-redaction.en.md
+++ b/docs/data-redaction.en.md
@@ -91,8 +91,6 @@ The app offers multiple ways to erase metadata from images. Namely:
 
     Metapho is a simple and clean viewer for photo metadata such as date, file name, size, camera model, shutter speed, and location.
 
-    Metapho is closed-source, however we recommend it due to the few choices there are for iOS.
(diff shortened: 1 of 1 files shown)


jonaharagon (approved):

mfwmyfacewhen (approved):

freddy-m (approved):
LGTM

PR #1703, opened 2022-08-30
Password manager corrections

jonaharagon:
Since #1666 wasn't reviewed by anyone after the latest changes to the PR 😜

### DIFF
--- a/docs/passwords.en.md
+++ b/docs/passwords.en.md
@@ -109,7 +109,7 @@ Bitwarden's server-side code is [open-source](https://github.com/bitwarden/serve
 
     ![1Password logo](assets/img/password-management/1password.svg){ align=right }
 
-    **1Password** is a password manager with a strong focus on security and ease-of-use, which allows you to store passwords, credit cards, software licenses, and any other sensitive information in a secure digital vault. Your vault is hosted on 1Password's servers for a monthly fee. 1Password is [audited](https://support.1password.com/security-assessments/) on a regular basis, provides exceptional customer support. 1Password is closed source however, the security of the product is thoroughly documented in their [security white paper](https://1passwordstatic.com/files/security/1password-white-paper.pdf).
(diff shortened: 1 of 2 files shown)


docs/passwords.en.md
```diff
@@ -109,7 +109,7 @@ Bitwarden's server-side code is [open-source](https://github.com/bitwarden/serve
...
+    **1Password** is a password manager with a strong focus on security and ease-of-use, which allows you to store passwords, credit cards, software licenses, and any other sensitive information in a secure digital vault. Your vault is hosted on 1Password's servers for a monthly fee. 1Password is [audited](https://support.1password.com/security-assessments/) on a regular basis and provides exceptional customer support. 1Password is closed source, however the security of the product is thoroughly documented in their [security white paper](https://1passwordstatic.com/files/security/1password-white-paper.pdf).
```
jonaharagon, line 112: Given the internal debate I'll leave it up to @matchboxbananasynergy and @dngray to decide whether to remove this mention or not, lol:

```suggestion
    **1Password** is a password manager with a strong focus on security and ease-of-use, which allows you to store passwords, credit cards, software licenses, and any other sensitive information in a secure digital vault. Your vault is hosted on 1Password's servers for a monthly fee. 1Password is [audited](https://support.1password.com/security-assessments/) on a regular basis, provides exceptional customer support, and the security of the product is thoroughly documented in their [security white paper](https://1passwordstatic.com/files/security/1password-white-paper.pdf).
```

dngray: As we were **directly** comparing it to Bitwarden, it was a prudent thing to mention. It is also directly related to the existence of the whitepaper. The evaluation the user would make is if they trust the source, then the whitepaper details how the encryption works.

I think generally something is open or closed source, might not be particularly important if the software doesn't deal with particularly sensitive data. However, a password manager is about as sensitive as you can get so I think it's okay to state this *fact* and without connecting a negative connotation to it.

PR #1704, opened 2022-08-30
Blog replacement

jonaharagon:
Discussion: https://github.com/privacyguides/team/discussions/25

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,6 +12,7 @@ typing-extensions = "*"
 mkdocs-minify-plugin = "*"
 mkdocs-rss-plugin = "*"
 mkdocs-git-committers-plugin-2 = "*"
+mkdocs-macros-plugin = "*"
 
 [dev-packages]
 scour = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "d3a92aec93d430ec2770963ae2dc2f72ad9f0dd042170dc90ab7a0c07b6dcda6"
+            "sha256": "bc99c3ae97af3b039c20ec5ed0cb345e76d171f08cb8c1e2aface7fad4bab695"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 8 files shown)


dngray:
I guess this will break RSS though, which means we'll get requests about that?

freddy-m:
Are we going to hook this up to RSS or will it be done manually? Really like the way it looks just might not be able to constantly update it, thus leading to the problems of stagnation we mentioned earlier.

We could also add a blogroll underneath the recommended articles, linking to blogs we enjoy...

jonaharagon:
It'd be done manually, all you have to do is add the article to [`articles.yml`](https://github.com/privacyguides/privacyguides.org/blob/bfd3f22d318e3f283e07a54c19caaeca7b5b573f/data/articles.yml) so maintaining it is trivial, you can add articles as you tweet them, or I or anyone else can add articles we like easily, doesn't really matter.

I definitely want to manually curate articles, but we could _also_ use something like openring at the bottom of the page like you suggested, I'm just not sure what sources we'd like to add. I think it's probably for the best if we are manually reviewing everything we link to on the website.

freddy-m:
> I think it's probably for the best if we are manually reviewing everything we link to on the website.

You make a good point, potential for some damage if we use an RSS feed. We just need to ensure it recieves at least one new article per week. Perhaps we could say: email suggestions to [idk who's email, probably mine].

freddy-m (approved):
LGTM

mfwmyfacewhen (approved):

blacklight447 (approved):

dngray (approved):

Victor239:
> You make a good point, potential for some damage if we use an RSS feed. 

I'm not sure what damage there is by using a RSS feed, but it's a real shame that one isn't present currently because that was how I was following this project. I'm not realistically going to check this single website when I get dozens of blogs, YouTube channels, Hacker News feeds and everything else via RSS. Reminds me of proprietary silos like Facebook and Instagram which refuse to allow open-source apps protocols to receive their feeds.

jonaharagon:
@Victor239 that's not what we're referring to there, what we dismissed was pulling articles from RSS feeds we choose to automatically post on https://www.privacyguides.org/articles/. There was some background discussion for this PR that wasn't talked about on this page, but TL;DR we have this "Articles from around the web" page to post links to articles on other sites in addition to our own blog, but what we decided here was that those articles we link to will be manually curated. We should still have an RSS feed hosted by ourselves for this section... I think that was overlooked so it might be broken. I'll open an issue to re-create our RSS feed here now so I don't forget.

For what it's worth, you can follow the changelog of this entire website via RSS here: https://github.com/privacyguides/privacyguides.org/releases.atom

Victor239:
Ah my mistake then, I thought it was removed since my browser extension doesn't detect any RSS feed on https://www.privacyguides.org/articles/ nor https://www.privacyguides.org/blog/2022/06/09/hide-nothing/, but the old URL at https://blog.privacyguides.org/feed/ still works and I'm subscribed to. Glad to hear the normal blog RSS isn't going away then :)

jonaharagon:
Well blog.privacyguides.org is actually going away soon, but I'll make sure we add a feed to www.privacyguides.org before that happens.

freddy-m:
> Well blog.privacyguides.org is actually going away soon, but I'll make sure we add a feed to [www.privacyguides.org](http://www.privacyguides.org/) before that happens.

On that note, where are the old articles published on the site being stored? I'll migrate mine to my blog when it's finally public, but what about Dan's?

jonaharagon:
@freddy-m the ones currently on the site will stay as-is.

PR #1705, opened 2022-08-30
Grammar fixes on the 1Password card

mfwmyfacewhen:


jonaharagon (commented):
I'm not sure this PR makes sense, it adds more text to the inside of the card which we want to avoid, and it removes information relevant to why the recommendation was made.

mfwmyfacewhen:
I'm open for suggestions, matchbox pointed out that the text was above the card which isn't consistent with the rest of the site so I moved it.

blacklight447 (approved):

PR #1722, opened 2022-09-06
Re-order Password Managers

matchboxbananasynergy:
Re-order password managers to follow the common theme across the website which is:

Cloud options first, local options second. Also, put the Android option below the desktop/cross-platform one.

Also added the (Android) part to KeePassDX as we usually do this for Android and iOS apps across the website.


blacklight447 (approved):
Everything seems to be in order

mfwmyfacewhen (approved):

PR #1743, opened 2022-09-09
Add Strongbox to Password Managers

mfwmyfacewhen:
Adds Strongbox to the Password Manager recommendations #1487


freddy-m (approved):
Last update was 2 months ago, but thats not too long ago. Otherwise LGTM

PR #1771, opened 2022-09-16
Small VPN Criteria cleanup

matchboxbananasynergy:
This PR:

- Removes mention of F-droid from VPN criteria. If our criteria is that VPN clients should be open-source, that's completely fine. Mentioning F-droid by name, however, when we currently issue a lot of warnings and caveats about it felt off.
- Renames "Hidden Service" to "Onion Service", as that is the correct terminology now.


freddy-m (approved):
LGTM

dngray (approved):

PR #1774, opened 2022-09-16
Proton VPN - corrected the date in the annotation

Redre1l:
Corrected the date in the annotation of all Proton VPN countries.


Redre1l:
> We just had a discussion about this and would like to keep the dates. Can you change "as of" to "Last checked:" and change the date format to YYYY-MM-DD?
> 
> To make it clear, that date is not when a provider added or removed a country, but when we last had a look.
> 
> As you seem to have checked and updated all providers, I'd put today's date on all of them.

Ok, it seems to be better and more clear.

jonaharagon (approved):

PR #1788, opened 2022-09-18
Rename Video Streaming Page to Media Streaming

matchboxbananasynergy:
This PR renames the video streaming page to media streaming in order to make it more flexible. It means that we can consider options that are only for music, for example. Restricting the scope of the page to videos only felt arbitrary.


freddy-m (approved):
LGTM but double check given what happened last time lol

PR #1792, opened 2022-09-19
Add Frontends Page

matchboxbananasynergy:
This PR moves FreeTube, NewPipe, Invidious, Piped, and Librarian to a new "Frontends" page so that we can expand it with non-video streaming frontends in the future.

Discussion on the topic: https://github.com/privacyguides/privacyguides.org/discussions/1791


mfwmyfacewhen (approved):

jonaharagon (approved):
from matrix:
> with the frontends PR I want to have a paragraph or two at the top of the page explaining what they actually are and why/when they're useful
> would you rather add that before merging or merge this and add that in a future PR?

PR #1805, opened 2022-09-23
Use stylization of MobileCoin

taivlam:
Use stylization of MobileCoin, as used on [Wikipedia](https://en.wikipedia.org/wiki/MobileCoin)


freddy-m (approved):
LGTM

taivlam:
@matchboxbananasynergy This is ready for review.

Issue #1737, opened 2022-09-09
Standardize listed pricing info


matchboxbananasynergy:
### Discussed in https://github.com/privacyguides/privacyguides.org/discussions/1735

<div type='discussions-op-text'>

<sup>Originally posted by **Redre1l** September  9, 2022</sup>
In the email section, providers have listed prices while other paid/free services in different sections don't have that 
(i.e. Cloud Storage, DNS Resolvers or other sections **not** in the _Providers_ section like Password Managers).

Also if a service offers a free plan alongside paid plans it doesn't make sense to only list the _Free_ label.
(i.e. Proton Mail and Tutanota in the email section)

There is some further info in the text of the card but if you're giving labels under the text of the card, make it consistent.</div>

mfmw123:
I'd do it like this:
- list pricing info everywhere where applicable
- change the format to monthly prices
- include free and premium plans

or all prices can be removed and just redirect the user to the website for pricing info.

PR #1806, opened 2022-09-24
Email alias pricing cleanup

mfmw123:


dngray (approved):

freddy-m:
We don't do this for all of our email providers, let alone every recommendation on the site. Should we @privacyguides/team? Just seems odd to do this for a very small section when the rest of the site is different...

mfmw123:
There is an issue about it #1737

mfwmyfacewhen:
The problem I see with this is that prices change and if we put the price for every service, we'll have to constantly watch it and update it. IMO a link to the site is enough.

PR #1820, opened 2022-09-28
Relicensing: CC BY-ND 4.0

jonaharagon:
From the PR, which also explains this change:

> Unless otherwise noted, the original content on this website is made available under the [Creative Commons Attribution-NoDerivatives 4.0 International Public License](https://github.com/privacyguides/privacyguides.org/blob/43bed11f1c59e23e7cc9bf099cd0d190097408fe/LICENSE). This means that you are free to copy and redistribute the material in any medium or format for any purpose, even commercially; as long as you give appropriate credit to `Privacy Guides (www.privacyguides.org)` and provide a link to the license. You may do so in any reasonable manner, but not in any way that suggests Privacy Guides endorses you or your use. If you remix, transform, or build upon the content of this website, you may not distribute the modified material.
> 
> This license is in place to prevent people from sharing our work without giving proper credit, and to prevent people from modifying our work in a way that could be used to mislead people. If you find the terms of this license too restrictive for the project you're working on, please reach out to us at `jonah@privacyguides.org`. We are happy to provide alternative licensing options for well-intentioned projects in the privacy space!

As far as the "NoDerivs" thing: Basically, if someone emails me and wants to create a modified/derivative work for a legitimate reason then we can grant them a less restrictive license than this just for their project, but we want to do so on a case by case basis.


tommytran732:
Yeah this would probably make contributing to PG easier

mfwmyfacewhen (approved):

blacklight447 (approved):

freddy-m (approved):

dngray (approved):

PR #1829, opened 2022-10-03
Mention Neo Store in DivestOS recommendation

JustLuckNoSkill:
- [x] Please check this box to confirm your agreement to publish your work under the [Creative Commons Attribution-NoDerivatives 4.0 International](https://github.com/privacyguides/privacyguides.org/blob/main/LICENSE) license, and to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform and distribute your contribution as part of our project.

The current DivestOS recommendation lists F-Droid as a positive, which contradicts the F-Droid section at the bottom of the Android page. This PR adds a warning against using F-Droid despite it being installed by default. Alternatively, the F-Droid mention could be deleted entirely.


SkewedZeppelin:
@matchboxbananasynergy 
As long as users add the repos I don't care what client they use.

PrivExt was removed in April, so updates aren't attended.

SkewedZeppelin:
The updated version still isn't correct and woulds result in users missing critical WebView updates.

The available (Sept 10th) builds of DivestOS are currently shipping Chromium 105.0.5195.79.
But I've pushed out 105.0.5195.124, 105.0.5195.136, 106.0.5249.65, and 106.0.5249.79 via the WebView repo since then: https://divestos.org/misc/ch-dates.txt

mfwmyfacewhen (approved):

SkewedZeppelin:
> What do you think about the wording now?

Wording could be a bit stronger, but looks fine.

> I am correct in assuming that the F-Droid app can be installed but not uninstalled, right?

Default installed, can be disabled.

>  two separate repos 

When I switched to the Mulch webview I wanted to mtaintain compatibility with all previous builds, this mean signing the WebView for each device. So there are actually 120+ copies of the WebView apk in the repo.
Furthermore fdroidserver is very slow and buggy at processing Chromium apks, so it is actually a patched version to allow it to even generate and separate to prevent taking 5 minutes to regenerate for updates of the other apps.

It also has the benefit that if for some reason (like the [time](https://divestos.org/index.php?page=news#2022-03.2) they broke scrolling for some apps) there is a bad webview update, I can just move the old repo back into place.
For example you can see the previous version still there: https://divestos.org/fdroid/webview-106.0.5249.65-1/index-v1.json

docs/android.en.md
```diff
@@ -61,12 +61,13 @@ Google Pixel phones are the only devices that currently meet GrapheneOS's [hardw
...
 
+DivestOS uses F-Droid as its default app store. Normally, we would recommend avoiding F-Droid due to its numerous security issues. However, doing so on DivestOS isn't viable; the developers update their apps via their own F-Droid repositories ([DivestOS Official](https://divestos.org/fdroid/official/?fingerprint=E4BE8D6ABFA4D9D4FEEF03CDDA7FF62A73FD64B75566F6DD4E5E577550BE8467) and [DivestOS WebView](https://divestos.org/fdroid/webview/?fingerprint=FB426DA1750A53D7724C8A582B4D34174E64A84B38940E5D5A802E1DFF9A40D2)). We recommend disabling the official F-Droid app and using [Neo Store](https://github.com/NeoApplications/Neo-Store/) with the DivestOS repositories enabled to keep those components up to date. For other apps, our recommended methods of obtaining them still apply.
```
dngray, line 69: > Normally, we would recommend avoiding F-Droid due to its numerous security issues

Feeling like that needs a reference, rather than being a floating statement.

dngray (changes requested):
Think we should provide a link somewhere there, to what these "security issues" are. This will be a contentious topic amongst free software enthusiasts.

dngray:
> Not sure if we should link to the wonderfall article in the DivestOS section when we mention the security issues with F-Droid, or if we should just have a link to our F-Droid section below, especially as there are plans to expand it with more context/info. Thoughts @dngray?

I'm thinking expand the section [on the android page](https://www.privacyguides.org/android/#f-droid) to be a brief synopsis of some of the points in the wonderfall article. We should elaborate on that sentence:

> However, there are [notable problems](https://wonderfall.dev/fdroid-issues/) with the official F-Droid client, their quality control, and how they build, sign, and deliver packages.

I think I would like to see this sentence elaborated into a single paragraph. The wonderfall article is quite lengthy, and I think its points won't be read by readers. I'm not saying we need to mention all of the points there, but I do think 1-3 are most relevant. **Especially 3**.

dngray (approved):

PR #1845, opened 2022-10-17
Add more detail to recommendation overview page

jonaharagon:
Screenshot 2022-10-17 at 11 21 32@2x

- [x] Please check this box to confirm your agreement to publish your work under the [Creative Commons Attribution-NoDerivatives 4.0 International](https://github.com/privacyguides/privacyguides.org/blob/main/LICENSE) license, and to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform and distribute your contribution as part of our project.

### DIFF
--- a/docs/android.en.md
+++ b/docs/android.en.md
@@ -26,10 +26,7 @@ We recommend installing one of these custom Android operating systems on your de
 
 ### GrapheneOS
 
-!!! recommendation
-
-    ![GrapheneOS logo](assets/img/android/grapheneos.svg#only-light){ align=right }
-    ![GrapheneOS logo](assets/img/android/grapheneos-dark.svg#only-dark){ align=right }
+!!! recommendation "![GrapheneOS logo](assets/img/android/grapheneos.svg#only-light){ align=right }![GrapheneOS logo](assets/img/android/grapheneos-dark.svg#only-dark){ align=right }"
 
     **GrapheneOS** is the best choice when it comes to privacy and security.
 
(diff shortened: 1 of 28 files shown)


blacklight447:
Now we are talking about changing this page, is the tor network really the first entry we want on the tools page, wouldn't something like browsers be more appropriate?

freddy-m (commented):
Generally LGTM though as @blacklight447 points out we need to think about the order of these now...

KaiTebay:
Especially since the list is so long, I personally find the amount of text in this layout to be a little overwhelming—what about keeping the new layout but stripping-back the extra information so that each one has a little introduction or tag-line? I think that keeping things smaller helps to break-up the huge page.

So, instead of

> ### **Tor Browser**: Best For Browsing The Web Anonymously
> Tor Browser is your choice if you need anonymity. It's a fork of Firefox designed for anonymous browsing for desktop computers and Android, and it includes default settings and extensions that are automatically configured based on your set security level.

It could look like:

(Tag-line: more descriptive than existing cards but still small)

> ### **Tor Browser**
> Best for browsing the web anonymously.

(Tl;dr: much more descriptive but only 1-2 short sentences)

> ### **Tor Browser**
> Tor Browser is the best choice if you need anonymity. It's a fork of Firefox, available on desktop computers and Android, with default settings to support different security levels.

Generally, though, I think that making the cards descriptive is a great idea, since you otherwise have to open the page if you don't know what they are.

KaiTebay:
In terms of ordering, I think the alphabetized sections (Operating Systems onwards) already make a lot of sense. I think that this makes more sense for Internet Browsing (and it happens to be alphabetized, too). :heart_eyes: 

1. Desktop Browsers
2. Mobile Browsers
3. Tor Network

But mainly because the analytics indicate the desktop browsers page to be one of the most popular. :smile:

jonaharagon:
> But mainly because the analytics indicate the desktop browsers page to be one of the most popular.

I think this is true, but I also think the intent behind the current order is to suggest using Tor first and foremost, before delving into less private and more traditional browser options.

freddy-m:
I quite like @KaiTebay's idea of 'much more descriptive but only 1-2 short sentences)' cards. It will also be a neat way of evaluating all of our recommendations; if we can't sum up why they are on the site in a succint way why recommend them? We also need to make it clear that there is a 'Learn more -->' section, which longer cards _could_ hide?

mfwmyfacewhen:
We'll probably need to strip back our recommendations a bit or this list will be way too long.

jonaharagon:
I forgot what the point of this PR was supposed to be.

Issue #1514, opened 2022-07-06
Check LUKS2 Encrypted boot


dngray:
## Description

**URL of affected page: https://privacyguides.org/linux-desktop/hardening/#secure-boot**

Looks like [GRUB supports LUKS2](https://www.phoronix.com/scan.php?page=news_item&px=GRUB-Boots-LUKS2-Disk-Encrypt). This [comment](https://libredd.it/r/PrivacyGuides/comments/timghp/new_linux_desktop_page/i1h30je/?context=3) seems to suggest `systemd-boot` can also boot an encrypted LUKS2 partition, but I've not seen any information about that.

We should investigate this and update the page.

d4rklynk:
I found this https://www.gnu.org/software/grub/manual/grub/grub.pdf at chapter `16.3.19 cryptomount`

> GRUB suports devices encrypted using LUKS, LUKS2 and geli. Note that necessary
modules (luks, luks2 and geli) have to be loaded manually before this command can be used. For LUKS2 only the PBKDF2 key derivation function is supported, as
Argon2 is not yet supported.

dngray:
At this point we wouldn't be recommending a git tagged version of Grub. If I'm being honest, I really don't like the "encrypted boot" idea because it will never be part of an installer, and is certainly  redundant when you're using TPM  [`systemd-measure`](https://www.freedesktop.org/software/systemd/man/systemd-measure.html) which will no doubt become a better developed solution.

When I tried out encrypted encrypted Grub it was **super slow**, because the Grub implementation of even LUKS1 didn't have SSE support if I remember correctly.  I doubt it has been audited either, as nobody really uses except for hardcore nerds following the arch wiki.

I also totally see distributions in the future doing away with BIOS support, and only supporting [`systemd-boot`](https://en.wikipedia.org/wiki/Systemd-boot) only. Redhat already has made some attempt to do this [in F37](https://fedoraproject.org/wiki/Changes/DeprecateLegacyBIOS) and although rejected they have settled for making [UEFI the default](https://fedoraproject.org/wiki/Releases/37/ChangeSet).

PR #1850, opened 2022-10-22
Remove mention of encrypted boot

dngray:
Closes: https://github.com/privacyguides/privacyguides.org/issues/1514

I'm not a fan of encrypted boot, or suggesting it for a number of reasons.

1. It's slow, as decryption happens without acceleration features like AES-NI, SSE, etc
2. Only LUKS1 is supported unless you use a patch/development branch, that means compiling it yourself, this is not likely to be compatible with distribution upgrades
3. It doesn't relate to privacy, private data is not stored in `/boot`
4. It does not provide authentication, ie any form of tamper protection ie if the content of `/boot` is in any way predictable, an attacker can modify the volume to contain whatever they want
5. Better off would be to write a TPM guide that uses [`systemd-cryptenroll`](https://www.freedesktop.org/software/systemd/man/systemd-cryptenroll.html), and later [`systemd-measure`](https://www.freedesktop.org/software/systemd/man/systemd-cryptenroll.html). Use TPM to make measurements a requirement now for Windows 11. For more info about that see https://github.com/privacyguides/privacyguides.org/issues/1855

See "Summary of Resources and their Protections" of https://0pointer.net/blog/authenticated-boot-and-disk-encryption-on-linux.html

### DIFF
--- a/docs/linux-desktop/hardening.en.md
+++ b/docs/linux-desktop/hardening.en.md
@@ -116,14 +116,6 @@ One of the problems with Secure Boot, particularly on Linux is, that only the [c
 
 To eliminate the need to trust Microsoft's keys, follow the "Using your own keys" section on the [Arch Wiki](https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface/Secure_Boot). The important thing that needs to be done here is to replace the OEM's key with your own Platform Key.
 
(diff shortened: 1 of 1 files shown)


mfwmyfacewhen (approved):

blacklight447 (approved):
Loosk good

PR #1905, opened 2022-12-10
Fix search suggestions

mfwmyfacewhen:
Firefox Suggest doesn't send data but [search suggestions](https://support.mozilla.org/en-US/kb/search-suggestions-firefox) does. I went ahead and fixed this.


dngray (approved):

dngray:
Also seems like [Firefox Suggest](https://support.mozilla.org/en-US/kb/firefox-suggest-faq) is off by default, and only available to people in the US.

PR #1933, opened 2023-01-09
Criteria updates to email: Require export and headers

dngray:
We don't stipulate that providers must support IMAP connectivity, because in some cases that is not a requirement for users, we do believe **all** providers must have export functionality in some way or another so that users can choose to leave at a later date.

We also believe all providers must have a feature to see "advanced information" about the emails such as their headers.

### DIFF
--- a/docs/email.en.md
+++ b/docs/email.en.md
@@ -209,7 +209,7 @@ For everything else, we recommend a variety of email providers based on sustaina
         - [:simple-linux: Linux](https://tutanota.com/#download)
         - [:octicons-browser-16: Web](https://mail.tutanota.com/)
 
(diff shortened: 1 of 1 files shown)


freddy-m (approved):
LGTM

dngray:
Hmm, thinking I should probably mention that Tutanota at this time doesn't really allow for a bulk export. https://github.com/tutao/tutanota/issues/1292. While they have made an attempt there with this prototype feature https://github.com/tutao/tutanota/commit/3dfd7aba328473471e7a1d74c93de2033fa3e90a

>  Prototype export mails feature
>
> This is per folder, and currently only available on custom folder, but
this is only a UI limitation. Works even for folders with 1000+ emails.
It does get some 429 (rate limits?), but in the end it works. It does
take quite a bit of time, though. A clear message of "downloading" and
maybe a progress bar would be nice.

It seems there are still some limitations.

PR #1977, opened 2023-02-14
Add mention of Libredirect

dngray:


dngray:
> Is the functionality that this provides worth recommending another extension?

If you want your browser to redirect to these things, then this is probably the most maintained, customizable thing out there. As far as fingerprinting goes, no concern because it only redirects to the instances you want.

mfwmyfacewhen (dismissed):

blacklight447:
Doesn this mean you have to trust the maker of the extension to not redirect you in cases you don't want? 

I mean it could redirect you to a fake bank site even if you fill in the right url right?

mfwmyfacewhen:
> Doesn this mean you have to trust the maker of the extension to not redirect you in cases you don't want?
> 
> I mean it could redirect you to a fake bank site even if you fill in the right url right?

Pretty much yeah.

dngray:
> Security concerns

I'm wondering if we should reformat that page to have one warning at the top, because it applies to all frontends

> When self-hosting, it is important that you have other people using your instance as well in order for you to blend in. You should be careful with where and how you are hosting ProxiTok, as other peoples' usage will be linked to your hosting.

For example is repeated throughout as is:

> When you are using a ProxiTok instance, make sure to read the privacy policy of that specific instance. ProxiTok instances can be modified by their owners and therefore may not reflect their associated privacy policy. Some instances have Tor .onion addresses which may grant some privacy as long as your search queries don't contain PII.

blacklight447:
I am personally not really convinced that the added convience is worth the security risk here, can't folks better not just use bookmarks?

dngray:
After some further thinking about this, I think the main issue is this would amount to us suggesting particular instances, which of course we are not vetting.

dngray:
> Doesn this mean you have to trust the maker of the extension to not redirect you in cases you don't want?

I do think the risk, is minimal, and it would be very obvious if they did.

There are security benefits from frontends when you're not using JavaScript or you are using .onion services, which Libredirect can optionally select.

The ability to do something harmful without JavaScript is pretty limited.

dngray:
> Also, is the default for it to redirect to a random instance? If so, that introduces a whole heck of a lot more trusted party. What if an instance is compromised, and people are redirected to it?

That is the thing I'm concerned about the most.

> Suggesting this feels like a regression. People should be making educated choices about which instances they choose to trust. And while I'm sure that Libredirect allows you to pin specific instances, that should be the default (it might be, not sure), and even then, the potential issues that this can cause just don't make the juice worth the squeeze imo.

Which is a fair point, it's not the default.

jonaharagon:
I was going to reopen this because I think you are wrong @matchboxbananasynergy, the default is to not redirect anything at all, so enabling a redirect requires manual config intervention, **and** if you enable a redirect it defaults to a single pre-set server and not a random selection out of all the available options. (And switching instances is on the same page you're already on to enable the redirect anyways, so I don't have any UX concerns there with people accidentally choosing an instance they're not familiar with).

However, I then saw it requires manual installation in developer mode on Chromium browsers, which I'm not comfortable with recommending :)

**If** they move to Manifest v3 and publish to the Chrome Web Store in the future though, I do want to reopen this discussion at that time.

dngray:
> I was going to reopen this because I think you are wrong @matchboxbananasynergy, the default is to not redirect anything at all, so enabling a redirect requires manual config intervention, **and** if you enable a redirect it defaults to a single pre-set server and not a random selection out of all the available options

Oh, that's not so bad. I couldn't remember if I had set that or not.

> If they move to Manifest v3 and publish to the Chrome Web Store in the future though, I do want to reopen this discussion at that time.

We'll revisit it then. I've restored the branch, so we can look at it then.

PR #1978, opened 2023-02-14
Add Vanadium guide

dngray:
Going to add this, as we don't want GrapheneOS users to think they should be using Brave instead.


jonaharagon:
> Vanadium **only on GrapheneOS** can be updated [...]

Vanadium is exclusive to GrapheneOS anyways, no?

Also does anyone know if Vanadium supports password managers yet? I think it does not judging by issue 276 on their repo (which I won't link to here for _obvious reasons_).

dngray:
It does seem there was some interest in an adblocker https://github.com/GrapheneOS/Vanadium/issues/10 but not sure if it's stalled.

jonaharagon:
Where are we at with this PR? To be honest, after thinking about this I don't think the benefits of Vanadium over our current recommendation has really been demonstrated at all. I'm tempted to mark this as missing discussion and push this back to the forum- thoughts?

jonaharagon:
In fact, the previous discussions about Vanadium I can find ended with rejection. I definitely would rather see a forum thread opened with *actual specific reasons and evidence* that this PR needs to exist at all. 

https://github.com/privacyguides/privacyguides.org/discussions/527#discussioncomment-1893756

mfwmyfacewhen:
> > [bitwarden/mobile#2199 (comment)](https://github.com/bitwarden/mobile/pull/2199#issuecomment-1426435642)
> > @ph00lt0 The comment above suggests that you're able to get it working with the new app ID.
> 
> Right seems to be resolved I wasn't aware. Yet adblocking still not existing.

I think the proper way to do it is to disable JS on sites by default and whitelist sites as needed. This will remove most ads/ad tracking. The few sites with really bad ads that you also need JS enabled can usually be circumvented using an alternate frontend like with YouTube. I only see adblockers as useful in browsers where you can't whitelist JS per-site.

kimg45:
Vanadium has built-in content blocking now: https://github.com/GrapheneOS/Vanadium/issues/10

redoomed1:
Are you accepting suggestions/drafts for this PR? I have some time on my hands, so I would be happy to help write this guide.

As @kimg45 noted, Vanadium has had built-in content blocking since mid-February 2024, so it currently meets the [minimum requirements for recommendation](https://www.privacyguides.org/en/mobile-browsers/#minimum-requirements). Privacy Guides also recommends GrapheneOS as a recommended Android OS, so it seems sensible to recommend the browser that comes with it.

dngray:
> Are you accepting suggestions/contributions for this PR? If so, I'm happy to help draft/write this guide as I have some time on my hands.

I think we should otherwise it sort of says that Vanadium isn't good enough you're on GrapheneOS which simply isn't true.

Also Vanadium now has a content blocker so there is that.

docs/mobile-browsers.md
```diff
@@ -128,6 +152,26 @@ Brave allows you to select additional content filters within the internal `brave
...
+
+Vanadium is built upon the Chromium web browser project, so it should feel familiar and have minimal website compatibility issues.
+
+<details class="downloads" markdown>
+<summary>Downloads</summary>
+
+- [:octicons-home-16: Homepage](https://vanadium.app){ .md-button .md-button--primary }
+- [:octicons-code-16:](https://github.com/GrapheneOS/Vanadium){ .card-link title="Source Code" }
+
+</details>
```
redoomed1, line 171:
```suggestion
- [:octicons-home-16: Homepage](https://vanadium.app){ .md-button .md-button--primary }
- [:octicons-code-16:](https://github.com/GrapheneOS/Vanadium){ .card-link title="Source Code" }
```

docs/mobile-browsers.md
```diff
@@ -167,6 +211,12 @@ We would suggest installing [uBlock Origin](browser-extensions.md#ublock-origin)
 
 Mull comes with privacy protecting settings configured by default. You might consider configuring the **Delete browsing data on quit** options in Mull's settings if you want to close all your open tabs when quitting the app automatically, or clear other data such as browsing history and cookies automatically.
 
+#### Adblocking
+
+Currently the only way to block ads with Vanadium is to set a [DNS Resolver](../dns.md) that has this functionality. The option for that can be found in :material-dots-vertical: → **Settings** → **Privacy and security**. If you're using a VPN we suggest using your VPN provider's server.
+
+- [x] Select **Use secure DNS** > **Choose another provider**
+
```
redoomed1, line 219:
```suggestion
```

redoomed1 (changes requested):

docs/mobile-browsers.md
redoomed1: I suggested basic formatting changes and other fixes first. I'll suggest description changes and recommended configuration later.

docs/mobile-browsers.md
redoomed1: Vanadium is installed by default on GrapheneOS, so a "Downloads" toggle isn't needed here.

jonaharagon: @all-contributors add @redoomed1 for review

jonaharagon:
@all-contributors add @redoomed1 for review

redoomed1 (commented):
I think that Vanadium should be placed above Brave **not** because this implies an order, but because the Mobile Browsers page would flow better.

Here's my vision of how this page reads:

## Android
For people using [GrapheneOS] (android.md#grapheneos) on a [Google Pixel] (android.md#google-pixel), we recommend using Vanadium as it makes use of the security hardening present in GrapheneOS, such as [hardware memory tagging] (https://security.googleblog.com/2019/08/adopting-arm-memory-tagging-extension.html) on the Pixel 8 series.

### Vanadium
...

Here are the web browsers we recommend for other Android operating systems.

### Brave
...

### Mull
...

docs/mobile-browsers.md
```diff
@@ -128,6 +152,26 @@ Brave allows you to select additional content filters within the internal `brave
...
 
+### Vanadium (GrapheneOS)
+
+<div class="admonition recommendation" markdown>
+
+![Vanadium logo](assets/img/browsers/vanadium.svg#only-light) ![Vanadium logo](assets/img/browsers/vanadium-dark.svg#only-dark){ align=right }
+
+**Vanadium** is the default browser that is included with [GrapheneOS](../android.md#grapheneos), it includes a number of [patches](https://github.com/GrapheneOS/Vanadium/tree/main/patches) which assist in improving security and privacy of WebView.
+
+Vanadium is built upon the Chromium web browser project, so it should feel familiar and have minimal website compatibility issues.
```
redoomed1, line 163:
```suggestion
**Vanadium** is the default browser that is included with [GrapheneOS](../android.md#grapheneos). The Chromium-based browser includes a number of [patches](https://github.com/GrapheneOS/Vanadium/tree/main/patches) which help improve the security and privacy of the WebView used by other apps.

Among the numerous privacy and security enhancements listed in their documentation, Vanadium notably includes a built-in content blocker and disables JavaScript Just-In-Time (JIT) compilation by default, both of which can be toggled on a per-site basis.
```

docs/mobile-browsers.md
```diff
@@ -128,6 +152,26 @@ Brave allows you to select additional content filters within the internal `brave
...
+
+<details class="downloads" markdown>
+<summary>Downloads</summary>
+
+- [:octicons-home-16: Homepage](https://vanadium.app){ .md-button .md-button--primary }
+- [:octicons-code-16:](https://github.com/GrapheneOS/Vanadium){ .card-link title="Source Code" }
+
+</details>
+
+</div>
```
redoomed1, line 173:
```suggestion
</div>

#### Recommended Vanadium Configuration

<div class="annotate" markdown>

As explained in their documentation, Vanadium prioritizes better privacy- and security-enhancing defaults to avoid users making themselves stand out by opting into non-default settings.

You might, however, consider configuring the following privacy-related settings, all of which can be found in :material-dots-vertical: → **Settings** → **Privacy and security**.

- [x] Select **Close tabs on exit**
- [x] Select **Open external links in incognito** (1)
- [x] Select **Lock Incognito tabs when you leave Vanadium**

</div>

1. With this setting enabled, **all** links from other apps, custom tabs, search intents and share intents will open in incognito tabs.
```

redoomed1: > As explained in their documentation, Vanadium prioritizes better privacy- and security-enhancing defaults to avoid users making themselves stand out by opting into non-default settings.

I reference the paragraph starting with "Extension support isn't planned" on their [documentation](https://grapheneos.org/features#vanadium).

docs/mobile-browsers.md
```diff
@@ -128,6 +152,26 @@ Brave allows you to select additional content filters within the internal `brave
...
+
+**Vanadium** is the default browser that is included with [GrapheneOS](../android.md#grapheneos), it includes a number of [patches](https://github.com/GrapheneOS/Vanadium/tree/main/patches) which assist in improving security and privacy of WebView.
+
+Vanadium is built upon the Chromium web browser project, so it should feel familiar and have minimal website compatibility issues.
+
+<details class="downloads" markdown>
+<summary>Downloads</summary>
+
+- [:octicons-home-16: Homepage](https://vanadium.app){ .md-button .md-button--primary }
+- [:octicons-code-16:](https://github.com/GrapheneOS/Vanadium){ .card-link title="Source Code" }
```
redoomed1, line 169:
```suggestion
[:octicons-home-16: Homepage](https://vanadium.app){ .md-button .md-button--primary }
[:octicons-info-16:](https://grapheneos.org/usage#web-browsing){ .card-link title=Documentation }
[:octicons-code-16:](https://github.com/GrapheneOS/Vanadium){ .card-link title="Source Code" }
```

redoomed1 (changes requested):
This review is for suggesting changes to the description and the recommended configuration.

jonaharagon:
> For people using [GrapheneOS] (android.md#grapheneos) on a [Google Pixel] (android.md#google-pixel), we recommend using Vanadium

I'm not okay with telling GrapheneOS users they _must_ use Vanadium at this time, **and** I'm not okay with the first mobile browser we list only being available on a niche custom ROM and not all of Android.

In fact, I would prefer these headers in this order unless we are also going to add Mulch (https://discuss.privacyguides.net/t/mulch-android-browser/14461) which is more widely available:

1. Brave
2. Mull
3. Vanadium (GrapheneOS)

This is consistent with how we handle ordering based on platform availability and popularity across the site.

dngray:
> I'm not okay with telling GrapheneOS users they _must_ use Vanadium at this time, **and** I'm not okay with the first mobile browser we list only being available on a niche custom ROM and not all of Android.

Fair point, given it only works on GOS. Lets just keep it alphabetical and have the random table at the top maybe.

redoomed1:
> This is consistent with how we handle ordering based on platform availability and popularity across the site.

Makes sense, I appreciate all the reasoning you gave.

> have the random table at the top maybe.

I like this idea and see high value in adding it, considering that there have been posts on the forum like [Chromite vs Vanadium (gOS)](https://discuss.privacyguides.net/t/chromite-vs-vanadium-gos/13758) for which a randomized table would have been useful.

docs/tools.md
```diff
@@ -42,6 +42,7 @@ For more details about each project, why they were chosen, and additional tips o
 <div class="grid cards" markdown>
 
 - ![Brave logo](assets/img/browsers/brave.svg){ .twemoji loading=lazy } [Brave (Android)](mobile-browsers.md#brave)
+- ![Vanadium logo](assets/img/browsers/vanadium.svg){ .twemoji loading=lazy }![Vanadium logo](assets/img/browsers/vanadium-dark.svg#only-dark){ .twemoji loading=lazy } [Vanadium (GrapheneOS)](mobile-browsers.md#vanadium-grapheneos)
```
redoomed1, line 45:
```suggestion
- ![Vanadium logo](assets/img/browsers/vanadium.svg#only-light){ .twemoji loading=lazy }![Vanadium logo](assets/img/browsers/vanadium-dark.svg#only-dark){ .twemoji loading=lazy } [Vanadium (GrapheneOS)](mobile-browsers.md#vanadium-grapheneos)
```
Oops, missed this in my first review :sweat_smile:

redoomed1 (changes requested):

jonaharagon:
Maybe the comparison table could work. I wasn't intending to use that outside of providers though. The main difference is that the providers we recommend are largely interchangeable, whereas the software we recommend is either more often intended for a specific use-case, or more likely to be chosen based on personalized factors, or both. With a random table we'll have, for example, iOS browsers interspersed with Android browsers (OS is not the only factor but it's most obvious) which just makes it difficult to read at a glance.

*So,* when writing about software (like browsers), it makes more sense to me to have a narrative flow going from more broad to more specific. There are a couple factors as far as why I prefer [this order](https://github.com/privacyguides/privacyguides.org/pull/1978#issuecomment-2058050165), but I'll give the most obvious narrative here as an example:

1. Start with **Brave**: usable by any Android user, decent defaults out of the box <- makes this a good starting point to base our mobile browser discussion around.

    - Then, ask: why would someone not use Brave?

        1. I think the most _common_ reason is a desire to avoid Chromium (valid): Which is to say, I believe the percentage of people using Android who don't wish to use Chromium is higher than the percentage of people using GrapheneOS. Does that seem like a fair assessment?

            - Thus, we present **Mull** & explain the benefits and drawbacks of Mull on Android, etc.

        2. Then, we go *back* to the question above. Given our audience, the _second_ most likely reason is probably that they are currently using GrapheneOS.

            - Thus, we present **Vanadium** with a note about the benefits on GrapheneOS specifically

So the order would not be interpreted as Mull > Vanadium, because as you see we are not comparing those two directly in our narrative at all. The order of those two is not related to their features, but to factors like flow and relevance to the reader. What I want to avoid is thinking about the order strictly linearly, when it would be considered more like this:

```mermaid
flowchart
	n2{{"Hexagon"}}
	n2{{"Brave"}} --- n5{{"Mull"}}
	n2 --- n1{{"Vanadium"}}
```

And after mapping it out, then we would present it from top to bottom, left to right in writing.* And again to be perfectly clear this diagram isn't a mapping of what we think is "best" feature-wise either, but what we think is the most relevant recommendation. I don't think it's controversial to say that if a random Android user on the street asked me to recommend a browser, Brave is the obvious go-to. Everything else requires additional considerations like what device do you have, etc.

---

\* For the sake of completeness (in case this comment is useful for future reference), you could expand this tree concept out to the whole page to get a broader idea of how the tree translates to linear header order when writing:

```mermaid
flowchart
	n1[/"1. Mobile Browsers"\]
	n1 --- n2["2. Android"]
	n1 --- n3["7. iOS"]
	n2 --- n4["3. Brave"]
	n4 --- n5["4. Mull"]
	n4 --- n6["5. Vanadium"]
	n3 --- n7["8. Safari"]
	n7 --- n8["9. Theoretical\n second iOS browser"]
	n7 --- n9["10. Theoretical\n third iOS browser"]
	n6 --- n10["6. Mulch?\nJust for example"]
```

jonaharagon:
In this case I guess Brave/Mull/Vanadium is already alphabetical, so now that I think about it my comment will really have no impact on this PR... but I wanted to share why the order makes sense for reasons _other_ than that they're merely alphabetically ordered, and give us something to think about when deciding the order of other pages.

redoomed1:
Thank you for taking the time to map out and elaborate your rationale for how the recommendations on the Mobile Browsers page are presented!

> The main difference is that the providers we recommend are largely interchangeable, whereas the software we recommend is either more often intended for a specific use-case, or more likely to be chosen based on personalized factors, or both.

I agree with your assessment for the most part.

> With a random table we'll have, for example, iOS browsers interspersed with Android browsers (OS is not the only factor but it's most obvious) which just makes it difficult to read at a glance.

This is also a good point that I overlooked. I initially thought that the idea of a table that dngray brought up would be helpful since there are two pending additions to the Mobile Browsers page (Vanadium and Cromite), and a table would help to elucidate the differences between them.

On second thought, however, I think that it would simply reinvent the wheel, the wheel being the [DivestOS Browser Comparison Table](https://divestos.org/pages/browsers).

friadev:
> @dngray are we still going on with this or can this be closed?

I think we should add this, Vanadium is the best option on GrapheneOS right now.

jonaharagon:
These recommendations are published at FORUM-27338. I'll leave it up to you whether this is sufficient for now (and then please close this PR), or if you want to update this PR and merge this to add Vanadium as a recommendation @dngray. I'm fine either way really.

jonaharagon:
I don't really see anyone pushing for this 👀

PR #1986, opened 2023-02-15
Other uBO lists

dngray:
I reworded the description to sound a bit better.

It seems that `AdGuard URL Tracking Protection` and `Actually Legitimate URL Shortener Tool` were removed, https://github.com/privacyguides/privacyguides.org/pull/1238/files and these are pretty harmless. This makes our recommendations compatible with https://github.com/arkenfox/user.js/wiki/4.1-Extensions#-recommended which I think is quite reasonable.

Also came up in https://github.com/privacyguides/privacyguides.org/discussions/1328


mfwmyfacewhen (approved):

jonaharagon (approved):
SGTM

PR #1276, opened 2022-05-20
Listing common threat examples

jonaharagon:
Alternative implementation of #960 

https://privacyguides-privacyguides-org-preview-pr-1276.surge.sh/basics/common-threats/

Page sections to write:

- [x]  Common threats categories
- [x]  Anonymity vs Privacy
- [x]  Privacy from Service Providers
- [x]  Security and Privacy (section on how they are related/rely on each other)
- [x]  Mass Surveillance Programs
- [x]  Limiting Public Information
- [x]  Censorship

Tasks to complete:

- [x] Add CTA to threat modeling page https://privacyguides-privacyguides-org-preview-pr-1276.surge.sh/basics/threat-modeling/

Removed for future PR:

- ~~Add tags to browser recommendations (for example) https://privacyguides-privacyguides-org-preview-pr-1276.surge.sh/browsers/~~
- ~~Add tags to all recommendations~~

Closes #960

PR #1987, opened 2023-02-15
Add some other threat modeling links

dngray:
This incorporates discussions from

- https://github.com/privacyguides/privacyguides.org/discussions/740

Note I did not include https://teamsammut.com/scf/ as it seems not really to our target audience. We could use a [graphic on that page](https://teamsammut.com/files/images/scf-boxes.jpg) like that though. On a side note, I do like some of the pictures used in this article https://opsec101.org

- https://github.com/privacyguides/privacyguides.org/discussions/1631

### DIFF
--- a/docs/basics/threat-modeling.en.md
+++ b/docs/basics/threat-modeling.en.md
@@ -95,7 +95,11 @@ Making a security plan will help you to understand the threats that are unique t
 
 ## Further Reading
 
-For people looking to increase their privacy and security online, we've compiled a list of common threats our visitors face or goals our visitors have, to give you some inspiration and demonstrate the basis of our recommendations. 
+For people looking to increase their privacy and security online, we've compiled a list of common threats our visitors face or goals our visitors have, to give you some inspiration and demonstrate the basis of our recommendations.
+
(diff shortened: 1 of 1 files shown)


mfwmyfacewhen (approved):

jonaharagon:
#740 saying that this page needs improvements is very out of date, we discussed this page extremely thoroughly when we did #1819 and #1276 and I think we were all happy with it, I don't really know what this adds.

If you like information from these other websites, I would rather that information be incorporated onto this page and those links added to the sources section instead, generally speaking I don't think we should direct people elsewhere for further reading.

dngray:
> If you like information from these other websites, I would rather that information be incorporated onto this page and those links added to the sources section instead, generally speaking I don't think we should direct people elsewhere for further reading.

Yes, which is why I did mark this as a draft. It is more of a reminder for me to think about how it could fit into the page.

freddy-m:
Do we still want to do this @privacyguides/team?

jonaharagon:
I don't think we still want to do this, no.

dngray:
Agreed, we don't really need to do this anymore, we've got a better page IMO

PR #1994, opened 2023-02-15
Remove Session

mfwmyfacewhen:
I would like to tighten our requirements to include perfect forward secrecy, which would exclude session.

- [x] Please check this box to confirm your agreement to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform, relicense, and distribute your contribution as part of our project.
- [x] Please check this box to confirm you are the sole author of this work, or that any additional authors will also reply to this PR on GitHub confirming their agreement to these terms.


jonaharagon (changes requested):
I don't have a problem with this, but it has to be removed from the tools page and the logos have to be removed.

blacklight447:
I am personally split on this topic.
On one side I do think this would be a good addition to the criteria.

On the other hand, I think we too easily delist things. People can visit our page one day and decide to do the work and migrate to a new messenger, only to be annoyed to see its no longer recommended the next day.

How about we move the addition of pfs to the "best case" section first, notify projects that do not reach that criteria, and then move it up to minimum criteria by 2024. That way we have a way of moving the bar in an upward trend with having to constantly remove and adding recommendations.

What do you think?

dngray:
Does PFS even make sense for a IM that runs on a self contained network? We know that Session uses the [Lokinet](https://github.com/privacyguides/privacyguides.org/discussions/810) network, which my understanding is that is alternative to .onion style services.

If we added a no PFS requirement, how would that effect things like Briar?

mfwmyfacewhen:
> If we added a no PFS requirement, how would that effect things like Briar?

Does Briar not have PFS? Because we state on the site that it does.

dngray:
> Does Briar not have PFS? Because we state on the site that it does.

It appears [it does](https://code.briarproject.org/briar/briar-spec/-/blob/master/protocols/BTP.md).

mfwmyfacewhen:
> > Does Briar not have PFS? Because we state on the site that it does.
> 
> It appears [it does](https://code.briarproject.org/briar/briar-spec/-/blob/master/protocols/BTP.md).

So it shouldn't be a problem, right?

blacklight447:
I gave this some more thought, and considering banana's about how we already include SimpleX, I think going forward with removing session in favor of adding PFS to the minimum requirements.

blacklight447 (approved):

jonaharagon (dismissed):
> if anything, I think Privacy Guides has been conservative in many cases on that front.

I agree.

jonaharagon:
We're looking into a separate issue with this, Matrix doesn't use PFS either despite it seeming to be theoretically possible, so this might need to be a larger discussion. See:

- [https://github.com/vector-im/element-web/issues/7101](https://github.com/vector-im/element-web/issues/7101)
- https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/megolm.md#partial-forward-secrecy

blacklight447:
Well to be honest it was never really listed as competition for signal, more as option for folks who want a messenger on a federated platform. I would see matrix/element as more of  discord alternative.

IIRC, megolm was audited, but i guess that shows that audits are not the  magic security pixie dust people make them out to be.

PR #2021, opened 2023-02-24
Configure HTTP Headers

jonaharagon:
```
X-Frame-Options = "DENY"
X-XSS-Protection = "0"
X-Content-Type-Options = "nosniff"
Strict-Transport-Security = "max-age=63072000; includeSubDomains; preload"
Content-Security-Policy = "default-src 'none'; script-src https://www.privacyguides.org https://api.privacyguides.net 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; font-src 'self'; img-src data: 'self'; connect-src https://api.github.com https://api.privacyguides.net 'self'"
```

### DIFF
--- a/mkdocs.production.yml
+++ b/mkdocs.production.yml
@@ -2,10 +2,8 @@ INHERIT: mkdocs.yml
 plugins:
   meta: {}
   privacy:
-    external_assets_exclude:
-      - cdn.jsdelivr.net/npm/mathjax@3/*
-      - api.privacyguides.net/*
-      - giscus.app/*
+    external_assets_exclude: ['https://cdn.jsdelivr.net/npm/mathjax@3/*', 'https://api.privacyguides.net/*']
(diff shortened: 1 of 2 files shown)


jonaharagon:
Jury's out on whether `X-XSS-Protection` should be set to `0` (disable browser XSS protection) or `1; mode=block` (enable and block the entire page load if XSS is detected. Hardenize prefers the former and Mozilla Observatory prefers the latter. Modern consensus (https://github.com/github/secure_headers/issues/439) seems to agree that Mozilla is wrong and Hardenize is correct in this case, so I went with `0`. [I think this only affects Safari(?) anyways nowadays](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection#browser_compatibility), so not a big deal outside of losing 10 points on our Mozilla Observatory score.

PR #1984, opened 2023-02-15
Mention linux updates

dngray:
This closes:

- https://github.com/privacyguides/privacyguides.org/discussions/1281
- https://github.com/privacyguides/privacyguides.org/discussions/1280

I think it's important not to duplicate too much of upstream documentation.

Windows and macOS instructions should in their respective pages.

PR #2030, opened 2023-02-26
Permissions and Tracking (Android)

dngray:
- [x] Please check this box to confirm your agreement to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform, relicense, and distribute your contribution as part of our project.
- [x] Please check this box to confirm you are the sole author of this work, or that any additional authors will also reply to this PR on GitHub confirming their agreement to these terms.

### DIFF
--- a/docs/os/android-overview.md
+++ b/docs/os/android-overview.md
@@ -52,9 +52,44 @@ It's important to not use an [end-of-life](https://endoflife.date/android) versi
 
 ## Android Permissions
 
-[Permissions on Android](https://developer.android.com/guide/topics/permissions/overview) grant you control over what apps are allowed to access. Google regularly makes [improvements](https://developer.android.com/about/versions/11/privacy/permissions) on the permission system in each successive version. All apps you install are strictly [sandboxed](https://source.android.com/security/app-sandbox), therefore, there is no need to install any antivirus apps. A smartphone with the latest version of Android will always be more secure than an old smartphone with an antivirus that you have paid for. It's better not to pay for antivirus software and to save money to buy a new smartphone such as a Google Pixel.
+[Permissions on Android](https://developer.android.com/guide/topics/permissions/overview) grant you control over what apps are allowed to access. Google regularly makes [improvements](https://developer.android.com/about/versions/11/privacy/permissions) on the permission system in each successive version. All apps you install are strictly [sandboxed](https://source.android.com/security/app-sandbox), therefore, there is no need to install any antivirus apps.
 
-Should you want to run an app that you're unsure about, consider using a user or work profile.
+A smartphone with the latest version of Android will always be more secure than an old smartphone with an antivirus that you have paid for. It's better not to pay for antivirus software and to save money to buy a new smartphone such as a Google Pixel.
+
+Android 10:
+
+- [Scoped Storage](https://developer.android.com/about/versions/10/privacy/changes#scoped-storage) gives you more control over your files and can limit what can [access external storage](https://developer.android.com/training/data-storage#permissions). Apps can have a specific directory in external storage as well as the ability to store specific types of media there.
+- Tighter access on [device location](https://developer.android.com/about/versions/10/privacy/changes#app-access-device-location) by introducing the `ACCESS_BACKGROUND_LOCATION` permission. This prevents apps from accessing the location when running in the background without express permission from the user.
+
+Android 11:
+
+- [One-time permissions](https://developer.android.com/about/versions/11/privacy/permissions#one-time) which allows you to grant a permission to an app just once.
+- [Auto-reset permissions](https://developer.android.com/about/versions/11/privacy/permissions#auto-reset), which resets [runtime permissions](https://developer.android.com/guide/topics/permissions/overview#runtime) that were granted when the app was opened.
+- Granular permissions for accessing [phone number](https://developer.android.com/about/versions/11/privacy/permissions#phone-numbers) related features.
+
+Android 12:
+
+- A permission to grant only the [approximate location](https://developer.android.com/about/versions/12/behavior-changes-12#approximate-location).
+- Auto-reset of [hibernated apps](https://developer.android.com/about/versions/12/behavior-changes-12#app-hibernation).
+- [Data access auditing](https://developer.android.com/about/versions/12/behavior-changes-12#data-access-auditing) which makes it easier to determine what part of an app is performing a specific type of data access.
+
+Android 13:
+
+- A permission for [nearby wifi access](https://developer.android.com/about/versions/13/behavior-changes-13#nearby-wifi-devices-permission). The MAC addresses of nearby WiFi access points was a popular way for apps to track a user's location.
+- More [granular media permissions](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions), meaning you can grant access to images, videos or audio files only.
+- Background use of sensors now requires the [`BODY_SENSORS`](https://developer.android.com/about/versions/13/behavior-changes-13#body-sensors-background-permission) permission.
+
+An app may request a permission for a specific feature it has. For example, any app that can scan QR codes will require the camera permission. Some apps can request more permissions than they need.
+
+[Exodus](https://exodus-privacy.eu.org/) can be useful when comparing apps that have similar purposes. If an app requires a lot of permissions and has a lot of advertising and analytics this is probably a bad sign. We recommend looking at the individual trackers and reading their descriptions rather than simply **counting the total** and assuming all items listed are equal.
+
+!!! warning
+
+    If an app is mostly a web-based service, the tracking may occur on the server side. [Facebook](https://reports.exodus-privacy.eu.org/en/reports/com.facebook.katana/latest/) shows "no trackers" but certainly does track users' interests and behavior across the site. Apps may evade detection by not using standard code libraries produced by the advertising industry, though this is unlikely.
+
+!!! note
+
+    Privacy-friendly apps such as [Bitwarden](https://reports.exodus-privacy.eu.org/en/reports/com.x8bit.bitwarden/latest/) may show some trackers such as [Google Firebase Analytics](https://reports.exodus-privacy.eu.org/en/trackers/49/). This library includes [Firebase Cloud Messaging](https://en.wikipedia.org/wiki/Firebase_Cloud_Messaging) which can provide [push notifications](https://en.wikipedia.org/wiki/Push_technology) in apps. This [is the case](https://fosstodon.org/@bitwarden/109636825700482007) with Bitwarden. That doesn't mean that Bitwarden is using all of the analytics features that are provided by Google Firebase Analytics.
 
 ## Media Access
 


dngray:
> I'm not so sure that this makes sense. I do like the sentence in there that explains that this approach is flawed (as can be seen with Facebook), which simply highlights how bypassable such "tracker lists" are.

I wanted to highlight that point, as the site is regularly mentioned without any of the caveats.

> In other cases, (such as with [Bitwarden](https://reports.exodus-privacy.eu.org/en/reports/com.x8bit.bitwarden/latest/)), the "trackers" which are mentioned are not actually trackers at all, and are being used to provide app functionality. 
>
> The approach with labeling a specific library as a "tracker" no matter how it's being used within the app, or whether it's opt-in or opt-out makes the usefulness of such tools, dubious.

Oh, that's not really too useful. I thought it did more analysis than just looking at imports

> Can you provide me with some more information about how the decision to include this came up?

I think we discussed it on Matrix some time ago. It does come up from time to time, so I thought it might be useful as a preliminary investigation tool.

docs/android.en.md
```diff
@@ -232,6 +232,18 @@ Main privacy features include:
...
+
+If you are unsure about whether an app might contain trackers, we suggest looking at the report produced by [https://reports.exodus-privacy.eu.org](https://reports.exodus-privacy.eu.org). This site can produce an automated report and present information such as the required permissions and trackers detected in an app.
+
+!!! note
+
+    Android applications often require permissions for simple features. Don't be alarmed if your app requires a certain permission. If you are concerned investigate further.
+
+!!! warning
+
+    Exodus is able to determine some trackers it is not be perfect. Remember to check the app's privacy policy. Facebook for example shows [no trackers](https://reports.exodus-privacy.eu.org/en/reports/com.facebook.katana/latest/), but that does not mean there aren't any the website.
```
noClaps, line 245:
```suggestion
    While Exodus is able to determine some trackers, it is not perfect. Facebook, for example, shows [no trackers](https://reports.exodus-privacy.eu.org/en/reports/com.facebook.katana/latest/), but that does not mean there aren't any on the website. Remember to check the app's privacy policy.
```

dngray:
> Regarding Bitwarden specifically, because I brought it up as an example in my previous reply, here's a reply from Bitwarden themselves after being asked by someone who was looking at the Exodus report about it:

Excellent.

I really think we should use this as an example of how the results might not be accurate. The reason is because it often gets referred to on social media/reddit, and I think the Facebook/Bitwarden example is a perfect opportunity to discuss how these tools aren't 100% foolproof.

blacklight447:
Isnt exodus a form of badness Enumeration?

dngray:
> Isnt exodus a form of badness Enumeration?

No, because it does actually look at the app to see what libraries are used. The point is it does work quite well, but you just need to be careful you're reading the results right.

So for example if we look at an app that says maybe it has some "telemetry", while this is counted as a "tracker", it may be benign or harmless. Likewise with push notifications.

However,

There are many things which are absolutely 100% advertising and profiling, and cannot be anything else, so I do think it has value. If you look at [recent reports](https://reports.exodus-privacy.eu.org/en/reports/list/), and then click on some of the items that are bad for privacy.

jonaharagon:
@matchboxbananasynergy what are your thoughts on recommending the use of Exodus to review app permissions? https://reports.exodus-privacy.eu.org/en/reports/com.facebook.katana/latest/ lays them out relatively nicely, and they highlight permissions which are considered dangerous/special by the Android developer guidelines.

jonaharagon:
Makes sense, I think I agree with you and @blacklight447. 

@dngray I think even if I did agree with you, I don't really see this specific tool being a solid recommendation for the site, seeing as 90% of the text in this PR itself is basically explaining why Exodus is not super useful, so I'm going to close this PR.

Exodus maintaining a list of trackers "which are absolutely 100% advertising and profiling, and cannot be anything else" **is** \<scary-word>*badness enumeration*\</scary-word>, and even though I generally disagree with the premise that "badness enumeration" is always a bad thing which many people here like to believe, I think if that's something you want to do on your own device there are better alternatives, like blocking the trackers entirely with a DNS-level ad/tracker blocking tool for example, so merely listing them like Exodus does is not a super compelling use-case to me.

dngray:
> Dangerous permissions (the ones that are marked as such on Exodus' reports as well) are not granted by default, and require explicit consent from the user for the app to obtain them.

That is fine, it's also important to note that Exodus does link to the [Google article on permissions](https://developer.android.com/guide/topics/permissions/overview) that expands on this further.

> By having a gigantic list of all defined permissions, be they dangerous or not, someone who is trying to evaluate an app may just get overloaded with information and lose the plot, and I don't really want to see someone deciding whether they should use an app on the grounds of "This app requires 20 permission, and this one requires 3, so I'll choose the latter" when the one with the 3 permission may require some of the most invasive/dangerous permissions out there.

I also don't think we should treat every reader as if they are so dumb they can't figure out anything for themselves. It is not a recommendation card, simply an entry on a page that already exists.

I think what we say about it, which is that:

> Android applications often require permissions for simple features. Don't be alarmed if your app requires a certain permission. For example anything that has QR code functionality will require the camera permission. If you are concerned, you will need investigate further.

What I wrote sufficiently addresses that.

> All in all, a list of permissions, much like the "trackers" caught by Exodus doesn't tell the full story, and is a very incomplete tool that doesn't really replace properly looking into an app and determining whether it's something that infringes on your privacy.

Just because one app doesn't have any third party trackers (Facebook's would be internal), and Bitwarden only has some crash analytics, does not make the tool useless.

Most apps will import libraries from adtech developers simply because re implementing that in their own code is going to be annoying to maintain. They're going to want to use simple APIs in a modular method which is in line with maintenance practices of their app. Remember Exodus is such a small website they are not going to be actively mitigating that.

Remember is to not let perfect be the enemy of good.

If you look at example of *other apps* and I suggest you do, you'll see that you can get a pretty good idea from the trackers found, what the app will actually do.

You don't need much imagination to see what these are allowing:

- https://reports.exodus-privacy.eu.org/en/trackers/72/
- https://reports.exodus-privacy.eu.org/en/trackers/65/
- https://reports.exodus-privacy.eu.org/en/trackers/328/ (IAB are [these folk](https://shkspr.mobi/blog/2023/01/the-iab-loves-tracking-users-but-it-hates-users-tracking-them/)

dngray:
So I was having a look at religious apps. In this case the Quaran:

Here's a good example:

- https://reports.exodus-privacy.eu.org/en/reports/com.quran7m.free.main/latest/
  - which in turn had this entry https://reports.exodus-privacy.eu.org/en/trackers/106/
  - In this particular case it shows the privacy policy for Inmobi and lawsuits they were involved in.

The point is it is what is labeled on the box, other Quaran apps, we can see they are more privacy friendly:

- https://reports.exodus-privacy.eu.org/en/reports/quran.ayat.alquran.holyquran/latest/
- https://reports.exodus-privacy.eu.org/en/reports/com.noorisys.sqr/latest/
- https://reports.exodus-privacy.eu.org/en/reports/com.fekracomputers.the.quran/latest/
- https://reports.exodus-privacy.eu.org/en/reports/com.arabiait.quran/latest/

So yes, I still think Exodus can be useful as a tool to decide whether you want an app on your phone.

I am well aware they *could* encode, decode their imports in hex or whatever to hide from the analysis, but I think it's pretty doubtful. It's important to remember the threat model of this app is not to get around Exodus, and they do have to get approved to be on Google Play.

dngray:
I've decided to re-write this whole section and make it more about Permissions and Trackers in general, as well as discuss some of the recent permission changes in supported versions of Android. This is in line with the thinking in https://github.com/privacyguides/privacyguides.org/pull/1984

> If you want to avoid "trackers" in these apps, why wouldn't you just block connections to these services via DNS instead

Perhaps you want to just avoid these things altogether rather than try to neuter them.

> Bitwarden's entry on Exodus will make them not want to use it.

I've expanded upon that, and I still think it is better to explain that, rather than ignore it, considering Bitwarden is a product we also recommend.

> Do apps use tracking and ad SDKs? Absolutely they do. The extent of how much that matters varies wildly.

It can, and you can get a pretty good idea, based on the kinds of data that a specific company collects.

> If Exodus is going to be mentioned in any way shape or form on Privacy Guides, I think the section needs to be much more elaborate and talk about what it can and cannot provide much more, because as much as it can provide some information that may be useful when evaluating whether an app makes sense for you, it can be very confusing and turn people away from great apps because libraries they contain are being misinterpreted.

Perhaps, and I am happy to elaborate on that, because Exodus does get mentioned **a lot**, so we should educate people on that. I have added some of the ones I can think of.

> People are generally **not** going to be able to tell them apart, so we either need to go in depth, or not at all imo.

That will happen anyway whether we mention it or not.

dngray:
> wouldn't this be better under https://www.privacyguides.org/en/os/android-overview/#android-permissions?

I think so, this works

blacklight447 (approved):

docs/android.md
```diff
@@ -320,6 +322,45 @@ That said, the [F-Droid](https://f-droid.org/en/packages/) and [IzzyOnDroid](htt
...
+- Background use of sensors now requires the [`BODY_SENSORS`](https://developer.android.com/about/versions/13/behavior-changes-13#body-sensors-background-permission) permission.
+
+An app may request a permission for a specific feature it has. For example, any app that can scan QR codes will require the camera permission. Some apps can request more permissions than they need.
+
+[Exodus](https://exodus-privacy.eu.org/) can be useful when comparing apps that have similar purposes. If an app requires a lot of permissions and has a lot of advertising and analytics this is probably a bad sign. We recommend looking at the individual trackers and reading their descriptions rather than simply **counting the total** and assuming all items listed are equal.
+
+!!! warning
+
+    If an app is mostly a web-based service the tracking may occur on the server side, [Facebook](https://reports.exodus-privacy.eu.org/en/reports/com.facebook.katana/latest/) shows "no trackers" but certainly does track users interests and behaviour across the site. Apps may evade detection by Exodus by not using standard code libraries produced by the advertising industry, though this is probably not too likely.
```
mfwmyfacewhen, line 358:
```suggestion
    If an app is mostly a web-based service, the tracking may occur on the server side. [Facebook](https://reports.exodus-privacy.eu.org/en/reports/com.facebook.katana/latest/) shows "no trackers" but certainly does track users' interests and behavior across the site. Apps may evade detection by not using standard code libraries produced by the advertising industry, though this is unlikely.
```

docs/android.md
```diff
@@ -320,6 +322,45 @@ That said, the [F-Droid](https://f-droid.org/en/packages/) and [IzzyOnDroid](htt
...
+
+!!! note
+
+    Privacy friendly apps such as [Bitwarden](https://reports.exodus-privacy.eu.org/en/reports/com.x8bit.bitwarden/latest/) may show some trackers such as [Google Firebase Analytics](https://reports.exodus-privacy.eu.org/en/trackers/49/). This library includes [Firebase Cloud Messaging](https://en.wikipedia.org/wiki/Firebase_Cloud_Messaging) which can be to provide [push notifications](https://en.wikipedia.org/wiki/Push_technology) in apps, which happens to [be the case](https://fosstodon.org/@bitwarden/109636825700482007) with Bitwarden. That doesn't mean that Bitwarden is using all of the analytics features that are provided by Google Firebase Analytics.
```
mfwmyfacewhen, line 362:
```suggestion
    Privacy-friendly apps such as [Bitwarden](https://reports.exodus-privacy.eu.org/en/reports/com.x8bit.bitwarden/latest/) may show some trackers such as [Google Firebase Analytics](https://reports.exodus-privacy.eu.org/en/trackers/49/). This library includes [Firebase Cloud Messaging](https://en.wikipedia.org/wiki/Firebase_Cloud_Messaging) which can provide [push notifications](https://en.wikipedia.org/wiki/Push_technology) in apps. This [is the case](https://fosstodon.org/@bitwarden/109636825700482007) with Bitwarden. That doesn't mean that Bitwarden is using all of the analytics features that are provided by Google Firebase Analytics.
```

docs/android.md
```diff
@@ -320,6 +322,45 @@ That said, the [F-Droid](https://f-droid.org/en/packages/) and [IzzyOnDroid](htt
...
+- More [granular media permissions](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions), meaning you can grant access to images, videos or audio files only.
+- Background use of sensors now requires the [`BODY_SENSORS`](https://developer.android.com/about/versions/13/behavior-changes-13#body-sensors-background-permission) permission.
+
+An app may request a permission for a specific feature it has. For example, any app that can scan QR codes will require the camera permission. Some apps can request more permissions than they need.
+
+[Exodus](https://exodus-privacy.eu.org/) can be useful when comparing apps that have similar purposes. If an app requires a lot of permissions and has a lot of advertising and analytics this is probably a bad sign. We recommend looking at the individual trackers and reading their descriptions rather than simply **counting the total** and assuming all items listed are equal.
```
mfwmyfacewhen, line 354:
```suggestion
- [Scoped Storage](https://developer.android.com/about/versions/10/privacy/changes#scoped-storage) gives you more control over your files and can limit what can [access external storage](https://developer.android.com/training/data-storage#permissions). Apps can have a specific directory in external storage as well as the ability to store specific types of media there.
- Tighter access on [device location](https://developer.android.com/about/versions/10/privacy/changes#app-access-device-location) by introducing the `ACCESS_BACKGROUND_LOCATION` permission. This prevents apps from accessing the location when running in the background without express permission from the user.

Android 11:

- [One-time permissions](https://developer.android.com/about/versions/11/privacy/permissions#one-time) which allows you to grant a permission to an app just once.
- [Auto-reset permissions](https://developer.android.com/about/versions/11/privacy/permissions#auto-reset), which resets [runtime permissions](https://developer.android.com/guide/topics/permissions/overview#runtime) that were granted when the app was opened.
- Granular permissions for accessing [phone number](https://developer.android.com/about/versions/11/privacy/permissions#phone-numbers) related features.

Android 12:

- A permission to grant only the [approximate location](https://developer.android.com/about/versions/12/behavior-changes-12#approximate-location).
- Auto-reset of [hibernated apps](https://developer.android.com/about/versions/12/behavior-changes-12#app-hibernation).
- [Data access auditing](https://developer.android.com/about/versions/12/behavior-changes-12#data-access-auditing) which makes it easier to determine what part of an app is performing a specific type of data access.

Android 13:

- A permission for [nearby wifi access](https://developer.android.com/about/versions/13/behavior-changes-13#nearby-wifi-devices-permission). The MAC addresses of nearby WiFi access points was a popular way for apps to track a user's location.
- More [granular media permissions](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions), meaning you can grant access to images, videos or audio files only.
- Background use of sensors now requires the [`BODY_SENSORS`](https://developer.android.com/about/versions/13/behavior-changes-13#body-sensors-background-permission) permission.

An app may request a permission for a specific feature it has. For example, any app that can scan QR codes will require the camera permission. Some apps can request more permissions than they need.

[Exodus](https://exodus-privacy.eu.org/) can be useful when comparing apps that have similar purposes. If an app requires a lot of permissions and has a lot of advertising and analytics this is probably a bad sign. We recommend looking at the individual trackers and reading their descriptions rather than simply **counting the total** and assuming all items listed are equal.
```

mfwmyfacewhen (changes requested):

dngray:
> I was pointing to the fact that there is a section on Android Permissions in the Anroid Overview page, under the Knowledge Base:

Oh, yes you're right. I actually forgot about that.

mfwmyfacewhen (approved):

jonaharagon (approved):

PR #2062, opened 2023-03-01
Use flag for language selector icon

jonaharagon:
Screenshot 2023-03-01 at 11 11 01@2x

### DIFF
--- a/config/mkdocs.common.yml
+++ b/config/mkdocs.common.yml
@@ -16,15 +16,19 @@ extra:
     - name: English
       link: /en/
       lang: en
+      icon: https://raw.githubusercontent.com/twitter/twemoji/master/assets/svg/1f1fa-1f1f8.svg
     - name: Français
       link: /fr/
       lang: fr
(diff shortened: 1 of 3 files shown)


jonaharagon:
@kamilkrzyskow yeah I was testing without that override and it still didn't work, but it seems to be related to #2063 in that case, which is yet another entirely unrelated bug :)

Anyways I can just do this a different way with jinja formatting it turns out, so I'll just go with that for now. Ideally `theme/partials/header.html` could be replaced with dynamic overrides like in https://github.com/squidfunk/mkdocs-material/discussions/5060 in the long-term, but at the moment I'm not good enough at Python to figure that out.

kamilkrzyskow:
Your solution with the additional `icon` in the alternate is very clean. I really like it.

---

I mentioned your partial override, because the commit didn't remove it.
As for the #2063, you're most likely correct, as the `on_post_page` event includes the changes made by thet hook.
But why would it crash for an `.svg` external resource with some Empty Document error 🤔

I tested the hook locally and after removing your custom partial it works fine 😮‍💨 (of course this is the old Public version without the privacy plugin).

flags

blacklight447 (approved):

PR #2096, opened 2023-03-21
MIT License source code

jonaharagon:
This releases the source code (only) as MIT, as the new README section in this PR explains:

> Privacy Guides content is licensed under the Creative Commons Attribution-NoDerivatives 4.0 International Public License, and the underlying source code used to format and display that content on [www.privacyguides.org](https://www.privacyguides.org) is licensed under the MIT License.

In reality we need to do this because:

1. CC BY-ND 4.0 is not recommended for source code.
2. Some files are derivatives of files from https://github.com/squidfunk/mkdocs-material and we were not preserving the required copyright notices in those cases. You'll see he is now credited in many files in this PR.

As far as I am aware I am the only contributor (i.e. copyright holder) to all of these files (according to git blame), hence the notices in my name, if anyone else makes a substantial contribution to the source code in the future their name can be added to a new copyright notice on a separate line in the file as well.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,23 @@
+# Copyright (c) 2020-2023 Jonah Aragon <jonah@triplebit.net>
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to
+# deal in the Software without restriction, including without limitation the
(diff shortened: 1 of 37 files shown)


jonaharagon:
I'm just going to merge this because I am the only copyright holder of all of these files anyways, and it doesn't affect the licensing of anything on the website itself, so I don't believe this needs to be team reviewed.

PR #2097, opened 2023-03-21
Relicensing: CC BY-SA 4.0

jonaharagon:
Changes proposed in this PR:

- Change site license from CC BY-ND 4.0 to CC BY-SA 4.0

We should have consent from all copyright holders to relicense content w/ team approval already, so that is not an issue. My current feeling is that #1820 was probably a mistake, and that the No Derivatives license doesn't really provide us with significant advantages, but it does mean we're not really "giving back" to the community in a sense, it doesn't feel open.


dngray:
My only concern is that some of our articles could be considered "testimony" as in what we observed when we tested something. In such a situation [no derivatives makes sense](https://opensource.stackexchange.com/questions/357/is-the-cc-by-nd-an-open-license):

> The CC BY-ND is not compatible with either the OSI's Open Source definition or the FSF's Free Software Definition.
>
> That said, the FSF still considers it to have an appropriate use that is compatible with their movement: to licence opinions and testimonies. The licence shouldn't be used for documentation or project assets, but they do consider it to be appropriate for works which represent the thoughts of a single person and which it would be misleading to change. To quote [Richard Stallman](http://www.gnu.org/philosophy/copyright-and-globalization.html#opinions):
>
>> The second class of work is works whose purpose is to say what certain people think. Talking about those people is their purpose. This includes, say, memoirs, essays of opinion, scientific papers, offers to buy and sell, catalogues of goods for sale. The whole point of those works is that they tell you what somebody thinks or what somebody saw or what somebody believes. To modify them is to misrepresent the authors; so modifying these works is not a socially useful activity. And so verbatim copying is the only thing that people really need to be allowed to do.

jonaharagon:
I guess that is a convincing argument... We can reopen this if anyone else has a different opinion but I'm fine with closing this issue.

jorgesumle:
> We can reopen this if anyone else has a different opinion 

I have a different opinion.

Non-derivative licenses promote centralization. If I want to make my own version and publish it elsewhere, I can't. If I want to improve something or translate it (maybe I don't want to use the centralized and privacy-invasive software that you promote here), I can't if I don't have the permission.

jonaharagon:
> maybe I don't want to use the centralized and privacy-invasive software that you promote here

If you don't like our work then why use it at all?

jorgesumle:
> If you don't like our work then why use it at all?

I'm talking about Crowdin and GitHub (both proprietary and non-free). It's called [network effect](https://en.wikipedia.org/wiki/Network_effect). I host my projects on [NotABug](https://notabug.org/jorgesumle/) and mostly [contribute translations using Weblate](https://hosted.weblate.org/user/jorgesumle/).

jonaharagon:
Looking at your Weblate profile, the projects that you contribute to are simply very technically different than privacyguides.org.

Weblate was something we heavily investigated, and I had my own Weblate server set up to try to use for this purpose. The problem was that Weblate simply could not work with unstructured Markdown files effectively.

The alternative would have been somehow converting the entire website into individual strings and a format understandable by translation platforms like Weblate, but that would have made it virtually impossible to contribute on the English side of things.

Crowdin has native support for Markdown documents, and as far as I'm aware it is the only free (as in beer) solution that does so.

jorgesumle:
> The problem was that Weblate simply could not work with unstructured Markdown files effectively.

https://github.com/WeblateOrg/weblate/issues/3106

dngray:
Now that we have proper [translations](https://www.privacyguides.org/en/meta/translations/), there is less reason for the whole ND restriction. Anyone plagiarizing the site and cloning it entirely will be punished by the various duplicate content algorithms which search engines have. This encourages things to stay on the main site. Also in the past that restriction was because clones of the website appeared in other languages and were all hugely out of date with source material.

A BY-SA license is more like a GPL license, in that content based off it still has to be available under the license that we provide it under and not more restrictive, so this is also a good thing.

dngray (approved):

hook9 (approved):

blacklight447 (approved):

freddy-m (approved):

PR #2099, opened 2023-03-22
update!: Add Phone Service Providers

### LINKED ISSUE #2009: VOIP Providers
### Discussed in https://github.com/privacyguides/privacyguides.org/discussions/466

<sup>Originally posted by **ph00lt0** December 14, 2021</sup>
We should add a recommendation to use VOIP numbers for phone calls. A couple of reasons:

- You can use multiple numbers on one(or multiple) devices
- You will not be vulnerable to broad-com band attacks
- Your location stays private when making a call

IntelTechniques covered their solution on his website: https://inteltechniques.com/voip.suite.html

Another solution I am aware of: https://jmp.chat/

It should obviously be made very clear that calls and SMS are not private. I am however certain that everyone will need both of them from time to time, and it's still worth it to take measures in this area as well. </div>

jonaharagon:
Changes proposed in this PR:

- Adds phone number provider page
- Replaces duplicate images with symlinks

Closes: #2009 

To-do list:

- [x] **Criteria**
- [ ] JMP.chat logo
- [x] MySudo logo
- [ ] Hushed logo
- [ ] Cheogram logo
- [ ] VOIPSuite logo
- [ ] Silent Link Logo


dngray:
One of the advantages of telnum.net is [they take monero](https://telnum.net/pay-with-cryptocurrencies/) which I'm not aware of the others doing.

Their [once-use disposable SMS numbers](https://telnum.net/services/disposable-number/) are useful for things like Discord. Some rooms there require a set phone number. Discord does do some checking to see if it's a VOIP number, and you don't get charged if no SMS is received from the VOIP number, so you can just try again.

jonaharagon:
There is also https://silent.link/

jonaharagon:
@ph00lt0 see: https://github.com/privacyguides/privacyguides.org/pull/2099/files#diff-1e0a622877fe626bbe5eceee6525b85baed7d4eaf7addea856f4f04b6eaac7f1R77-R79

Google Voice doesn't meet at least two of our criteria. I'm not interested in recommending less privacy-respecting providers on a cost-basis alone, which seems to be the only advantage it has.

jonaharagon:
After researching PGPP a bit more I think that the concerns about it I had in https://github.com/privacyguides/privacyguides.org/discussions/1615#discussioncomment-3355447 are correct, I'm not going to include them.

I did initially think that PGPP might still provide protection against third-party IMSI-catchers like Stingrays even though it doesn't protect against network operator tracking, but even in that case it appears that more advanced catchers can track IM**E**I numbers as well, which is the whole problem anyways.

jonaharagon:
> But one thing they do well is that using cryptogaphy they made it impossible to figure out who pays for which subscription. This could still be a huge advantage that I have not seen elsewhere.

I'm unsure how this is an advantage over paying for Silent Link with Monero?

jonaharagon:
Okay, I'm following you. I added PGPP with the requisite warnings for review, it does seem like there is at least one compelling use-case. I also like that they have unlimited data.

nopeitsnothing:
Consider the [registration policies](https://prepaid-data-sim-card.fandom.com/wiki/Registration_Policies_Per_Country) for each service per-country, as well. Buying a "burner" e-SIM can be considered illegal in many regions, so some providers actually break the law while providing anonymous communication.

BionicBison05:
What about [Cloaked](cloaked.app)? They seem to be more of an all-in-one solution that includes email aliases but one of their notable features is unlimited phone numbers at a flat monthly fee which lends itself to creating a new phone number for every interaction rather than having to decide how to allocate a limited amount of numbers. This would minimize the inconvenience of any one recipient of a phone number doing something that would require you to delete it, since you could just deactivate a unique number rather than having to migrate every service that relies on a shared one.

lepras:
This link right here is too important to not include on the page

https://wiki.soprani.ca/FAQ/Why%20isn%27t%20my%20number%20being%20accepted%20by%20an%20online%20service%3F

Jonah King include it wherever you see fit!

dngray:
> if we should recommend disposable numbers

For accounts you don't really care about it might be fine. Some silly discord room for example where you want to ask about support of a product or something.

They also do rental numbers too.

docs/phone.md
```diff
@@ -0,0 +1,195 @@
...
+
+Therefore, unless you *also* physically swap your phone hardware every few days, ==it would be trivial for the network operator to build a location profile of a specific device despite IMSI randomization, because your IMEI is a static identifier visible to the network.== Additionally, PGPP will not even protect against anything but the most basic third-party [IMSI-catchers](https://en.wikipedia.org/wiki/IMSI-catcher), because most modern IMSI-catchers can track IMEI as well.
+
+This service requires an eSIM compatible Android phone, like the [Google Pixel](android.md#android-devices).
```
redoomed1, line 185: Invisv recently announced the shutdown of PGPP on their blog: https://invisv.com/articles/service_shutdown.html
```suggestion

```

friadev:
I think we should also take a look at [Cape](https://www.cape.co) for this section.

[Google Fi](https://fi.google.com/about/) also provides opportunistic [encrypted](https://support.google.com/fi/answer/11303397?hl=en#zippy=%2Chow-end-to-end-encryption-works) calls between Fi customers, and Google Messages provides E2EE RCS texting between Google Messages users, although this will soon be cross-platform. They also have their very interesting [Fi Virtual Carrier Network](https://www.gstatic.com/fi/wormhole/whitepaper-a00cc4732620f382da5b7aac2bcb6905f970ba6b.pdf).

PR #2112, opened 2023-04-05
Add additional information

jermanuts:
Changes proposed in this PR:

- 

- [x] I have disclosed any relevant conflicts of interest in my post.
- [x] I agree to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform, relicense, and distribute my contribution as part of this project.
- [x] I am the sole author of this work. 
- [x] I agree to the [Community Code of Conduct](https://www.privacyguides.org/en/code_of_conduct/).

PR #2117, opened 2023-04-08
VPN Overview Rewrite

### LINKED ISSUE #2118: VPN ownership article for blog
### Affected page

https://www.privacyguides.org/en/basics/vpn-overview/#sources-and-further-reading

### Description

Looking at the VPN page we could clean up the "Sources and Further Reading" and "Related VPN Information" sections by writing our own article on VPN ownership.

I'm not a fan of cross linking to those sites mentioned in https://github.com/privacyguides/privacyguides.org/pull/2117 and https://github.com/privacyguides/privacyguides.org/pull/2112/files#diff-6ff877175c1e0d1e526cece2a0fb849033a13b945a45da94f309780edb4d0f6f (to a lesser degree) due to the fact they're owned by astro-turfing companies, or VPN companies themselves.

The data is public information, so there's no issues of plagiarism there.

The map on Windscribe's article is a nice touch though https://embed.kumu.io/9ced55e897e74fd807be51990b26b415#vpn-company-relationships and we might be able to do something with that.

### Sources

a

### Before submitting

- [X] I am reporting something that is verifiably incorrect, not a suggestion or opinion.
- [X] I agree to the [Community Code of Conduct](https://www.privacyguides.org/en/code_of_conduct/).

mfwmyfacewhen:
Changes proposed in this PR:

- Incorporate additional info links into the article
- Update info and elaborate on things that need to be more in-depth

closes #2118 


docs/basics/vpn-overview.md
```diff
@@ -12,7 +12,7 @@ A VPN can help as it can shift trust to a server somewhere else in the world. As
 
 ## Should I use a VPN?
 
-**Yes**, unless you are already using Tor. A VPN does two things: shifting the risks from your Internet Service Provider to itself and hiding your IP from a third-party service.
+You should [take IVPN's quick test](https://www.doineedavpn.com) to see if a VPN is right for your threat model. A VPN does two things: shifting the risks from your Internet Service Provider to itself and hiding your IP from a third-party service.
```
jonaharagon, line 15: This I think is the same issue I mentioned here: https://github.com/privacyguides/privacyguides.org/pull/2112#issuecomment-1500928034

> If information is important, we should be adding it to the site directly.

docs/basics/vpn-overview.md
```diff
@@ -60,18 +60,4 @@ A VPN may still be useful to you in a variety of scenarios, such as:
 1. Hiding your downloads (such as torrents) from your ISP and anti-piracy organizations.
 1. Hiding your IP from third-party websites and services, preventing IP based tracking.
 
-For situations like these, or if you have another compelling reason, the VPN providers we listed above are who we think are the most trustworthy. However, using a VPN provider still means you're *trusting* the provider. In pretty much any other scenario you should be using a secure**-by-design** tool such as Tor.
```
dngray, line 63: I wouldn't actually remove these. If we can think of a way to reformat them into the page that would be good.

docs/basics/vpn-overview.md
```diff
@@ -60,18 +60,4 @@ A VPN may still be useful to you in a variety of scenarios, such as:
...
-
-## Sources and Further Reading
-
-1. [VPN - a Very Precarious Narrative](https://schub.io/blog/2019/04/08/very-precarious-narrative.html) by Dennis Schubert
```
dngray, line 67: This one may be condensed and perhaps incorporated into our vpn-overview without depending on this entire article.

dngray:
There was one article from https://github.com/privacyguides/privacyguides.org/pull/2112/ worth saving imho https://blog.windscribe.com/the-vpn-relationship-map.

I don't think it really makes sense for us to reproduce the VPN relationship content that others have authored, but I feel mentioning it is still worthwhile.

docs/basics/vpn-overview.md
dngray: Sounds good to me.

docs/basics/vpn-overview.md
```diff
@@ -60,18 +60,4 @@ A VPN may still be useful to you in a variety of scenarios, such as:
...
-1. [Tor Network Overview](../advanced/tor-overview.md)
-1. [IVPN Privacy Guides](https://www.ivpn.net/privacy-guides)
-1. ["Do I need a VPN?"](https://www.doineedavpn.com), a tool developed by IVPN to challenge aggressive VPN marketing by helping individuals decide if a VPN is right for them.
-
-## Related VPN Information
-
-- [The Trouble with VPN and Privacy Review Sites](https://blog.privacyguides.org/2019/11/20/the-trouble-with-vpn-and-privacy-review-sites/)
```
dngray, line 74: Probably want to keep this one, as it's a link to our previous blog article.

docs/basics/vpn-overview.md
```diff
@@ -60,18 +60,4 @@ A VPN may still be useful to you in a variety of scenarios, such as:
 1. Hiding your downloads (such as torrents) from your ISP and anti-piracy organizations.
 1. Hiding your IP from third-party websites and services, preventing IP based tracking.
 
-For situations like these, or if you have another compelling reason, the VPN providers we listed above are who we think are the most trustworthy. However, using a VPN provider still means you're *trusting* the provider. In pretty much any other scenario you should be using a secure**-by-design** tool such as Tor.
-
-## Sources and Further Reading
```
jonaharagon, line 65: We **do** want to keep a "Sources" section on pages where the information is based on pre-existing sources, so we can't just remove all of this I think. It's only the "Further Reading" and "Related Information" links we want to eliminate.

jonaharagon:
We should look into them more and probably mention them. The main issue is that the only two commercial options I know of are the two mentioned in that blog post (the blog author's and iCloud), but since Tor is also technically a Multi-Party Relay, it would be good to outline the different threat models they serve compared to traditional VPNs.

docs/basics/vpn-overview.md
dngray: I added this one to our notice board https://github.com/privacyguides/privacyguides.org/issues/2118 because it is one which will take some time to put together.

docs/basics/vpn-overview.md
dngray: That does make sense.

dngray:
I also wouldn't be opposed to writing something about DVPNs and mentioning https://www.ivpn.net/privacy-guides/comparing-dvpns-centralized-vpns-privacy-protection/ and the risks with them and also why we don't recommend them. They do come up from time to time and it would be nice to be able to point to a particular section. While that article is by a centralized VPN service, it does raise some very important points being that:

- You don't have transparency over who runs the nodes, they could be logging
- Most are not multi-node services like Tor, and are single nodes, which yield less security than regular centralized VPN service with infrastructure that is regularly audited and set up by professionals
- A lot of them are closed networks, like .i2p, .onion so there's actually no proxying regular data over them.
- I briefly tried to discuss it here https://discuss.privacyguides.net/t/decentralized-vpns-and-routing-networks/11818 but we all came to the same conclusion
- Further when I last checked some were rather [lacking in documentation](https://github.com/privacytools/privacytools.io/issues/1831#issuecomment-613543963)
- and had very few nodes in general. One of them I looked at was about 400, meaning that a some kind of sybil attack would be quite possible with low cost to the attacker.

docs/basics/vpn-overview.md
freddy-m: Definitely keep this one, its always neat to promote the blog :)

jermanuts:
@jonaharagon The problem with all of MPRs (two-hop) is that they require some server (or servers) which relay the traffic and someone has to pay for those servers and their associated bandwidth. iCloud Private Relay and most VPNs are not free, so the user is the one who pays. Tor (three hops) is different: instead of having a single provider such as Apple or your VPN provider, Tor servers are operated by the Tor community on a volunteer basis and are free to users.

While the problem with tor is the 1st and 3rd nodes with MPRs (two hops), you have to trust the both relay companies to not [correlate/share ](https://lobste.rs/s/hzxakf/why_vpns_are_wrong_for_privacy_mprs_are#c_x9an6i) information about connected user.

@dngray sounds good. It is also worth to mention https://educatedguesswork.org/posts/public-wifi/ (biggest selling point for VPNs)

jermanuts:
>Well, this is not a problem, all of our VPN recommendations are paid services, as they should be.

I actually meant that you have to pay using your credit card to use these commercial MPRs.

docs/basics/vpn-overview.md
mfwmyfacewhen: I tried to incorporate these into the page, let me know what you think.

docs/basics/vpn-overview.md
```diff
@@ -5,74 +5,114 @@ icon: material/vpn
...
+### For E2EE
+
+Encryption offered by VPN providers is between your devices and their servers. It guarantees that this specific link is secure. This is a step up from using unencrypted proxies where an adversary on the network can intercept the communications between your devices and said proxies and modify them. However, encryption between your apps or browsers with the service providers are not handled by this encryption.
+
+In order to keep what you actually do on the websites you visit private and secure, you must use HTTPS. This will keep your passwords, session tokens, and queries safe from the VPN provider. Always check for the lock icon in your browser's URL bar to mitigate downgrade attacks like [SSL Strip](https://www.blackhat.com/presentations/bh-dc-09/Marlinspike/BlackHat-DC-09-Marlinspike-Defeating-SSL.pdf).
```
namazso, line 58: should probably just recommend HTTPS-only mode instead, no one is going to check little locks everywhere.

mfwmyfacewhen: I think the lock icon is going the [way of the dodo](https://blog.chromium.org/2023/05/an-update-on-lock-icon.html) anyway.

dngray:
Merged into https://github.com/privacyguides/privacyguides.org/tree/mfmyfw-pr-info

PR #2111, opened 2023-04-05
Enable Spanish Translation

jonaharagon:
Changes proposed in this PR:

- Enable `/es/`

PR #2120, opened 2023-04-10
Download translations dynamically

jonaharagon:
Removes translations from this Git repo, Netlify will download them from Crowdin automatically every time it builds the site (at the expense of making Netlify previews take much longer lol)


blacklight447:
Whats the pro here actually? As i don't think repo size matters that much honestly.

jonaharagon:
@blacklight447 the advantages are a cleaner commit history and changelog; less manual work since we wouldn't have to check translations on here and Crowdin; less confusion about where files should be edited, because currently there are a lot of files which can not be manually edited on GitHub (only on Crowdin) so there's no reason for them to exist here; changes on Crowdin can be previewed on Netlify much easier (for example changes to the upcoming #2111 Spanish translation would be preview-able live quicker); and that the translated versions of the site will be updated much quicker when changes to the main site are made, instead of outdated information and missing pages sitting on them until the next language release.

The disadvantage is that translations won't be manually reviewed via PR here before going live. I'm unsure whether this is a big risk since we don't accept translations from unapproved contributors (not that the approval process is particularly rigorous, but it is a roadblock to spam bots/trolls at least). Also, longer Netlify build times.

blacklight447 (approved):

PR #2149, opened 2023-04-29
Add offline paid subcription for Standard Notes

d4rklynk:
You need to pay an offline subscription while self-hosting

### DIFF
--- a/docs/notebooks.md
+++ b/docs/notebooks.md
@@ -63,7 +63,7 @@ Since January 2023, Joplin supports biometrics app lock for [Android](https://jo
         - [:simple-linux: Linux](https://standardnotes.com)
         - [:octicons-globe-16: Web](https://app.standardnotes.com/)
         
-You still need to pay an [offline subscription](https://standardnotes.com/help/self-hosting/subscriptions) to be able to use premium features for self-hosting. The payement is done by paying monthly or yearly. The prices are really different, monthly is 5.85$/month (meaning 70.2$/year) and yearly is 3.25$/month (meaning 39$/year).
+You still need to pay an [offline subscription](https://standardnotes.com/help/self-hosting/subscriptions) to be able to use premium features for self-hosting.
 
 ### Cryptee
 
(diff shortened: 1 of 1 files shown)


docs/notebooks.md
```diff
@@ -62,6 +62,8 @@ Since January 2023, Joplin supports biometrics app lock for [Android](https://jo
         - [:simple-apple: macOS](https://standardnotes.com)
         - [:simple-linux: Linux](https://standardnotes.com)
         - [:octicons-globe-16: Web](https://app.standardnotes.com/)
+        
+You still need to pay an [offline subscription](https://standardnotes.com/help/self-hosting/subscriptions) to be able to use premium features for self-hosting. The payement is done by paying monthly or yearly. The prices are really different, monthly is 5.85$/month (meaning 70.2$/year) and yearly is 3.25$/month (meaning 39$/year).
```
mfwmyfacewhen, line 66:
```suggestion
You still need to pay an [offline subscription](https://standardnotes.com/help/self-hosting/subscriptions) to be able to use premium features for self-hosting.
```

mfwmyfacewhen (approved):

docs/notebooks.md
mfwmyfacewhen: No reason to list the specific prices/pricing model imo since those can change and not everyone uses USD.

d4rklynk: I just thought it was important since the prices are completely different between monthly and yearly.

freddy-m:
> Standard Notes requires an [offline subscription](https://standardnotes.com/help/self-hosting/subscriptions) for self-hosting.

This seems friendlier, and more to-the-point, to me...

jonaharagon (changes requested):
This isn't strictly true, because you can install community extensions that replicate premium features, e.g. https://github.com/jonhadfield/awesome-standard-notes

I would rather not mention this one way or the other personally, I don't see it being super relevant here.

mfwmyfacewhen:
> This isn't strictly true, because you can install community extensions that replicate premium features, e.g. https://github.com/jonhadfield/awesome-standard-notes
> 
> I would rather not mention this one way or the other personally, I don't see it being super relevant here.

Yeah fair, I think I'll just close this then.

PR #2261, opened 2023-08-09
Fix inaccurate info in macOS

mfwmyfacewhen:
Changes proposed in this PR:

- Remove activation lock
- Change "Allow Applications Downloaded from" setting to be more accurate
- Remove MAC address randomization
- More accurate wording for App Sandbox
- Remove notarization from antivirus setting as it provides very little protection

- [x] I have disclosed any relevant conflicts of interest in my post.
- [x] I agree to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform, relicense, and distribute my contribution as part of this project.
- [x] I am the sole author of this work. 
- [x] I agree to the [Community Code of Conduct](https://www.privacyguides.org/en/code_of_conduct/).

PR #2266, opened 2023-08-17
Fix inaccurate info in macOS

dngray:
Changes proposed in this PR:

- Remove activation lock
- Change "Allow Applications Downloaded from" setting to be more accurate
- Remove MAC address randomization
- More accurate wording for App Sandbox
- Remove notarization from antivirus setting as it provides very little protection

See previous discussion: https://github.com/privacyguides/privacyguides.org/pull/2261

### DIFF
--- a/docs/os/macos-overview.md
+++ b/docs/os/macos-overview.md
@@ -133,20 +133,6 @@ On older Intel-based Mac computers, FileVault is the only form of disk encryptio
 
 - [x] Click **Turn On**
 
-### MAC Address Randomization
-
-Unlike iOS, macOS doesn't give you an option to randomize your MAC address in the settings, so you'll need to do it with a command or a script.
-
-You open up your Terminal and enter this command to randomize your MAC address:
-
-``` zsh
-openssl rand -hex 6 | sed 's/\(..\)/\1:/g; s/.$//' | xargs sudo ifconfig en1 ether 
-```
-
-en1 is the name of the interface you're changing the MAC address for. This might not be the right one on every Mac, so to check you can hold the option key and click the Wi-Fi symbol at the top right of your screen.
-
-This will be reset on reboot.
-
 ## Security Protections
 
 macOS employs defense in depth by relying on multiple layers of software and hardware-based protections, with different properties. This ensures that a failure in one layer does not compromise the system's overall security.
--- a/docs/os/macos-overview.md
+++ b/docs/os/macos-overview.md
@@ -129,12 +129,6 @@ macOS employs defense in depth by relying on multiple layers of software and har
 
     macOS allows you to install beta updates. These are unstable and may come with extra telemetry since they're for testing purposes. Because of this, we recommend you avoid beta software in general.
 
-### App Revocation Checks
-
-macOS performs online OCSP checks using HTTPS encryption when you open an app to verify whether an app contains known malware, and whether the developer’s signing certificate is revoked.
-
-We recommend against blocking these checks.
-
 #### Signed System Volume
 
 macOS's system components are protected in a read-only signed system volume, meaning that neither you nor malware can alter important system files.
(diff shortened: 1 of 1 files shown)


jonaharagon (changes requested):
These changes are based on some random messages on Matrix, sent by somebody who has 0 knowledge in this area. I'll provide additional information about our information here and make any changes that do turn out to be necessary later today.

docs/os/macos-overview.md
```diff
@@ -137,20 +113,6 @@ On older Intel-based Mac computers, FileVault is the only form of disk encryptio
...
-
-You open up your Terminal and enter this command to randomize your MAC address:
-
-``` zsh
-openssl rand -hex 6 | sed 's/\(..\)/\1:/g; s/.$//' | xargs sudo ifconfig en1 ether 
-```
-
-en1 is the name of the interface you're changing the MAC address for. This might not be the right one on every Mac, so to check you can hold the option key and click the Wi-Fi symbol at the top right of your screen.
-
-This will be reset on reboot.
```
jonaharagon, line 152: Given that we cover MAC randomization on other platforms (Linux), removing this section doesn't seem necessary for people who are interested in doing so.

However, this command could be improved as follows:

```
openssl rand -hex 6 | sed 's/^\(.\{1\}\)./\12/; s/\(..\)/\1:/g; s/.$//' | xargs sudo ifconfig en1 ether
```

This would set the second character in the generated MAC address to `2`, denoting it as a randomized MAC address and preventing collisions. Setting the first 3 octets to an existing OUI, as suggested on Matrix, is less preferable (this is the behavior many hypervisors will use to generate a MAC address) because it does not match how iOS and Android generate random MAC addresses, nor the spec.

Technically the second character could be `2`, `6`, `A`, or `E` to denote a locally administered unicast MAC address, but I don't think a single unrandomized bit will affect anyone's privacy.

docs/os/macos-overview.md
```diff
@@ -117,12 +99,6 @@ Decide whether you want personalized ads based on your usage.
 
 - [ ] Uncheck **Personalized Ads**
 
-##### Security
-
-Apps from the App Store are subject to stricter security guidelines, such as stricter sandboxing. If the only apps you need are available from the App Store, change the **Allow applications downloaded from** setting to **App Store** to prevent accidentally running other apps. This is a good option particularly if you are configuring a machine for other, less technical users such as children.
-
-If you choose to also allow applications from identified developers, be careful about the apps you run and where you obtain them.
```
jonaharagon, line 124: I'm okay with removing this section, not because it is bad advice, but because configuring a computer for a less technical user "such as children" might be out of scope for Privacy Guides. Do we think we should be including advice for bubble-wrapping other people's devices on the site?

docs/os/macos-overview.md
```diff
@@ -175,20 +137,19 @@ System Integrity Protection makes critical file locations read-only to protect a
...
+macOS performs online OCSP checks using HTTPS encryption when you open an app to verify whether an app contains known malware, and whether the developer’s signing certificate is revoked.
 
-!!! warning
+We recommend against blocking these checks.
 
-    Software downloaded from outside the official App Store is not required to be sandboxed. You should avoid non-App Store software as much as possible.
+##### App Sandbox
 
-##### Antivirus
+macOS apps downloaded from the App Store are required to use the [App Sandbox](https://developer.apple.com/documentation/security/app_sandbox). You should avoid non-App Store software as much as possible.
```
jonaharagon, line 148: Rearranging these sections doesn't seem to serve a purpose except to create additional work for translators, and the heading levels here just seem to be wrong (h3 on App Revocation Checks underneath an h4)

jonaharagon (changes requested):
The only changes I see in this PR that would improve this guide are removing [L120-L125](https://github.com/privacyguides/privacyguides.org/pull/2266/files#diff-fab019238d4e754539587953bf8fda9b5bab7a7dcd3d210450e47c930040fd5fL120-L125) and changing the MAC address randomization command slightly. I've included detailed notes below.

jonaharagon:
mm why don't we merge #2303 ASAP, and then we can discuss the rest of the changes here further.

freddy-m:
@privacyguides/team / @jonaharagon should we have another look at this now that #2303 is merged?

jonaharagon:
The review comments I added [above](https://github.com/privacyguides/privacyguides.org/pull/2266#pullrequestreview-1658446125) still need to be addressed. If nobody disagrees with what I said there, then no further changes need to be made and this PR can be closed. LMK.

PR #2288, opened 2023-10-04
Capitalize Mac OS overview page title.

blacklight447:
This makes the os overview section more uniform.


jonaharagon (changes requested):
Apple brand names are properly capitalized as shown on https://www.apple.com/legal/intellectual-property/trademark/appletmlist.html

jonaharagon:
Actually this makes it _less_ uniform, because it would no longer match "iOS" directly above it.

PR #2293, opened 2023-10-06
Recommend F-Droid Basic over Neo Store

jonaharagon:
Changes proposed in this PR:

- Recommend the first-party F-Droid Basic client now that it supports unattended upgrades.


dngray (approved):

dngray:
Agreed, I think the language can be softened a bit as they have made some progress according to their blog.

dngray:
@matchboxbananasynergy iirc, that was the original and main reason for not really recommending F-Droid as much as we used to initially. As F-droid allows apps with lower-API levels, that means weaker sandboxing for those apps, than if they were in Google Play etc.

docs/android.md
```diff
@@ -381,17 +381,17 @@ If you download APK files to install manually, you can verify their signature wi
...
+==We only recommend F-Droid as a way to obtain apps which cannot be obtained via the means above.== F-Droid is often recommended as an alternative to Google Play, particularly in the privacy community. The option to add third-party repositories and not be confined to Google's walled garden has led to its popularity. F-Droid additionally has [reproducible builds](https://f-droid.org/en/docs/Reproducible_Builds/) for some applications and is dedicated to free and open-source software. However, there are some security-related downsides to how F-Droid builds, signs, and delivers packages:
```
dngray, line 384: I think if we remove the privsec link here we should mention that apps in Google Play have to target a higher API level, and therefore have stronger sandboxing. We should perhaps show this article https://developer.android.com/google/play/requirements/target-sdk

dngray (changes requested):

jonaharagon:
A forum member [pointed out](https://discuss.privacyguides.net/t/recommend-f-droid-basic-instead-of-neo-store/14311/4) that F-Droid Basic does show a warning on apps with a low target SDK. My understanding is that minimum target SDK for when this warning will appear will advance similarly to Google Play's policy with each new Android release:

> The app that's being installed targets API level 29 (Android 10) or higher. (Google notes that the target API level requirement will advance in future Android versions, a policy that's in line with [Google Play policy](https://www.xda-developers.com/play-store-updated-requirements-api-level-64-bit/) on API target requirement.)

https://www.xda-developers.com/android-12-alternative-app-stores-update-apps-background/

dngray:
> that F-Droid Basic does show a warning on apps with a low target SDK

It looks like it does, but it only says that it won't auto update. I think we should mention something which summarizes:

> Every new Android version introduces changes that bring security and performance improvements and enhance the Android user experience. Some of these changes only apply to apps that explicitly declare support through their `targetSdkVersion` manifest attribute (also known as the target API level).

heeresbach (approved):

jonaharagon:
I really don't see that change as critical given that:

1. We say:
    > Additionally, the requirements for an app to be included in the official F-Droid repo are less strict than other app stores like Google Play, meaning that F-Droid tends to host a lot more apps which are older, unmaintained, or otherwise no longer meet modern security standards.
2. We also say:
    > You should use your best judgement when looking for new apps via this method, and keep an eye on how frequently the app is updated. Outdated apps may rely on unsupported libraries, among other things, posing a potential security risk.
3. F-Droid Basic displays a prominent warning on apps with a low SDK, and that the cutoff to display that warning automatically increases with new Android releases

*So...* I don't know how to word the additional changes you're suggesting to fit it in cleanly on the page tbh, but if you want to commit a change adding that somewhere you think is best, that's fine w/ me.

dngray:
That is fair I guess.

dngray (approved):

PR #2306, opened 2023-10-11
Switch hosting provider to Cloudflare Pages

jonaharagon:
Just exploring this option, this PR would let us move the main website hosting from Netlify to having it built on GitHub Actions and hosted on Cloudflare Pages.

There's a few reasons we might want to do this:

1. Netlify doesn't allow multiple users to access it (well, it does but it's mega expensive). Both @dngray and I have access to our Cloudflare account, and we could grant very granular access to other people as needed for free.
2. Any team member could re-attempt and troubleshoot a failed site build in the [Actions](https://github.com/privacyguides/privacyguides.org/actions) tab on GitHub, instead of needing me to log in to Netlify.
3. Netlify analytics is an additional $9/month, and we are considering using it because of #2305 
4. Benefiting from ECH on Cloudflare's platform might legitimately outweigh the privacy concerns people have with Cloudflare, if ECH is adopted by major browsers.

Cloudflare Pages is *not* as nice as Netlify in some respects:

1. The redirect options are way worse, we can't redirect by browser language or country code, so everyone visiting privacyguides.org would just default to `/en/` unless they manually select their language using the flag on the site.

We would continue using Netlify for PR Previews only, because there are security issues with creating PR Previews with GitHub Actions, and the alternative Cloudflare Pages app we could normally use does not build our site properly.

---

Test preview: https://jonaharagon-cloudflare-pages.preview-privacyguides-org.pages.dev/ - I think everything functions as intended here...

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,160 @@
+# Copyright (c) 2023 Jonah Aragon <jonah@triplebit.net>
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to
+# deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+# sell copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+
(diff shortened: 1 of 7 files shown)


blacklight447 (approved):

freddy-m (approved):

Dyrimon:
Websites protected by Cloudflare are often painful to access in Tor browser. Is this going to affect Tor users?

PR #2317, opened 2023-10-25
Small cosmetic update to Silverblue & Kinoite

eylenburg:
Changes proposed in this PR:

- Desktop recommendation: Also mention Fedora Kinoite in the title 

Previously only Silverblue was mentioned in the title, but both were mentioned in the text. As far as I know, both projects have an equal status in Fedora (unlike Workstation vs the Spins).

- [X] I have disclosed any relevant conflicts of interest in my post.
- [X] I agree to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform, relicense, and distribute my contribution as part of this project.
- [X] I am the sole author of this work. 
- [X] I agree to the [Community Code of Conduct](https://www.privacyguides.org/en/code_of_conduct/).

### DIFF
--- a/docs/desktop.md
+++ b/docs/desktop.md
@@ -60,7 +60,7 @@ A large portion of [Arch Linux’s packages](https://reproducible.archlinux.org)
 
 ## Immutable Distributions
 
-### Fedora Silverblue
+### Fedora Silverblue & Kinoite
 
 !!! recommendation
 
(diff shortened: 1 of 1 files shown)


jonaharagon:
Hm. Another thing to think about is that apparently since writing this recommendation Fedora has also released [Onyx](https://fedoraproject.org/onyx/) and [Sericea](https://fedoraproject.org/sericea/) now, in addition to Silverblue and Kinoite. I'm not sure if we should continually be expanding this recommendation to account for every DE they end up releasing alongside rpm-ostree.

eylenburg:
> Hm. Another thing to think about is that apparently since writing this recommendation Fedora has also released [Onyx](https://fedoraproject.org/onyx/) and [Sericea](https://fedoraproject.org/sericea/) now, in addition to Silverblue and Kinoite. I'm not sure if we should continually be expanding this recommendation to account for every DE they end up releasing alongside rpm-ostree.

Yes I agree with this. I guess Onyx doesn't (yet) support Wayland so it's not meeting the PG criteria. Sericea/Sway should be okay, but probably not as relevant as a beginner wouldn't look for a minimalist tiling window manager, and those people who know what Sway is don't have to get a distro recommendation from PG. 

So from my perspective, best to limit it to just Gnome and KDE which are the two big desktops anyway accounting for probably 70% of all users (e.g. as per https://www.gamingonlinux.com/users/statistics/)

jonaharagon (approved):

dngray (approved):

dngray (changes requested):
I'm not a fan of changing the heading like that.

All of these other distributions are technically part of Silverblue, like Kubuntu is a part of Ubuntu.

I think we should just mention them in an additional paragraph.

eylenburg:
> All of these other distributions are technically part of Silverblue

Well, not really. 

For "normal" Fedora, the website shows "Workstation" and then (separately) "Spins". For the immutable ones that's not the case, they're all shown together without any indication that Silverblue is the main one and the rest is somehow subordinate.

Fedoraproject.og -> Get Fedora

- Editions ("our flagship variants"): Workstation, Server, CoreOS, IoT, Cloud
- Immutable Desktops: Silverblue, Kinoite, Sericea, Onyx
- Spins: KDE Desktop, Xfce Desktop, ...
- Labs: ...

PR #2367, opened 2024-01-15
Use HTML for admonitions

jonaharagon:
Changes proposed in this PR:

- potential fix for crowdin issues

### DIFF
--- a/config/mkdocs-offline.yml
+++ b/config/mkdocs-offline.yml
@@ -155,6 +155,7 @@ nav:
     - Contributing:
       - Writing Guide:
         - 'meta/writing-style.md'
+        - 'meta/admonitions.md'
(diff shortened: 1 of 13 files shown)


jonaharagon:
It seems like the only two pages affected by this are `device-integrity.md` and `photo-management.md` **for now**, so what we could do is fix those two pages now by merging this PR, so that we can actually update the site, and then continue this change over time as needed.

No need to do this all at once, I don't think?

dngray:
> `device-integrity.md` and `photo-management.md` **for now**

No doubt because the other pages went through their old parser.

I think this solution doesn't add too much syntax complexity for contributors, but will mean that the site is not heavily dependent on indentation being just right.

dngray (approved):

PR #2462, opened 2024-03-31
Use GitHub Actions instead of Netlify for builds

jonaharagon:
Changes proposed in this PR:

- Builds the site using GitHub Actions, then deploys that build to Netlify production

Why:

- Netlify's build process is less reliable, so we can eliminate a dependency on that and only use them for hosting
- Lets us swap out hosting providers more easily
- GitHub Actions are (mostly) compatible with Forgejo Actions, so if we leave GitHub this build process can be reused


dngray:
Works for me, although I normally just do a local build anyway.

dngray (commented):
LGTM

PR #2477, opened 2024-04-03
Replace MathJax with MathML

jonaharagon:
So #2476 got me thinking, why don't browsers just support math? duh - and it turns out they do, so we can write math without any dependencies at all.

conversion: https://temml.org/

Closes #2476 obviously.


dngray:
Was pretty sure MathML wasn't supported by Chrome.

dngray (approved):

PR #2496, opened 2024-04-06
Broaden frontend software criteria

TechFanTheo:
Changes proposed in this PR:

Add that website mandating an account and blocking VPNs can also be included. We may also include website that block TOR, but that would significantly escape it's scope, as a lot of websites block TOR．

### DIFF
--- a/docs/frontends.md
+++ b/docs/frontends.md
@@ -252,6 +252,8 @@ Recommended frontends...
 - Must be self-hostable.
 - Must provide all basic website functionality available to anonymous users.
 
-We only consider frontends for websites which are...
+We only consider frontends if one of the following is true for a platform:
 
 - Normally only accessible with JavaScript enabled.
+- Normally only accessible with an account.
+- Blocks access from commercial [VPNs](vpn.md).


jonaharagon:
Thanks for the PR! However, we recommend frontends for which those criteria would not apply, so we can't list them. For example YouTube doesn't block access without an account, but we still recommend frontends for it.

TechFanTheo:
I was thinking of an OR situation, but this wasn't clear. It should be at least one of the criteria. This is necessary because technically Old Reddit doesn't meet the criteria, while it blocks VPN which is absurd.

dngray:
> while it blocks VPN which is absurd.

whereas old. tor onion works just fine.

docs/frontends.md
```diff
@@ -255,3 +255,5 @@ Recommended frontends...
 We only consider frontends for websites which are...
 
 - Normally only accessible with JavaScript enabled.
+- Normally only accesible with an Account.
+- Block access from VPNs.
```
jonaharagon, line 259:
```suggestion
We only consider frontends if one of the following is true for a platform:

- Normally only accessible with JavaScript enabled.
- Normally only accessible with an account.
- Blocks access from commercial [VPNs](vpn.md).
```

dngray (approved):

jonaharagon (approved):

PR #2566, opened 2024-05-08
warning for yubikey MFA for Windows

oppressor1761:
Changes proposed in this PR:

- Point out the potential risks of using third-party software to modify the operating system login process.

<summary>

  </details>

### DIFF
--- a/docs/basics/multi-factor-authentication.md
+++ b/docs/basics/multi-factor-authentication.md
@@ -119,10 +119,6 @@ If you use SMS MFA, use a carrier who will not switch your phone number to a new
 
 Beyond just securing your website logins, multi-factor authentication can be used to secure your local logins, SSH keys or even password databases as well.
 
-### Windows
-
-Yubico has a dedicated [Credential Provider](https://learn.microsoft.com/windows/win32/secauthn/credential-providers-in-windows) that adds Challenge-Response authentication for the username + password login flow for local Windows accounts.
-
 ### macOS
 
 macOS has [native support](https://support.apple.com/guide/deployment/intro-to-smart-card-integration-depd0b888248/web) for authentication with smart cards (PIV). If you have a smartcard or a hardware security key that supports the PIV interface such as the YubiKey, we recommend that you follow your smartcard/hardware security vendor's documentation and set up second factor authentication for your macOS computer.


jonaharagon (changes requested):
Thanks for the PR!

I don't think we need to explicitly say "consider your threat model" on the site though. We could just as well say it under every single recommendation. This is kind of implied...

dngray (approved):

docs/basics/multi-factor-authentication.md
```diff
@@ -121,7 +121,7 @@ Beyond just securing your website logins, multi-factor authentication can be use
 
 ### Windows
 
-Yubico has a dedicated [Credential Provider](https://learn.microsoft.com/windows/win32/secauthn/credential-providers-in-windows) that adds Challenge-Response authentication for the username + password login flow for local Windows accounts. If you have a YubiKey with Challenge-Response authentication support, take a look at the [Yubico Login for Windows Configuration Guide](https://support.yubico.com/hc/articles/360013708460-Yubico-Login-for-Windows-Configuration-Guide), which will allow you to set up MFA on your Windows computer.
+Yubico has a dedicated [Credential Provider](https://learn.microsoft.com/windows/win32/secauthn/credential-providers-in-windows) that adds Challenge-Response authentication for the username + password login flow for local Windows accounts.
```
jonaharagon, line 124: This sentence refers to the same thing as the sentence that was removed, so this should be removed entirely if we don't want to recommend this behavior.

Adding information about Windows Hello instead could be a good alternative to add here, instead of removing this Windows section entirely, but I haven't tested a Yubikey with Windows Hello myself. Does it work for logging in in lieu of a password?

jonaharagon (changes requested):

jonaharagon:
Actually, I disagree with the premise that this could lead to account compromise in the first place. The app only adds a second factor in addition to the existing username+password security. In the unlikely event that the app fails, you should not be _worse off_ than single-factor authentication.

Windows Hello would _replace_ a password I believe, which is not necessarily desirable behavior here.

jonaharagon (approved):
Sorry I double checked later and you're right. These changes LGTM now 👍

PR #2603, opened 2024-06-01
Remove OS index.md page

kimg45:
- remove os index.md page

This page causes the UI to act inconsistently as all the other options with an arrow just act like a regular menu and show more options instead of leading to a page. The page itself also seems to just be duplicating the information that’s already in the menu. The blurb also claims certain OS’s are more “privacy respecting” which I don’t think makes sense in a threat model framework; either it works in your threat model or it doesn’t. This sort of language feels more like PG is picking a threat model for the reader. PG should just present the dry facts about each operating system and let the reader decide which one is best for their threat model. I don’t really think this page adds anything and should just be removed.

<summary>

  </details>


jonaharagon (changes requested):
Thanks for the PR!

It probably makes more sense to add index pages to other sections to make these consistent. That wasn't an option in mkdocs when we originally began using it.

The recommendations on this page are in line with the resources and recommendations on the rest of the site, so changing that would be a much larger change. I'm going to close this because this isn't a good place for this type of discussion, compared to the "Site Development" category on the forum.

jonaharagon:
> It probably makes more sense to add index pages to other sections to make these consistent.

To add to this, the reason we'd want to do so is to make the breadcrumbs more useful, otherwise if you clicked "Operating System" in the first highlighted area in the screenshot below, you would be directed to Android Overview. Functional breadcrumbs has SEO benefits as well.

Also, index pages for subheadings are the only way to add icons to those subheadings. This is why we do:

- Windows (linked to windows overview)
  - Group Policy Settings (linked)

instead of:

- Windows (unlinked header)
  - Overview (linked)
  - Group Policy Settings (linked)

2024-06-02 at 10 52 04

Note this isn't unprecedented because we do it already for the About page at https://www.privacyguides.org/en/about/

PR #2605, opened 2024-06-02
feat: Add threat model labels

kimg45:
Changes proposed in this PR:

- add threat model labels

<summary>

  </details>


redoomed1 (commented):
First of all, I think that adding threat model labels is a great idea! It helps promote the "right tools for the right job" approach that I find far more productive than the absolutist "does _x_ tool(s) make me private/secure" discourse that is unfortunately so common in privacy spaces. In the same vein, these labels can direct people to the resources about threat modeling if they somehow skipped the Knowledge Base.

---
(Apologies in advance for the wall of text)
Having reviewed the deploy preview, here are some suggestions on how you can improve this feature:

1. **I don't think having an icon next to a tool _alone_ is that helpful, since readers don't know what the icon signifies.** What you can do to make the threat model labels more helpful is to embed them with an internal link to the [Common Threats](https://www.privacyguides.org/en/basics/common-threats/) KB page. That way, readers have a reference of what each label signifies.
    - Ideally, on the Common Threats page, **the explainers for each threat would have their own hyperlink that could then be embedded to the threat model labels on the Recommended Tools pages**. That way, when readers hover over a specific label, the instant preview that pops up gives them the respective explainer at a glance. Right now, not every threat explanation on the Common Threats page has its own hyperlinked section, but that can be changed with a simple PR.
2. There are some pages in the deploy preview where, according to the threat model labels, protect against the same threats (e.g., the Password Managers page). **This can lead to unnecessary redundancy, especially when there are a great number of tools on the page (hence why I brought up the example of the PwdMgr page).** What you can do to alleviate this issue is that, for a set of tools that protect against the same types of threats, the labels can be placed next to either the title of the page or right before the first introductory sentence of the page. Then, these labels don't need to be repeated anywhere else on the page.
    - I'm aware of some recommendation pages where one tool protects against an additional threat that other tools on the page don't (e.g., Peergos on the Cloud Storage page). **For such cases, the previous suggestion applies, and there can be an additional sentence or two in the recommendation card that expands on why this tool protects against this additional threat**. This explanation does not need to compare e.g., Peergos to Proton Drive or Tresorit; it could simply highlight its P2P architecture as an advantage against censorship.

These are my notes for how you can improve the implementation of the threat model labels. Hopefully, any part of that wall of text is helpful.

---
Also, here are some technical suggestions related to this PR:
- You might want to rebase your PR with the latest updates in `main`, especially considering that the Android and Productivity Tools pages were recently completely overhauled :sweat_smile: 
- There are some pages in the deploy preview where your labels were not visible for some reason, like the ones on the Tor and VPN pages.

PR #2619, opened 2024-06-17
android: Update F-Droid Basic comparison

redoomed1:
Changes proposed in this PR:

- Update the sentence outlining the advantages of F-Droid Basic over the official F-Droid client
  - Since February of this year, the official F-Droid client also [supports unattended, automatic updates](https://f-droid.org/en/2024/02/01/twif.html). Therefore, highlighting this in the F-Droid Basic section is not really applicable anymore as an advantage over the official client.
  - However, I think that it is worth highlighting that F-Droid Basic targets a higher SDK level, a point that has been brought up on the [forum](https://discuss.privacyguides.net/t/f-droid-client-got-more-modern/12994) before.

<summary>

  </details>

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -430,7 +430,7 @@ That said, the [F-Droid](https://f-droid.org/en/packages) and [IzzyOnDroid](http
 <div class="admonition note" markdown>
 <p class="admonition-title">F-Droid Basic</p>
 
-In some rare cases, the developer of an app will only distribute it through F-Droid ([Gadgetbridge](https://gadgetbridge.org) is one example of this). If you really need an app like that, we recommend using the newer [F-Droid Basic](https://f-droid.org/en/packages/org.fdroid.basic) client instead of the original F-Droid app to obtain it. F-Droid Basic supports automatic background updates without privileged extension or root, and has a reduced feature set (limiting attack surface).
+In some rare cases, the developer of an app will only distribute it through F-Droid ([Gadgetbridge](https://gadgetbridge.org) is one example of this). If you really need an app like that, we recommend using the newer [F-Droid Basic](https://f-droid.org/en/packages/org.fdroid.basic) client instead of the original F-Droid app to obtain it. F-Droid Basic has a reduced feature set (limiting attack surface) and targets a [higher API level](https://developer.android.com/google/play/requirements/target-sdk#why-target) (taking advantage of security improvements brought on by new Android versions).
 
 </div>
 


docs/android.md
```diff
@@ -430,7 +430,7 @@ That said, the [F-Droid](https://f-droid.org/en/packages) and [IzzyOnDroid](http
 <div class="admonition note" markdown>
 <p class="admonition-title">F-Droid Basic</p>
 
-In some rare cases, the developer of an app will only distribute it through F-Droid ([Gadgetbridge](https://gadgetbridge.org) is one example of this). If you really need an app like that, we recommend using the newer [F-Droid Basic](https://f-droid.org/en/packages/org.fdroid.basic) client instead of the original F-Droid app to obtain it. F-Droid Basic supports automatic background updates without privileged extension or root, and has a reduced feature set (limiting attack surface).
```
jonaharagon, line 433: > F-Droid Basic supports automatic background updates without privileged extension or root, and has a reduced feature set (limiting attack surface).

At the end of the day this is still a true statement, does it need to be changed at all?

docs/android.md
```diff
@@ -430,7 +430,7 @@ That said, the [F-Droid](https://f-droid.org/en/packages) and [IzzyOnDroid](http
...
+In some rare cases, the developer of an app will only distribute it through F-Droid ([Gadgetbridge](https://gadgetbridge.org) is one example of this). If you really need an app like that, we recommend using the newer [F-Droid Basic](https://f-droid.org/en/packages/org.fdroid.basic) client instead of the original F-Droid app to obtain it. F-Droid Basic has a reduced feature set (limiting attack surface) and targets a [higher API level](https://developer.android.com/google/play/requirements/target-sdk#why-target) (taking advantage of security improvements brought on by new Android versions).
```
jonaharagon, line 433: I don't think we should talk about minor technical details like API levels that aren't discussed elsewhere on the site. I've found that most of the time when we're linking to external resources like developer docs, it's a pretty good sign that the information we're writing about isn't very useful.

This is why over time we've been _removing_ external links like these as we either decide they're irrelevant, or they're relevant enough to warrant writing about ourselves in a way that's easier to understand than some esoteric docs.

I think that if we think API levels are important for general audiences to think about, that should make its way into an Android overview knowledge base article we link to instead.

However, I also personally don't think the vast majority of readers would or should care about something like this in the first place, so... I wouldn't recommend any changes.

jonaharagon (changes requested):
Thanks for the PR!

I'm not sure this change is necessary though, which I've detailed below. I think the current wording is more in line with what a reader would want to know about F-Droid Basic.

docs/android.md
redoomed1: I suppose it remains true and doesn't need to be changed.

I interpreted the statement as purely a comparative one. If I recall correctly, it was the case that only F-Droid Basic supported automatic background updates at the time this section was written.

PR #2731, opened 2024-08-23
docs: Update PR template

jonaharagon:
Changes proposed in this PR:

- Remove "contribution terms" from PR description template.

This was really only necessary if we ever desired to relicense the site in the future. I was concerned about this back when the site was licensed CC BY-ND. However, I think we should just commit fully to our current license, and eliminate any temptation to keep messing with licenses, we've switched things around enough as-is.

I will note that this would also remove the potential for us to switch to a less restrictive license, like CC BY or even back to CC0 again. It's my understanding that nobody here desires to do so though.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,29 +1,14 @@
-Changes proposed in this PR:
+List of changes proposed in this PR:
 
 -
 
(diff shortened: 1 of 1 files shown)


blacklight447 (approved):
Lgtm

dngray (approved):

dngray (approved):

dngray (approved):

PR #2738, opened 2024-08-27
update: Note beta status of GPG Suite on macOS Sonoma

redoomed1:

- Mention that GPG Suite does not yet have a stable release for macOS Sonoma (still in [beta](https://gpgtools.com/sonoma), with the last blog post update about it published in Dec 2023), which directly affects the Apple Mail (macOS) recommendation
  - Relevant discussion: https://discuss.privacyguides.net/t/apple-mail-no-longer-meets-minimum-criteria/14950
  - This note is intentionally conservative because I'm not sure if there should be an actionable suggestion for this. Would it make sense to advise people on macOS Sonoma to use Thunderbird instead?
- Add the same mention to the GPG Suite listing on the Encryption Software page

### DIFF
--- a/docs/email-clients.md
+++ b/docs/email-clients.md
@@ -92,6 +92,13 @@ These options can be found in :material-menu: → **Settings** → **Privacy & S
 
 </div>
 
+<div class="admonition info" markdown>
+<p class="admonition-title">For those using macOS Sonoma</p>
+
+Currently, GPG Suite does [not yet](https://gpgtools.com/sonoma) have a stable release for macOS Sonoma.
+
+</div>
+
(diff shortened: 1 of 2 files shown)


kimg45:
GPG Suite is also in the Encryption Software so might be good to add the warning there too. Honestly the requirement to support OpenPGP specifically is strange to me. Apple Mail supports E2EE mail through S/MIME.

kimg45 (changes requested):
Add the same message to the Encryption Software section.

kimg45 (approved):

dngray:
> OpenPGP specifically is strange to me. Apple Mail supports E2EE mail through S/MIME.

The main reason I think is S/MIME isn't really used, requires you to pay for a certificate and your provider needs to [trust the CA](https://kb.mailbox.org/en/private/security-privacy-article/s-mime-encryption/#smime-certificates-compatible-with-mailboxorg) chosen.

dngray (approved):

PR #2740, opened 2024-08-29
update: Add self-hosting warnings across multiple pages

redoomed1:

- WIP

### DIFF
--- a/docs/photo-management.md
+++ b/docs/photo-management.md
@@ -11,6 +11,8 @@ cover: photo-management.webp
 
 Most cloud **photo management solutions** like Google Photos, Flickr, and Amazon Photos don't secure your photos against being accessed by the cloud storage provider themselves. These options keep your personal photos private, while allowing you to share them only with family and trusted people.
 
+All the options here can also be self-hosted. However, self-hosting requires more in-depth technical knowledge to deploy safely without compromising your privacy and security. Unless you have adequately researched the risks of self-hosting and strategies to mitigate them, we recommend sticking with the end-to-end encrypted solutions offered by these providers.
+
 ## Ente Photos
 
 <div class="admonition recommendation annotate" markdown>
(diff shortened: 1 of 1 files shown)


docs/photo-management.md
```diff
@@ -41,6 +41,8 @@ Most cloud **photo management solutions** like Google Photos, Flickr, and Amazon
 
 </div>
 
+1. Self-hosting requires more in-depth technical knowledge to utilize safely without compromising your privacy and security. We recommend avoiding self-hosting unless you have adequately researched the risks of self-hosting, as it can leave you worse off than using an E2EE service provider.
```
jonaharagon, line 44: The way I see this is that we have information about self-hosting in a high priority area (the main recommendation card text), and then we are adding a warning to de-prioritize self-hosting in people's minds. I see this as pretty conflicting, which is not a reader experience I like.

If we want to "discourage" self-hosting like this, something we could consider is:

1. Removing "it is also self-hostable" (and this warning) from the recommendation card entirely
2. Adding new text at the bottom of the listing—i.e. in a lower priority placement on the page—that mentions both the option of self-hosting and the warning.

docs/photo-management.md
```diff
@@ -41,6 +41,8 @@ Most cloud **photo management solutions** like Google Photos, Flickr, and Amazon
...
+
 Ente Photos underwent an [audit by Cure53](https://ente.io/blog/cryptography-audit) in March 2023 and by [Fallible](https://ente.io/reports/Fallible-Audit-Report-19-04-2023.pdf) in April 2023.
```
jonaharagon, line 47: Follow up to point #2 in the comment above.

```suggestion

For more technical readers, Ente Photos is also [self-hostable](https://github.com/ente-io/ente/tree/main/server#self-hosting). If you are confident in your ability to host the software after reading their instructions you could consider this option. However, it is best avoided if you are not already comfortable doing so, because self-hosting privacy software improperly can leave you worse off than using an E2EE service provider.

```

jonaharagon (commented):
I generally dislike annotations, and while they have their purpose in some cases, I'm not sure re-using them across many locations on the site to deliver the same warning in the manner suggested here is the best use of them. A different warning style I'd put forth as a suggestion:

redoomed1:
> I'm not sure re-using them across many locations on the site to deliver the same warning in the manner suggested here is the best use of them

Upon reviewing the Photo Management, I agree that annotations aren't the best use for them when the warning applies to all of the tools listed on this page.

redoomed1:
Closing this in favor of a more proper solution that involves a forthcoming Knowledge Base article about the risks of self-hosting

PR #2741, opened 2024-08-29
update: Note that blocking cookies on Thunderbird may affect email logins

redoomed1:

- Add condition for blocking cookies on Thunderbird, as it may prevent people from logging in to Gmail if it's enabled beforehand
  - Relevant discussion: https://discuss.privacyguides.net/t/set-accept-cookies-from-sites-and-set-accept-third-party-cookies-to-never-in-mozilla-thunderbird/20411/10
  - As well as Microsoft365 via SSO according to this reply: https://discuss.privacyguides.net/t/set-accept-cookies-from-sites-and-set-accept-third-party-cookies-to-never-in-mozilla-thunderbird/20411/17

### DIFF
--- a/docs/email-clients.md
+++ b/docs/email-clients.md
@@ -51,6 +51,8 @@ OpenPGP also does not support [forward secrecy](https://en.wikipedia.org/wiki/Fo
 
 #### Recommended Configuration
 
+<div class="annotate" markdown>
+
 We recommend changing some of these settings to make Thunderbird a little more private.
 
 These options can be found in :material-menu: → **Settings** → **Privacy & Security**.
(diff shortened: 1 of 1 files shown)


redoomed1:
> I guess it's fine but depends on whether you use this kind of login method.

Oh I see what you mean. I think this happens if you let Thunderbird go through [Automatic Account Configuration](https://support.mozilla.org/en-US/kb/automatic-account-configuration).

I don't know if this applies to manual configuration. I can test it later and specify the recommendation if needed.

redoomed1:
> I guess it's fine but depends on whether you use this kind of login method.

Ah, that's true, since all of the community members who chimed in to the thread only mentioned it with regard to Gmail. I can change the wording to be more specific on that, rather than a blanket recommendation.

redoomed1:
gmail-exception

jonaharagon (approved):

freddy-m (approved):
LGTM

dngray (approved):

PR #2727, opened 2024-08-22
docs: Expand on developer self-submission requirements

redoomed1:
Changes proposed in this PR:

- Add participation in the developer self-submission process (aka [Project Showcase](https://discuss.privacyguides.net/t/about-the-project-showcase-category/114)) as a requirement
  - It was briefly mentioned in the WIP [volunteer's handbook](https://github.com/privacyguides/privacyguides.org/pull/2700/files#diff-6f3b626aafb1ea347b16df5d199d2969f81a21a5f6d7c220e205d2f9078e69faR53-R59), but I think it's pertinent to mention on the General Criteria page as well.
- Add link to the Common Threats page to provide inspiration for developers when formulating the exact threat model for their projects
  - Given the increased emphasis on threat modeling with the addition of threat model labels on a majority of tool pages, it might be good to nudge prospective developers into thinking about their projects through this framework as well.
- Make a variety of grammar and wording changes

<details>

<summary>Changes that were reverted after feedback</summary>

- Expand a bit on why open source projects are preferred over proprietary alternatives
  - This was mostly spurred by the recent [forum discussion](https://discuss.privacyguides.net/t/what-does-privacy-guides-see-as-opensource/20230) started by @blacklight447.

</details>

---

<summary>

- [x] I agree to the terms listed below:
  <details><summary>Contribution terms (click to expand)</summary>
  1) I am the sole author of this work.
  2) I agree to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform, relicense, and distribute my contribution as part of this project.
  3) I have disclosed any relevant conflicts of interest in my post.
  4) I agree to the Community Code of Conduct.
  </details>

PR #2747, opened 2024-09-02
docs: Transfer security criteria from developer reqs to specific criteria

redoomed1:

This PR sprang from the discussion in https://github.com/privacyguides/privacyguides.org/pull/2727#issuecomment-2306972391 and some discussion in the team chat.

- General Criteria meta page
  - Remove the security whitepaper requirement from the developer self-submission requirements
    - Reason: In the current wording, this req only applies to certain categories of tools: "a project that involves handling of sensitive information like a messenger, password manager, encrypted cloud storage, etc." As such, I don't think it makes much sense to mention the req here. Rather, I think it is better to lay out this requirement in the specific criteria of tool categories where it applies (e.g., the criteria for cloud-based password managers).
  - Remove the line about third-party audit status
    - Reason: Like the whitepaper req, this only applies to certain categories of tools, so it's somewhat misplaced on this page. The developer will have to meet the specific criteria for the category that their tool falls under, anyway.
    - Moreover, requesting a developer for information about whether they have an audit lined up doesn't make much sense to me. The *promise* of a *future* audit shouldn't be taken into account when they are evaluated for inclusion on the site. **Either they have one at the time of evaluation, or they don't.** A developer putting it out to the world that they have requested a third-party audit is tantamount to not having one at the time of evaluation. In such a case, the tool suggestion thread would be marked as #rejected if the category-specific criteria demands an audit, I think.
- Cloud Storage page
  - Add security whitepaper minimum req to criteria to cover the first change in the previous section
- Messengers page
  - Add security whitepaper minimum req to criteria to cover the first change in the General Criteria section
  - Apply specific [guidance about the language for requirements](https://www.privacyguides.org/en/meta/writing-style/#use-must-for-requirements) from the Writing Style page to the criteria
- Make some minor grammar changes

The cloud-based password managers criteria already has a security whitepaper requirement.

### DIFF
--- a/docs/real-time-communication.md
+++ b/docs/real-time-communication.md
@@ -259,11 +259,11 @@ Session has a [whitepaper](https://arxiv.org/pdf/2002.04609.pdf) describing the 
 
 ### Minimum Requirements
 
-- Has open-source clients.
-- Does not require sharing personal identifiers (phone numbers or emails in particular) with contacts.
-- Uses E2EE for private messages by default.
-- Supports E2EE for all messages.
-- Has been independently audited.
+- Must have open-source clients.
+- Must not require sharing personal identifiers (phone numbers or emails in particular) with contacts.
+- Must use E2EE for private messages by default.
+- Must support E2EE for all messages.
+- Must have a security whitepaper as well as a published audit from a reputable, independent third party.
 
 ### Best-Case
 
--- a/docs/real-time-communication.md
+++ b/docs/real-time-communication.md
@@ -270,12 +270,12 @@ Session has a [whitepaper](https://arxiv.org/pdf/2002.04609.pdf) describing the 
 
 Our best-case criteria represents what we would like to see from the perfect project in this category. Our recommendations may not include any or all of this functionality, but those which do may rank higher than others on this page.
 
-- Supports forward secrecy[^1]
-- Supports Future Secrecy (Post-Compromise Security)[^2]
-- Has open-source servers.
-- Decentralized, i.e. [federated or P2P](advanced/communication-network-types.md).
-- Uses E2EE for all messages by default.
-- Supports Linux, macOS, Windows, Android, and iOS.
+- Should support forward secrecy[^1]
+- Should support Future Secrecy (Post-Compromise Security)[^2]
+- Should have open-source servers.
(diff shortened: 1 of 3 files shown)


docs/real-time-communication.md
```diff
@@ -259,22 +259,23 @@ Session has a [whitepaper](https://arxiv.org/pdf/2002.04609.pdf) describing the
...
-- Has open-source clients.
-- Does not require sharing personal identifiers (phone numbers or emails in particular) with contacts.
-- Uses E2EE for private messages by default.
-- Supports E2EE for all messages.
-- Has been independently audited.
+- Must have open-source clients.
+- Must not require sharing personal identifiers (phone numbers or emails in particular) with contacts.
+- Must use E2EE for private messages by default.
+- Must support E2EE for all messages.
+- Must have thoroughly documented encryption and security practices.
```
jonaharagon, line 266: As I've just pointed out at https://github.com/privacyguides/privacyguides.org/pull/2727#discussion_r1749748853 I think it might be intentional that this is only a stated requirement for developer self-submissions, and not as a requirement to be recommended.

The logic being: There may be ways for us to evaluate the security practices of a tool without this documentation, but if a developer is going to ask us to look at their project directly then there is a heightened expectation of work they should be putting in to make their project clear to us.

I am "requesting changes" to block this for now because **if** we do decide we actually want this to be our criteria, then we would have to remove some tools as pointed out in the other PR.

jonaharagon (changes requested):

blacklight447 (approved):

docs/real-time-communication.md
kimg45: Yeah especially since we're requiring a third party audit as well. I think someone like trail of bits is going to be able to do a much deeper analysis of it than we ever could anyway.

redoomed1: > The logic being: There may be ways for us to evaluate the security practices of a tool without this documentation, but if a developer is going to ask us to look at their project directly then there is a heightened expectation of work they should be putting in to make their project clear to us.

Thank you for the explanation! It really clears up why the security whitepaper requirement and 3p audit status is just for developer self-submissions. In that case, it makes sense that they remain on the General Criteria page.

I also understand that the General Criteria should be kept as succinct as possible, so maybe the explanation you wrote here could be something to include in the Volunteer Handbook.

PR #2756, opened 2024-09-12
update: Include more details about FOSS Molly distributions

eylenburg:

- The current description of Molly-FOSS is incorrect. Molly-FOSS still has working push notifications, just using Websockets rather than Google Play. 
- Added some more detail on Molly-UP. There's also the option of using a public Mollysocket instance, so you don't _have_ to self-host. However the only one I'm aware of is Adminforge.de. I added a link to a step-by-step tutorial, which is in German but it's the only easy-to-follow tutorial I know for a migration from Signal to Molly-UP. Feel free to remove it again if you only want English-language links.


jonaharagon (approved):
Hopefully we can find/create a non-German article in the future but LGTM for now

dngray (approved):

PR #2762, opened 2024-09-15
update: Safari description

kimg45:

- Updated description since Safari already blocks tracker by default
- Added fingerprint randomization
- Updated tools description. Content blockers aren't mentioned in the other browsers so I removed that. Also added mention of new features that are comparable to Firefox's anti-fingerprinting features.

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -97,7 +97,7 @@ We are testing a new feature that allows readers to better identify and understa
 
     ---
 
-    We recommend **Safari** alongside a tracker-blocker like [AdGuard](browser-extensions.md#adguard) on iOS, because all iOS browsers are forced to be based on Safari anyways. This reduces the number of parties you're required to trust.
+    We recommend **Safari** due to its [anti-fingerprinting](https://webkit.org/blog/15697/private-browsing-2-0/) features and default tracker blocking. It also separates your cookies in private browsing mode to prevent tracking between tabs.
 
     - [Read Full Review :material-arrow-right-drop-circle:](mobile-browsers.md#safari)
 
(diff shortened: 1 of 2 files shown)


docs/tools.md
```diff
@@ -97,7 +97,7 @@ We are testing a new feature that allows readers to better identify and understa
 
     ---
 
-    We recommend **Safari** alongside a tracker-blocker like [AdGuard](browser-extensions.md#adguard) on iOS, because all iOS browsers are forced to be based on Safari anyways. This reduces the number of parties you're required to trust.
+    We recommend **Safari** due to its [anti-fingerprinting](https://webkit.org/blog/15697/private-browsing-2-0/) features and default tracker blocking. It also separates your cookies in private browsing mode to prevent tracking between tabs.
```
jonaharagon, line 100: We link to AdGuard in this situation because Safari requires a unique solution to content blocking relative to other browsers, because our typical recommendation is unavailable in Safari.

We also should avoid linking to external resources without context like this. Maybe we can write about the relevant anti-fingerprinting features in a section on the mobile browsers page, or we can just remove this link here because we can link to it on the mobile browsers page in a paragraph that provides more context.

jonaharagon (changes requested):
changes in `mobile-browsers.md` lgtm

docs/tools.md
kimg45: Yeah I was just trying to make it look more like the Firefox one where it links to the Mozilla resource on their feature.

jonaharagon: Oh, that's a good point. Hm now I don't know, I kind of don't like the Firefox one either, but I'll see if someone else wants to chime in on this.

blacklight447: Honestly this looks fine as is to me.

blacklight447 (approved):

jonaharagon (approved):

PR #2793, opened 2024-10-29
update!: Remove PrivacyBlur

redoomed1:

- Remove PrivacyBlur from the Data Redaction tool recommendations, as the GitHub repository for the app appears to be deleted, which means that the source code is inaccessible
  - Relevant discussion: https://discuss.privacyguides.net/t/remove-privacyblur/11560/22
  - PrivacyBlur's website, which appears to be hosted on GitHub Pages, is also inaccessible.
  - I'm not sure where to put the admonition that warns about blurring text in images since its fairly out of place on a page where the rest of the tools are metadata scrubbers. There's already guidance on the site about how to [black out information on iOS](https://www.privacyguides.org/en/os/ios-overview/#blacking-out-facesinformation), so maybe the warning can find a place in the Android Overview or General Apps (for Android) page[^1]?

[^1]: Given that PrivacyBlur is only available for Android and iOS


docs/data-redaction.md
```diff
@@ -11,6 +11,15 @@ cover: data-redaction.webp
 
 When sharing files, be sure to remove associated metadata. Image files commonly include [Exif](https://en.wikipedia.org/wiki/Exif) data. Photos sometimes even include GPS coordinates in the file metadata.
 
+</div>
+
```
friadev:
```suggestion
```

friadev: Accidentally added a closing </div>

friadev (changes requested):

friadev (approved):

friadev (approved):

docs/data-redaction.md
```diff
@@ -11,6 +11,13 @@ cover: data-redaction.webp
 
 When sharing files, be sure to remove associated metadata. Image files commonly include [Exif](https://en.wikipedia.org/wiki/Exif) data. Photos sometimes even include GPS coordinates in the file metadata.
 
+<div class="admonition warning" markdown>
+<p class="admonition-title">Warning</p>
+
+You should **never** use blur to redact [text in images](https://bishopfox.com/blog/unredacter-tool-never-pixelation). If you want to redact text in an image, draw a box over the text. For this, we suggest apps like [Pocket Paint](https://github.com/Catrobat/Paintroid).
```
jonaharagon, line 17: I feel like I brought this up on the forum at one point, but Pocket Paint should just be an actual recommendation if we are going to recommend it like this.

jonaharagon (approved):

docs/data-redaction.md
friadev: Good point

redoomed1: I can remove the last sentence recommending it, and we can evaluate Pocket Paint separately in the [existing thread](https://discuss.privacyguides.net/t/paintroid-pocket-paint/16804).

friadev: Good idea I think

PR #2798, opened 2024-11-01
fix: Replace adguard.com link with adguard-dns.io

friadev:

- replaced broken link and updated other link to the proper new link


redoomed1 (approved):

freddy-m (approved):
Good spot

PR #2808, opened 2024-11-12
docs: Update size requirements for company logos

friadev:

- Removed the size recommendation for company logos on the Uploading Images page since they'll just be resized anyway, better to have the original version (after optimization)

### DIFF
--- a/docs/meta/uploading-images.md
+++ b/docs/meta/uploading-images.md
@@ -9,10 +9,7 @@ If you make changes to this website that involve adding new images or replacing 
 
 - We **prefer** SVG images, but if those do not exist we can use PNG images. Additionally, for cover images, we prefer that they are obtained from [Unsplash](https://unsplash.com) and are in the WebP format.
 
-Company logos have canvas size of:
-
-- 128x128px
-- 384x128px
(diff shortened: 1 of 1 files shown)


jonaharagon:
> better to have the original version

There's no difference for SVG files.

Actually we should _add_ a note that PNGs should always be **at least 200x200**, because logo images display at 100x100 and we want them looking good on 2x retina displays.

docs/meta/uploading-images.md
```diff
@@ -9,10 +9,7 @@ If you make changes to this website that involve adding new images or replacing
 
 - We **prefer** SVG images, but if those do not exist we can use PNG images. Additionally, for cover images, we prefer that they are obtained from [Unsplash](https://unsplash.com) and are in the WebP format.
 
-Company logos have canvas size of:
-
-- 128x128px
-- 384x128px
+Company logos that are PNGs should be at least 200x200px.
```
jonaharagon, line 12:
```suggestion
Company logos should be square if possible, and at least 200x200px if they are PNGs (non-vector images).
```
I'd also note this

jonaharagon (approved):

PR #2525, opened 2024-04-14
update!: Add AI chat recommendations

I-I-IT:
Need feedback on criteria, including reasonable rate limit. For the header, I will try to find how to upload the assets.

Feel free to add how to run AI models locally, with LM studio for example.

Changes proposed in this PR:

-

- [X] I have disclosed any relevant conflicts of interest in my post.
- [X] I agree to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform, relicense, and distribute my contribution as part of this project.
- [X] I am the sole author of this work. 
- [X] I agree to the [Community Code of Conduct](https://www.privacyguides.org/coc).

PR #2811, opened 2024-11-12
update!: Add AI provider recommendations

jonaharagon:
Continuing work from: https://github.com/privacyguides/privacyguides.org/pull/2525#issuecomment-2469636493


jonaharagon:
As I mentioned elsewhere I would prefer to not make these changes. We can use this PR as a reference in the future if someone else wants to take up the discussion and open another PR, otherwise I think we are fine with AI as-is for now.

PR #2815, opened 2024-11-15
update: Remove outdated master password info from Proton Pass entry

tedgravlin:
* Removes information that states Proton Pass cannot have a password different from the rest of your Proton account.

This PR removes text from the Proton Pass entry that states a separate password cannot be set for Proton Pass. 
Proton added an "Extra password" option to Proton Pass sometime around July 2024:
https://proton.me/blog/pass-secure-link-sharing

> [...]in response to feedback from the community, we’re also launching an Extra Password feature alongside Secure Links. This optional feature lets you create a second password to access credentials stored in Pass for greater compartmentalization and peace of mind. Extra Password is available to the entire Proton Pass community, including those with Free plans.

Screenshot:
image

Thanks!


jonaharagon (approved):
Good catch thanks!

redoomed1 (approved):

PR #2825, opened 2024-11-30
update!: Remove Metapho and update Removing Metadata blog post

redoomed1:

- Data and Metadata Redaction page
  - Remove Metapho from the recommendations due to it requiring a subscription or one-time purchase (see its [detailed description](https://apps.apple.com/app/id914457352) on the App Store)
    - Relevant discussion: https://discuss.privacyguides.net/t/remove-metapho-ios-exif-editor-not-free-doesnt-meet-criteria/22821
  - Mention the use of metadata removal shortcuts as a solution on iOS, linking to the relevant blog post

## To Do
- [ ] Update the instructions for creating metadata removal shortcuts on iOS
  - Relevant discussion: https://discuss.privacyguides.net/t/exif-removal-tools-ios-no-apps-required/16664


docs/data-redaction.md
```diff
@@ -48,6 +48,13 @@ On Linux, a third-party graphical tool [Metadata Cleaner](https://gitlab.com/rmn
 
 ## Mobile
 
+<div class="admonition note" markdown>
+<p class="admonition-title">Note</p>
+
+For those on iOS, you can remove photo and video metadata without using a third-party app by creating [Shortcuts](/articles/2022/04/09/integrating-metadata-removal) for both.
```
jonaharagon, line 54: I would prefer to have this written as a standalone "Shortcuts (iOS)" section rather than an admonition.

It wouldn't have to use a recommendation *card* for the Shortcuts app itself, see https://www.privacyguides.org/en/data-broker-removals/#manual-opt-outs-free for an example of what I'd call an "instructional recommendation"

jonaharagon: Also I think links to blog articles should be absolute URLs and not relative URLs to avoid issues with the build system, because the blog is a "separate website" technically.

jonaharagon (changes requested):

jonaharagon:
I wonder if we are able to make a Shortcut that can be shared as a link for people to download too. I'll try it out this weekend.

docs/data-redaction.md
redoomed1: > I would prefer to have this written as a standalone "Shortcuts (iOS)" section

> It wouldn't have to use a recommendation card for the Shortcuts app itself

Yeah, on further reflection, I think that the instructional recommendation for Shortcuts should just be on this page, rather than direct readers to the related blog post.

Looking at the updated instructions in the second forum post above, I think that the new screenshots (if we decide to include any for demonstration purposes) should just be added to this page rather than replacing old screenshots on the two-year-old blog post.

redoomed1: In which case, I'll close this PR and open a new one without the `ci:build blog` label if we decide this is the most optimal way to move forward with these changes.

docs/data-redaction.md
```diff
@@ -89,25 +89,15 @@ The app offers multiple ways to erase metadata from images. Namely:
 - It allows you to drag photos from another app into ExifEraser when they are both open in split-screen mode.
 - Lastly, it allows you to paste an image from your clipboard.
 
-### Metapho (iOS)
+### Shortcuts (iOS)
```
friadev, line 92:
```suggestion
### Shortcuts (iOS, macOS)
```
Could mention that this works on macOS as well.

docs/data-redaction.md
```diff
@@ -89,25 +89,15 @@ The app offers multiple ways to erase metadata from images. Namely:
...
 
-<div class="admonition recommendation" markdown>
-
-![Metapho logo](assets/img/data-redaction/metapho.jpg){ align=right }
-
-**Metapho** is a simple and clean viewer for photo metadata such as date, file name, size, camera model, shutter speed, and location.
+On iOS, you can remove image metadata without using any third-party apps by creating a [**shortcut**](https://apps.apple.com/app/id915249334) for this purpose. Here is an example shortcut you can download to use as is:
```
friadev, line 94:
```suggestion
On iOS and macOS, you can remove image metadata without using any third-party apps by creating a [**shortcut**](https://apps.apple.com/app/id915249334) for this purpose. Here is an example shortcut you can download to use as is:
```

friadev (changes requested):

friadev:
I think we could also now tighten the restrictions to be open source apps only. I don't think shortcuts counts since it's built-in functionality but we could specify third party apps.

docs/data-redaction.md
```diff
@@ -89,42 +85,30 @@ The app offers multiple ways to erase metadata from images. Namely:
...
-- [:simple-appstore: App Store](https://apps.apple.com/app/id914457352)
+[:material-tag-minus: Clean Image Metadata](https://icloud.com/shortcuts/fb774ddb7b5b4296871776c67ac0fff9){ .md-button }
 
-</details>
-
-</div>
+You can also use it as a model for your own shortcut; just make sure that the **Preserve Metadata** option under the **Convert** action is unchecked. Once added, the shortcut appears in the share sheet that appears when you select the :octicons-share-24: Share button. You can select multiple images to remove their metadata.
 
-## Command-line
+This shortcut removes metadata such as location, device model, lens model, location, and other camera information. It also sets the image creation date to the time the shortcut was used.
```
friadev, line 96:
```suggestion
This shortcut removes metadata such as location, device model, lens model, and other camera information. It also sets the image creation date to the time the shortcut was used.
```
Location is on here twice.

friadev (changes requested):

friadev (approved):

friadev (approved):

dngray (approved):

PR #2833, opened 2024-12-13
update: Add info about the new clean up feature in ios

friadev:

- added info about the new [Clean Up](https://support.apple.com/en-us/121429) feature in iOS that can automatically blur faces and redact information

### DIFF
--- a/docs/os/ios-overview.md
+++ b/docs/os/ios-overview.md
@@ -250,11 +250,20 @@ In addition to locking apps behind biometrics, you can also hide apps so that th
 
 You can hide an app by long-pressing on it and selecting **Require Face ID/Touch ID** → **Hide and Require Face ID/Touch ID**. Note that pre-installed Apple apps, as well as the default web browser and email app, cannot be hidden. Hidden apps reside in a **Hidden** folder at the bottom of the App Library, which can be unlocked using biometrics. This folder appears in the App Library whether you hid any apps or not, which provides you a degree of plausible deniability.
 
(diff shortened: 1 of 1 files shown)


dngray (approved):

dngray (approved):

docs/os/ios-overview.md
```diff
@@ -254,6 +254,22 @@ You can hide an app by long-pressing on it and selecting **Require Face ID/Touch
...
+<div class="admonition warning" markdown>
+<p class="admonition-title">Warning</p>
+
+Don't use Clean Up to remove text, block text out with a fully black box instead.
+
+</div>
+
+Otherwise, you can add a shape to fully block out a face or information.
+
 - Open the **Photos** app and tap the photo you have selected for redaction
```
redoomed1, line 273:
```suggestion
Our warning [against blurring text](../data-redaction.md) also applies here, so we recommend to instead add a black shape with 100% opacity over it. In addition to redacting text, you can also fully black out any face or object using the **Photos** app.

- Tap the image you have selected for redaction
```
This suggestion attempts to remove this admonition while retaining the same information. It also clarifies what a "fully black box" means.

redoomed1:
I also suggest renaming this section of the page to something like "Redacting Elements in Images" since the Clean Up feature now adds a redaction option that is not limited to blacking out elements.

redoomed1 (approved):
LGTM

jonaharagon (approved):
Did we want to add an example image to this article?

PR #2835, opened 2024-12-14
update: Add note about Mull open in private tab option

I-I-IT:
Wording is bad and can be improved.

First discussed here https://discuss.privacyguides.net/t/usefull-mull-feature-is-not-recommended/23164

If Cromite has this feature too then it could be added.

-Add info about Mull "Open links in private tabs" which when you use apps and their 'built-in' browser will open in private tabs instead of standard.


friadev:
#2839 Mull is ending support so I will be closing this PR

PR #2119, opened 2023-04-08
update!: Browser fingerprinting overview

jonaharagon:
Changes proposed in this PR:

- Add fingerprinting overview page

**Please do not add any suggestions or reviews correcting grammar or wording until this PR is undrafted, it will only waste time.** This draft PR is only open to accept suggestions about what _information_ needs to be incorporated on this page.

- [x] I have disclosed any relevant conflicts of interest in my post.
- [x] I agree to grant Privacy Guides a perpetual, worldwide, non-exclusive, transferable, royalty-free, irrevocable license with the right to sublicense such rights through multiple tiers of sublicensees, to reproduce, modify, display, perform, relicense, and distribute my contribution as part of this project.
- [x] I am the sole author of this work. 
- [x] I agree to the [Community Code of Conduct](https://www.privacyguides.org/en/code_of_conduct/).

PR #2919, opened 2025-03-04
update: Add warning about canvas fingerprint issues

akosnikhazy:
A warning about Canvas fingerprint protection breaks images and makes some site harder to use.

-

### DIFF
--- a/docs/desktop-browsers.md
+++ b/docs/desktop-browsers.md
@@ -64,6 +64,14 @@ These are our currently recommended **desktop web browsers** and configurations 
 
 If you need to browse the internet anonymously, you should use [Tor](tor.md) instead. We make some configuration recommendations on this page, but all browsers other than Tor Browser will be traceable by *somebody* in some manner or another.
 
+<div class="admonition warning" markdown>
+<p class="admonition-title">Warning about Canvas Fingerprinting</p>
+
+The Canvas API in HTML5, designed for drawing graphics via JavaScript and HTML, can also be used for fingerprinting. Browsers that protect against this will break images you upload and make some sites more difficult to use. An example is Proton Drive, where thumbnails generated for uploaded images are broken because the site generates thumbnails on the user's computer using the Canvas API.
+
+</div>
+
+
 ## Mullvad Browser
 
 <div class="admonition recommendation" markdown>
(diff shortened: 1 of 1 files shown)


docs/desktop-browsers.md
```diff
@@ -64,6 +64,14 @@ These are our currently recommended **desktop web browsers** and configurations
...
 If you need to browse the internet anonymously, you should use [Tor](tor.md) instead. We make some configuration recommendations on this page, but all browsers other than Tor Browser will be traceable by *somebody* in some manner or another.
 
+<div class="admonition warning" markdown>
+<p class="admonition-title">Warning about Canvas Fingerprinting</p>
+
+The Canvas API in HTML5, designed for drawing graphics via JavaScript and HTML, can also be used for fingerprinting. Browsers that protect against this will break images you upload and make some sites more difficult to use. An example is Proton Drive, where thumbnails generated for uploaded images are broken because the site generates thumbnails on the user's computer using the Canvas API.
+
+</div>
+
+
```
jonaharagon, line 74:
```suggestion

```

jonaharagon (changes requested):
Is there a forum discussion related to this PR already? I might have to look further into this. I'm not sure this is the proper location for this warning in any case...

docs/desktop-browsers.md
akosnikhazy: Found this topic only before making the change. https://discuss.privacyguides.net/t/cant-upload-pictures-with-anti-fingerprinting-enabled/19043

There is a good page for Threat Modeling but there are no place to put information like this. The closest would be "How much trouble am I willing to go through to try to prevent potential consequences?" part but that page is not about examples but a way of thinking. 

A lot of things that one blocks for privacy could break something they want to use. It might worth a page on its own. But for now I do not know where else this could be put.

redoomed1: Perhaps a better location for this warning would be the pending [overview on browser fingerprinting](https://github.com/privacyguides/privacyguides.org/pull/2119)? :)

akosnikhazy:
Sounds like a perfect place

friadev:
Firefox has [Fingerprinting Protection (FPP)](https://support.mozilla.org/en-US/kb/firefox-protection-against-fingerprinting) available through the settings which does randomization of image data without disrupting the image visibly. It might be better just to not recommend Arkenfox and have people use Mullvad Browser or Tor browser if they want really good anti fingerprinting.

jonaharagon:
Thanks for the PR. We may talk about this in a page like #2119 but I don't think this warning belongs here, and it isn't widely applicable to our browsers by default as @friadev mentioned.

PR #2921, opened 2025-03-04
fix: Revert #2790

jonaharagon:

- The bug addressed by #2790 appears to have been fixed upstream.


jonaharagon:
On a second look, it seems like we changed translated builds to not download a full git history anyways, so the modification time would no longer be accurate if we used this plugin again anyways.

PR #2930, opened 2025-03-06
TV & Projector Recommendation

GorujoCY:

- Added TV & Projectors recommendations

as suggested by https://discuss.privacyguides.net/t/television-and-projector-guide/23745/.
I wasn't sure where or if a guide would fit so I made a recommendation instead, not really a guide but if it can be expanded to be a guide, great. But for now this is the proposal. I used the [router.md](https://github.com/privacyguides/privacyguides.org/blob/main/docs/router.md) file as a template

I've decided to make my first PR based on it and what I thought would be good recommendations (conflict of interest is called so I recommend people pitch in please), as always feel free to correct me on things and it's genuinely my first contribution so I tried what I can. I am okay with letting me know what should change and what not  Or let me know if it's inappropriate etc. Also I created a proposed branch if necessary dont merge it into main but I simply couldn't choose so until a branch can be created it will start in main.


friadev (changes requested):
Needs to be approved on the forum first.

jonaharagon (changes requested):
This should probably be split into two pages, a "Media Player" page (this could also include non-home-theater tools like VLC in addition to Kodi perhaps) that goes under Software, and a "Streaming Devices" page that goes under Hardware.

Remember to add new pages to the navigation: https://github.com/privacyguides/privacyguides.org/blob/5b9b7d9eab2525f55532c2541338bc2947a54058/mkdocs.yml#L417-L437

jonaharagon:
The more I'm thinking about it, the more I think that not only should this be two separate pages, but each of those pages should be a separate PR. I am going to close this PR, do you mind opening one PR adding a software media players page and a second PR adding a streaming/media devices hardware page, and copying your work from this one to those two?

If you need help with doing this just ping me and I'll set it up, I don't want to lose the work you've put in so far by closing this :)

jonaharagon:
> conflict of interest is called so I recommend people pitch in please

also you need to share what your conflict of interest actually is

GorujoCY:
Hi @jonaharagon 

First of all thanks for letting me know where to place the md file(s) appreciated for the navigation and I was gonna modify the tools.md for it (and after it would be approved) but now I know where the navigation goes.
Second, for conflict of interest I may have misunderstood it for sure and that happens, I am not a native on English after all but I thought I'd call it as these recommendations were based on my own and not necessarily PG's organization or something yhr community can agree and same with the criteria for the PG Organization, in case it also doesn't meet your standards.
Third, Segregating them would in fact be a good idea I don't disagree at all. But I do want to state my reasons why I combined them. First, the goal here is to replace the smart tv to something on a more long term basis without needing to Change this page every time. second: I feel that if we're gonna recommend things like Monitors on the size of tvs (or any the consumer chooses), we should also provide the tools PG recommends with thier purchase of the monitor to turn it into a Television. But since you hold executive position please do let me know if you still want to proceed with segregating them. I just wanted to state my reasonings behind it
Also for you to suggest changes I presume it gets an approval? Because as @friadev said it needs to be approved on the forums first.

Thank you gor understanding

jonaharagon:
I think it still should be separated, yes.

You can make a PR without it being approved, it will just take longer to merge. Like anything tagged as [approved](https://discuss.privacyguides.net/tag/approved) can usually be merged within a few days of a PR being made, but if it isn't pre-approved it could take weeks or longer. You can still do it though 👍

PR #2932, opened 2025-03-08
update: Disable Windows Recall in gpedit guide

jonaharagon:
These instructions are based on these and I haven't tested them myself (not on Windows): https://docs.security.tamu.edu/docs/endpoint-security/Policies/MSRecall/

I am going to keep this marked as a draft until someone with Windows 11 confirms it's accurate, or I fire up a Windows computer somewhere and check myself. If someone can let me know that'd be great :)


dngray (approved):

docs/os/windows/group-policies.md
```diff
@@ -127,6 +127,17 @@ This last setting disables OneDrive on your system; make sure to change it to **
 
 - Improve inking and typing recognition: **Disabled**
 
+#### Windows AI
+
+<div class="admonition info" markdown>
+<p class="admonition-title">Windows Recall</p>
+
+Windows 11 recently introduced a feature called **Recall**, which records all your activity and creates a searchable archive of that activity history. This is a massive privacy vulnerability, because those archives can potentially store highly sensitive information (essentially anything displayed on your screen), and can be trivially accessed by local administrators or malicious actors with user-level access to your device.
```
friadev, line 135: [Recall](https://support.microsoft.com/en-us/windows/privacy-and-control-over-your-recall-experience-d404f672-7647-41e5-886c-a3c59680af15) doesn't let other users access your info and it's stored encrypted, protected by the TPM and biometrics. It's definitely not true that it can be trivially bypassed. It also by default tries to exclude sensitive data like passwords and credit cards, and you can tell it to exclude certain sensitive apps like a web browser or messenger. I think info on how to disable it is fine but maybe some more info on the different settings and how to configure it might be useful as well.

friadev (changes requested):

jonaharagon:
@ph00lt0 what isn't finished?

@friadev this is the only available setting related to Recall in gpedit as far as I can tell, so there are no additional configuration settings to talk about in this particular guide.

docs/os/windows/group-policies.md
jonaharagon: this is the only available setting related to Recall in gpedit as far as I can tell, so there are no additional configuration settings to talk about in this particular guide.

friadev: >Recall doesn't share snapshots with other users that are signed into Windows on the same device and IT admins can't access or view the snapshots on end-user devices.

I think we shouldn't say it can be accessed by admins or users since the documentation explicitly says that's not possible.

friadev:
```suggestion
Windows 11 recently introduced a feature called **Recall**, which records all your activity and creates a searchable archive of that activity history. This is a massive privacy vulnerability, because those archives can potentially store highly sensitive information (essentially anything displayed on your screen).
```

friadev (changes requested):

docs/os/windows/group-policies.md
friadev: In the official [documentation](https://learn.microsoft.com/en-us/windows/client-management/manage-recall#:~:text=Recall%20doesn't%20share%20snapshots%20with%20other%20users%20that%20are%20signed%20into%20Windows%20on%20the%20same%20device%20and%20IT%20admins%20can't%20access%20or%20view%20the%20snapshots%20on%20end%2Duser%20devices.), it says admins and other users can't access your recall data, so I think we should delete that part.

dngray: That's local administrators and it also includes corporate domain administrators (eg if it is part of Active Directory (AD), though in that case it is Recall is disabled by default. The document you linked refers to network administrators assuming the device is still yours.

Though it does say the data is encrypted.

friadev: I don't have a system that supports Recall can someone confirm if that's true. It does say later in the same paragraph that other users can't decrypt it.

docs/os/windows/group-policies.md
```diff
@@ -127,6 +127,17 @@ This last setting disables OneDrive on your system; make sure to change it to **
...
+
+</div>
+
+- Turn off saving snapshots of Windows: **Enabled**
```
friadev:
```suggestion
- Turn off saving snapshots for use with Recall: **Enabled**
```

friadev (changes requested):

docs/os/windows/group-policies.md
```diff
@@ -127,6 +127,17 @@ This last setting disables OneDrive on your system; make sure to change it to **
 
 - Improve inking and typing recognition: **Disabled**
 
+#### Windows AI
+
+<div class="admonition info" markdown>
+<p class="admonition-title">Windows Recall</p>
+
+Windows 11 recently introduced a feature called **Recall**, which records all your activity and creates a searchable archive of that activity history. This is a massive privacy vulnerability, because those archives can potentially store highly sensitive information (essentially anything displayed on your screen).
```
jonaharagon:
```suggestion
Windows 11 recently introduced a feature called **Recall**, which records all your activity and creates a searchable archive of that activity history. This is a massive privacy vulnerability, because those archives can potentially store highly sensitive information (essentially anything displayed on your screen), and malware without any special permissions or admin access can trivially [access](https://github.com/xaitax/TotalRecall) the full Recall database anytime after it is unlocked.
```

friadev (approved):

PR #2933, opened 2025-03-08
docs: Add team members to .mailmap

jonaharagon:
Team members should use their canonical email addresses rather than noreply.github addresses now that we have a new email setup.

This way we can get accurate statistics when running `git shortlog -nse --group=author --group=trailer:co-authored-by --no-merges` and `git shortlog -nse --group=trailer:signed-off-by`

only spam for the brand new @privacyguides.org accounts of course, the other emails in this file are already public in the git log :P

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,16 +1,22 @@
+# Please use this file to correct information in our git log.
+# Format: Proper Name <Proper Email> <Incorrect Email>
+#
+# Privacy Guides Team Members
+# ===========================
 Daniel Gray <dngray@privacyguides.org> <dng@disroot.org>
 Daniel Gray <dngray@privacyguides.org> <48640805+dngray@users.noreply.github.com>
 Daniel Gray <dngray@privacyguides.org> <dngray@privacytools.io>
 Daniel Gray <dngray@privacyguides.org>
+Em <em@privacyguides.org> <194856901+EmAtPrivacyGuides@users.noreply.github.com>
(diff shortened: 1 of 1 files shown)


.mailmap
```diff
@@ -1,16 +1,48 @@
...
-rollsicecream <rollsicecream@proton.me> <153316540+rollsicecream@users.noreply.github.com>
-rollsicecream <rollsicecream@proton.me> <waterfallnet@proton.me>
-Freddy <freddy@privacyguides.org> <freddy@decypher.pw>
-Freddy <freddy@privacyguides.org> <freddy@privacytools.io>
+mfwmyfacewhen <mfw@privacyguides.org> <94880365+mfwmyfacewhen@users.noreply.github.com>
 Niek de Wilde <niek@privacyguides.org> <github.ef27z@simplelogin.com>
 Niek de Wilde <niek@privacyguides.org> <blacklight447@privacytools.io>
+redoomed1 <redoomed1@privacyguides.org> <161974310+redoomed1@users.noreply.github.com>
+#
+# Misc. Duplicate Contributors
```
dngray, line 26: Do we really want to be adding all these, imo these people haven't had anything to do with Privacy Guides, or rather formerly Privacy Tools in years.

I doubt anyone would need to contact them about anything, likely that content doesn't exist anymore.

jonaharagon: No, I'll just add them to my global ~/.mailmap then

dngray (approved):

PR #2934, opened 2025-03-08
update: Note iOS limitations wrt Onion Browser

### LINKED ISSUE #2929: Onion Browser needs a more explicit warning about leaks
### Affected page

https://www.privacyguides.org/en/tor/#onion-browser-ios

### Description

Current wording is
> does not provide the same levels of privacy protections

needs to be adjusted to be more clear about the inherent leaks due to limitations of iOS.

The latest available (3.1.5) also has an issue where data is not cleared on close, although this one will be fixed in the next update (3.2.0). Presumably currently stuck waiting for approval from Apple.

### Sources

https://github.com/OnionBrowser/OnionBrowser/issues/509
https://github.com/OnionBrowser/OnionBrowser/issues/508

### Before submitting

jonaharagon:

- Closes #2929

### DIFF
--- a/docs/tor.md
+++ b/docs/tor.md
@@ -121,6 +121,8 @@ All versions are signed using the same signature, so they should be compatible w
 
 </div>
 
+On iOS, Orbot has some limitations that could potentially cause crashes or leaks: iOS does not have an effective OS-level feature to block connections without a VPN like Android does, and iOS has an artificial memory limit for network extensions that makes it challenging to run Tor in Orbot without crashes. Currently, it is always safer to use Tor on a desktop computer compared to a mobile device.
+
 ## Onion Browser (iOS)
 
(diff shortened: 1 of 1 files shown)


SkewedZeppelin:
these are additional separate issues of Orbot, not of the built in Tor feature

jonaharagon:
I thought it was both. Are there _not_ issues when you use the built-in version?

SkewedZeppelin:
Orbot is hinted that it is on its way out due to the memory limitations and can OOM due to iOS restrictions.

OnionBrowser with built-in Tor, which is hinted that it is the future default, will always leak when any audio/video is embedded or WebRTC is used.

jonaharagon:
Right I see what you mean, we can distinguish between these two cases.

SkewedZeppelin (approved):

PR #2938, opened 2025-03-08
update!: Home Theater (Software)

GorujoCY:

- Added Home Theater (Software)
  - Relevant discussion(s): see the linked forum topics [here](https://github.com/privacyguides/privacyguides.org/pull/2938#issuecomment-2730314567)

Follow up of what Jonah (@jonaharagon ) said in [PR #2930](https://github.com/privacyguides/privacyguides.org/pull/2930#issuecomment-2707891568) which was supposed to be a contribution for alternatives towards smart TVs, I'm doing what he asked which is to separate them into 2 pages and 2 separate PRs of those with the names that was suggested in the old PR. And we're beginning with the media players and will work on the hardware side later!

As always let me know if you spot any mistakes or any changes to be made or want to pitch in new recommendations and more importantly any objections thank you. For more context I do recommend looking at that past PR.


GorujoCY:
@jonaharagon could the build configuration be something wrong or something else, could you check it? If it fails like that something is up.

friadev (changes requested):
Needs to be approved on the forum.

GorujoCY:
Cancel, feel free to deny the re-request I got confused for a moment.

friadev:
Yeah I'm not happy with VLC. I think we should expect some baseline level of basic security when recommending third party apps, and here VLC doesn't offer an official flatpak and the one that is available is basically not sandboxed at all: full filesystem access etc. On macOS, VLC does not use the App Sandbox and has entitlements for JIT and microphone and camera access. Also it's not on the App Store which means no nice autoupdates from there. I really don't want to recommend this when the default media player is better.

GorujoCY:
> Yeah I'm not happy with VLC. I think we should expect some baseline level of basic security when recommending third party apps, and here VLC doesn't offer an official flatpak and the one that is available is basically not sandboxed at all: full filesystem access etc. On macOS, VLC does not use the App Sandbox and has entitlements for JIT and microphone and camera access. Also it's not on the App Store which means no nice autoupdates from there. I really don't want to recommend this when the default media player is better.

Anyone else's thought on this? Will consider removing if agreed upon.
And honestly idk how better kodi is anywho so if that's also the case can we even recommend media players? :thinking:

GorujoCY:
@friadev VLC is on the app store
https://apps.apple.com/us/app/vlc-media-player/id650377962

friadev:
> @friadev VLC is on the app store https://apps.apple.com/us/app/vlc-media-player/id650377962
> 
> In conclusion I am 50/50 on this.

If you want the macOS version you can't download off the app store, and the version you get from their website doesn't enable the App Sandbox. Also has some weird entitlements like allowing JIT for some reason.

friadev:
>And honestly idk how better kodi is anywho so if that's also the case can we even recommend media players? 🤔

Kodi at least offers an official flatpak. Looking at their macOS app, they have fewer entitlements like enabling JIT and camera/mic, but they still don't enable the App Sandbox and they disable library validation which is a security feature: https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.cs.disable-library-validation
Their flatpak also has a warning about it being potentially unsafe but I'm not familiar with flatpak.

jonaharagon:
I would prefer somebody open a VLC thread on the forum so we can get more eyes on this than just whoever is looking at PRs. 

My 2 cents though:

I think we should consider the context that people are using these media players in. If it's largely trusted content, like media people have ripped themselves, then using tools like this does seem like an improvement to me. 

This could actually be a great opportunity to address security issues with media on the site. I think @friadev had a good point that we've seen a lot of vulnerabilities directly caused by esoteric media engines in other apps like Chrome and iMessage. If we recommended tools like VLC we could add a section about these dangers and a warning about running internet content. 

I think that this approach would be more educational and useful to readers than simply avoiding the topic of media players on the site.

friadev:
>I would prefer somebody open a VLC thread on the forum so we can get more eyes on this than just whoever is looking at PRs.

Yeah I suggested that. I could but I don't actually want it to be added so it would be me making the suggestion and then immediately shutting it down which makes no sense.

GorujoCY:
it's been a wekk so about time we summarize and decide:

1. A bonus round was added and it's jellyfin. It has been overall praised by the community, not surprising I guess and it's at 9 votes
https://discuss.privacyguides.net/t/jellyfin-media-management/25867

2. Kodi, despite the 5 votes, Kodi has ended up in a mixed bag, mostly concerning insecurities or otherwise many users not regarding it as a good option even for that, I completely disagree and it seems jonah too as it is a far better altenative for the big screen imo or in our opinion..
Unfortunately Kodi has left us still needing to debate
https://discuss.privacyguides.net/t/kodi-home-theater-software/25866

3. VLC, VLC seems like it will be down, most have disregarded it with very few pointing it's uses which I think for sure it does, also only 4 votes
https://discuss.privacyguides.net/t/vlc-media-player-software/25865
Maybe therefore it is agreeable we remove VLC?

let's see for the following CVEs of each of them also:

- last VLC CVE:  [CVE2024-46461](https://nvd.nist.gov/vuln/detail/CVE-2024-46461),
- Kodi stands at [CVE-2023-30207](https://nvd.nist.gov/vuln/detail/CVE-2023-30207) [So I feel they've been able to secure their software lately which is good]
- Despite the praise Jellyfin is getting, the last cve is [CVE-2024-43801](https://nvd.nist.gov/vuln/detail/cve-2024-43801)

that's all on report based on what I gathered and experience. What do you think is our next move?

jonaharagon:
**Alternatively**, we could wait for #2970 to be merged, add Jellyfin to _that_ page, and not list Kodi for now and not create a new media players/home theater category at all. Might make more sense now that I'm thinking about it.

GorujoCY:
will be waiting for the that PR to merge first and decide

dngray:
> **Alternatively**, we could wait for #2970 to be merged, add Jellyfin to _that_ page, and not list Kodi for now and not create a new media players/home theater category at all. Might make more sense now that I'm thinking about it.

That's now been done, so we could look at this.

I think in general media players don't make a whole lot of sense, although VLC probably does on windows, over say windows media player which no doubt has some AI abilities, because why not. If we're looking at Linux well, there's mpv as well. so I do agree in keeping this page about home theater options..

In terms of privacy this page does make more sense in terms of a minipc or single board computer running linux with something like Jellyfin or Kodi for privacy reasons over say smarttv firmware which probably sends all sorts of data back to the manufacturer (eg samsung).

I do think we should avoid making hardware recommendations like the Raspberry Pi however, as there's nothing particularly privacy friendly about that and for this purpose it may not even be the most cost effective idea.

GorujoCY:
> > **Alternatively**, we could wait for #2970 to be merged, add Jellyfin to _that_ page, and not list Kodi for now and not create a new media players/home theater category at all. Might make more sense now that I'm thinking about it.
> 
> That's now been done, so we could look at this.
> 
> I think in general media players don't make a whole lot of sense, although VLC probably does on windows, over say windows media player which no doubt has some AI abilities, because why not. If we're looking at Linux well, there's mpv as well. so I do agree in keeping this page about home theater options..
> 
> In terms of privacy this page does make more sense in terms of a minipc or single board computer running linux with something like Jellyfin or Kodi for privacy reasons over say smarttv firmware which probably sends all sorts of data back to the manufacturer (eg samsung).
> 
> I do think we should avoid making hardware recommendations like the Raspberry Pi however, as there's nothing particularly privacy friendly about that and for this purpose it may not even be the most cost effective idea.

Oohoo never expected the day to come back.

Anyways few things:

1. We likely need more discussions in regards so I can try bumping the raspberry pi thread if accepted but the only reason we recommend is for the [TV HAT](https://www.raspberrypi.com/products/raspberry-pi-tv-hat/) which should make for a relatively painless setup. Kinda like how Valve is building a hardware ecosystem with SteamOS. Now of course alongside Raspberry Pi we can say any PC as another option for home theater not just the raspberry pi but meant for those a bit/little more advanced in the technical skills.

2. If not could we add a recommendation to my community wiki for the hardware part (https://discuss.privacyguides.net/t/harm-reducing-replacing-your-smart-tv/27392/)?

3. We can also bump the VLC Discussion if needed but considering the past discussion, I think to align with the community values we shou Probably move forward from not recommending, otherwise we bump and see.

ph00lt0:
There seems no activity here and the PR contains commented out code

PR #2827, opened 2024-12-02
update!: Add map/navigation recommendations

eylenburg:
List of changes proposed in this PR:

Added another page recommending privacy-respecting maps apps.

The title photo is from Unsplash. Logos were optimized with scour and optipng.

Also I removed the F-Droid link for Organic Maps as it is not an external repo provided by the developers, however they link to F-Droid on their website so I'm not completely sure if this kind of endorsement means we should still add the F-Droid link.

PR #2947, opened 2025-03-11
update!: add Magic Earth recommendation

eylenburg:
This is the PR for adding Magic Earth once/if this is agreed. I had previously included it in the original Maps PR (https://github.com/privacyguides/privacyguides.org/pull/2827) but this was later reduced to only recommending Organic & Osmand, as further discussion is required before potentially approving Magic Earth.


freddy-m:
I'm going to close this until there has been a proper discussion on our forum 👍

eylenburg:
I started a discussion here: https://discuss.privacyguides.net/t/add-magic-earth-to-navigation-maps-tools/25752 

I don't think it had its own discussion in the forum yet, at least I couldn't find it with the search.

PR #2948, opened 2025-03-11
update!: add HERE WeGo recommendation

eylenburg:
This is the PR for adding HERE WeGo once/if this is agreed. I had previously included it in the original Maps PR (https://github.com/privacyguides/privacyguides.org/pull/2827) but this was later reduced to only recommending Organic & Osmand, as further discussion is required before potentially approving HERE WeGo.

Relevant discussion: https://discuss.privacyguides.net/t/here-wego-navigation/23752


v3rm1n0:
Hows the state of this?

PR #2961, opened 2025-03-17
update!: Add Secureblue

friadev:
List of changes proposed in this PR:

- Add secureblue to desktop recommendations

Relevant discussion: https://discuss.privacyguides.net/t/secureblue-atomic-fedora-hardening/16086/306

PR #2962, opened 2025-03-18
style: Apply Wikipedia's style of capitalization

redoomed1:

- Update capitalization of tools which break the stylization guideline linked [here](https://github.com/privacyguides/privacyguides.org/pull/2961#pullrequestreview-2691616204)


friadev (approved):

jonaharagon (approved):
Thank you!

I am unsure whether we should do `Dnscrypt-proxy` like in this PR or if we should do `DNSCrypt-Proxy`. I'm fine with either, and the former is _technically_ what the style guide would have us do, but the latter might make more sense.

DNSCrypt themselves always reference dnscrypt-proxy in all lowercase, however, they do have _other_ tools which use the latter capitalization style, such as DNSCrypt-Wrapper and DNSCrypt-Loader (both referenced as such on https://www.dnscrypt.org/)

redoomed1:
> the latter might make more sense

I agree with this and appreciate the other examples you brought up. I also changed `Dns0.eu` to `DNS0.eu`, considering that DNS as an acronym is capitalized for the most part.

friadev (approved):

PR #2963, opened 2025-03-18
update: Move custom domain support to best-case criteria for email services

redoomed1:

- Move custom domain support from Minimum to Qualify to Best Case based on the consensus reached in this forum discussion: https://discuss.privacyguides.net/t/should-we-require-custom-domain-support-for-email-providers/20735


jonaharagon (approved):

friadev:
Can we just remove it altogether? Seems like it's not relevant to privacy at all. I think the less we can clutter up our criteria the easier it'll be to parse, and if it's not relevant to privacy then it doesn't need to be there IMO.

friadev (approved):

PR #2986, opened 2025-03-28
update!: Add "Big Tech" page

kpham42:

-
-added `Big Tech.md`

-->

### DIFF
--- a/docs/getting-started/big-tech.md
+++ b/docs/getting-started/big-tech.md
@@ -2,3 +2,76 @@
 title: Avoiding Big Tech
 icon: material/domain
 ---
+---
+title: Avoiding Big Tech
+icon: material/domain
+---
+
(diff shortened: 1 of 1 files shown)


friadev (changes requested):
Not approved on the forum.

I don't really agree with this way of thinking about things. "Big Tech" includes things like Google Pixels which are some of the most secure devices money can buy, and makes people think they should avoid really solid E2EE services and devices that might fit their threat model. There's plenty of "small tech" companies that do a terrible job of keeping you private and secure, like /e/OS which advertises itself as "deGoogled" but its services are not E2EE and it lacks pretty much all security features of a Pixel and AOSP. I'd prefer if we push people toward looking for things like E2EE, sandboxing, audits, etc, things that tangibly improve security and not vague concepts like Big Tech.

jonaharagon:
Thanks Kevin! Because of the concerns fria mentioned, we're going to flesh out these guides on the forum (see FORUM-27347 and FORUM-27342) where they can be iterated on and discussed with the larger community, instead of working on them here.

PR #2990, opened 2025-04-02
Change title to april fools post

blacklight447:
Make it even more obvious for folks who missed it and read it after april fools.


jonaharagon:
Already addressed when it became April 2nd UTC time in 7cd2a723e38cdc95e687bffcf406257d09c995ae

PR #3006, opened 2025-04-15
update: Recommend both Fedora desktops equally

jonaharagon:
Fedora's website now lists KDE Plasma alongside Workstation under their official Editions. I see no reason to treat this differently than Fedora Atomic Desktops, so this brings both recommendations more in line with each other.

https://discuss.privacyguides.net/t/fedora-kde-is-now-on-the-samel-level-as-fedora-gnome/26706

### DIFF
--- a/docs/desktop.md
+++ b/docs/desktop.md
@@ -14,15 +14,15 @@ Linux distributions are commonly recommended for privacy protection and software
 
 ## Traditional Distributions
 
-### Fedora Workstation
+### Fedora Linux
 
 <div class="admonition recommendation" markdown>
 
 ![Fedora logo](assets/img/linux-desktop/fedora.svg){ align=right }
 
-**Fedora Workstation** is our recommended distribution for people new to Linux. Fedora generally adopts newer technologies (e.g., [Wayland](https://wayland.freedesktop.org) and [PipeWire](https://pipewire.org)) before other distributions. These new technologies often come with improvements in security, privacy, and usability in general.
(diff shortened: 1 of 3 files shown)


docs/desktop.md
```diff
@@ -14,23 +14,25 @@ Linux distributions are commonly recommended for privacy protection and software
 
 ## Traditional Distributions
 
-### Fedora Workstation
+### Fedora Linux
```
jonaharagon, line 17: I was trying to decide between **Fedora Linux** and **Fedora Desktop Editions** for this header. Ultimately I chose the former because it is in line with our Arch Linux recommendation. Thoughts?

friadev: It should be Fedora Linux since that's what they call it and we're linking to the main Fedora page now and not just the desktops.

blacklight447:
Did you take the fact the Gnome has a security benefit over KDE into account in this PR?

jonaharagon:
No, because we did not with Fedora Atomic Desktops. The *primary* reason we recommended Fedora with GNOME previously was simply because GNOME had Edition status, which KDE also does now.

Do you have a link to a forum discussion about these benefits?

I did see this discussion which was inconclusive. I am not considering people who think GNOME *might* be better in the long-term. **If** and when that turns out to be true then we can change our recommendation at that point. https://discuss.privacyguides.net/t/security-comparison-gnome-and-kde-plasma/26656/4

I'm unconvinced by historical arguments based on GNOME being more popular, because presently KDE has much more mainstream adoption from hardware desktop manufacturers (almost entirely due to the Steam Deck, but still)

Side note, but also: see #3001 please

dngray (approved):

friadev (approved):

PR #3007, opened 2025-04-15
update: Remove sentence about blocking OCSP checks in macOS

friadev:

- Removed link to a page recommending blocking OCSP checks and removing signatures from apps in macOS. These checks are an important security feature and shouldn't be removed altogether, as even the blog itself says. Their solution for individual apps to to remove the signature from the app, which prevents the integrity from being checked. On Apple Silicon Macs, it won't even let you run it. There's no [user info](https://support.apple.com/en-us/102445) included in the OCSP checks, and if you want to protect your IP address you should be using a VPN anyway.

### DIFF
--- a/docs/os/macos-overview.md
+++ b/docs/os/macos-overview.md
@@ -21,8 +21,6 @@ macOS performs online checks when you open an app to verify whether an app conta
 
 Apple's OCSP service uses HTTPS encryption, so only they are able to see which apps you open. They've [posted information](https://support.apple.com/HT202491) about their logging policy for this service. They additionally [promised](http://lapcatsoftware.com/articles/2024/8/3.html) to add a mechanism for people to opt-out of this online check, but this has not been added to macOS.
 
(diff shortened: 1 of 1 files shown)


docs/os/macos-overview.md
jonaharagon, line 1: OCSP is a well-known, significant privacy risk, which is why it is being deprecated by major Certificate Authorities like Let's Encrypt.

Our existing text notes the security implications of making this change, so no change is necessary here.

jonaharagon (changes requested):

docs/os/macos-overview.md
friadev: Can we at least delete the link to the blog? It’s really bad advice to remove the signature off an app, or block all OCSP checks.

friadev: I’m thinking about making a guide on `pf` which is the built-in firewall on macOS. One of the examples could be blocking these OCSP checks, or other telemetry. I will link to [this](https://support.apple.com/en-us/101555) and [this](https://support.apple.com/en-us/103229) that explain what their domains are for to make it easier for a user to decide what they want to block.

PR #3076, opened 2025-07-17
update: Set “Allow Diagnostic Data” to Disabled in Windows LGPO Overview

blackaly:
This PR changes the recommended setting for Allow Diagnostic Data from Enabled to Disabled to better align with privacy-focused practices. Disabling this setting reduces the amount of diagnostic data sent to Microsoft, which is more appropriate for users seeking minimal data sharing.


jonaharagon:
Incorrect:

> If you disable or do not configure this policy setting, the device will send required diagnostic data and the end user can choose whether to send optional diagnostic data from the Settings app.

PR #3081, opened 2025-07-24
style: Swap "Chromium" with "Blink"

ghost:

-Changing "Chromium engine" to "Blink engine". The current wording is not true, because Brave actually uses Chromium on iOS, like many other Browsers. You can observe that by typing in chrome://version, like on every other platform, or check the release notes or just check the build instructions. Chromium is available on iOS for a long time and Apple allows it. Many people just don't see the differnce between Chromium and Blink. Chromium itself is not even an engine, Blink, what is part of Chromium, is a engine. And Chromium on iOS exists, but it doesn't used the Blink engine, because that wouldn't be allowed by Apple.

### DIFF
--- a/docs/mobile-browsers.md
+++ b/docs/mobile-browsers.md
@@ -267,7 +267,7 @@ This disables update checks for the unmaintained Bromite adblock filter.
 
 ## Safari (iOS)
 
-On iOS, any app that can browse the web is [restricted](https://developer.apple.com/app-store/review/guidelines) to using an Apple-provided [WebKit framework](https://developer.apple.com/documentation/webkit), so a browser like [Brave](#brave) does not use the Chromium engine like its counterparts on other operating systems.
+On iOS, any app that can browse the web is [restricted](https://developer.apple.com/app-store/review/guidelines) to using an Apple-provided [WebKit framework](https://developer.apple.com/documentation/webkit), so a browser like [Brave](#brave) does not use the Blink engine (the core component of Chromium) like its counterparts on other operating systems.
 
 <div class="admonition recommendation" markdown>
 


docs/mobile-browsers.md
jonaharagon, line 1: I don't think the average reader will recognize Blink, but they understand what we mean when we say Chromium. We should either not make this change (my preference), or actually explain what Blink is for clarity.

docs/mobile-browsers.md
```diff
@@ -267,7 +267,7 @@ This disables update checks for the unmaintained Bromite adblock filter.
 
 ## Safari (iOS)
 
-On iOS, any app that can browse the web is [restricted](https://developer.apple.com/app-store/review/guidelines) to using an Apple-provided [WebKit framework](https://developer.apple.com/documentation/webkit), so a browser like [Brave](#brave) does not use the Chromium engine like its counterparts on other operating systems.
+On iOS, any app that can browse the web is [restricted](https://developer.apple.com/app-store/review/guidelines) to using an Apple-provided [WebKit framework](https://developer.apple.com/documentation/webkit), so a browser like [Brave](#brave) does not use the Blink engine like its counterparts on other operating systems.
```
dngray:
```suggestion
On iOS, any app that can browse the web is [restricted](https://developer.apple.com/app-store/review/guidelines) to using an Apple-provided [WebKit framework](https://developer.apple.com/documentation/webkit), so a browser like [Brave](#brave) does not use the Blink engine (the core component of Chromium) like its counterparts on other operating systems.
```

dngray: @jonaharagon what do you think about that? or can you think of something better?

friadev: We could just delete “engine” and have it say “does not use Chromium”

friadev:
```suggestion
On iOS, any app that can browse the web is [restricted](https://developer.apple.com/app-store/review/guidelines) to using an Apple-provided [WebKit framework](https://developer.apple.com/documentation/webkit), so a browser like [Brave](#brave) does not use Chromium like its counterparts on other operating systems.
```

ghost: @dngray LGTM. Not to complicated for the average reader, but actually correct.

ghost: @friadev The entire PR is about that exactly this wording would not be true, because it actually uses Chromium

friadev: https://developer.apple.com/support/alternative-browser-engines They only allow alternate browser engines in the EU.

ghost: @friadev 

> https://developer.apple.com/support/alternative-browser-engines They only allow alternate browser engines in the EU.

Yes, that's the reason why Chromium uses WebKit on iOS and not Blink. Chromium is just an open-source project with many sub-projects, there is no reason for Apple to forbide anyone to use it. But Blink, the actual "engine" is just a part of Chromium, and not included in the iOS version of Chromium that Brave, Chrome and some other Browsers are using, it still uses WebKit for the page rendering. It is exactly like that since many years and has nothing to do with the new EU-law, although it is theoretically possible since then to compile experimental builds with Blink (currently no public iOS browsers are using that). See https://chromium.googlesource.com/chromium/src/+/master/docs/ios/build_instructions.md

jonaharagon: > what do you think about that?

I like this change 👍

jonaharagon (approved):

friadev (approved):

ghost:
@jonaharagon @friadev Thanks for approving. Can you merge it? I can't merge because the commit is unsigened as I rebased (Don't know if that was even requiered.)

PR #3092, opened 2025-08-11
Refine Brave configucation

mmrabbani:
There is no need for user to change many configurations. Brave enables many privacy feature out of the box. So I showed which configurations should be changed


friadev:
I don’t think it’s necessary to show what’s enabled by default, you’re already there in the settings anyway and now we would have to keep track of what settings Brave enables by default, and do that for every other browser for consistency, which I don’t really think is necessary or useful.

mmrabbani:
At first glance, users see they should change so many settings, but in reality, there is few settings to change.

freddy-m:
Going to close this until an appropriate [forum discussion](https://discuss.privacyguides.net/) has taken place 👍

PR #3094, opened 2025-08-12
update: Add warning about Windows native metadata removal

kuferu:
added an important warning regarding the limitations of the built-in metadata remover of Windows

### DIFF
--- a/docs/data-redaction.md
+++ b/docs/data-redaction.md
@@ -9,7 +9,7 @@ cover: data-redaction.webp
 
 - [:material-account-search: Public Exposure](basics/common-threats.md#limiting-public-information){ .pg-green }
 
-When sharing files, be sure to remove associated metadata. Image files commonly include [Exif](https://en.wikipedia.org/wiki/Exif) data. Photos sometimes even include GPS coordinates in the file metadata.
+When sharing files, be sure to remove associated metadata. Most common file types (including documents, images, and videos) include metadata. Image files, for example, commonly include [Exif](https://en.wikipedia.org/wiki/Exif) data. Photos sometimes even include GPS coordinates in the file metadata. Windows has a built-in metadata remover, but it has a [very limited number](https://digitalconfidence.com/Remove-Properties-and-Personal-Information-a-Misleading-Feature.html) of supported file types and metadata types, necessitating a reliable third-party tool.
 
 <div class="admonition warning" markdown>
 <p class="admonition-title">Warning</p>


docs/data-redaction.md
```diff
@@ -9,7 +9,7 @@ cover: data-redaction.webp
 
 - [:material-account-search: Public Exposure](basics/common-threats.md#limiting-public-information){ .pg-green }
 
-When sharing files, be sure to remove associated metadata. Image files commonly include [Exif](https://en.wikipedia.org/wiki/Exif) data. Photos sometimes even include GPS coordinates in the file metadata.
+When sharing files, be sure to remove associated metadata. Most common file types (including documents, images, and videos) include metadata. Image files, for example, commonly include [Exif](https://en.wikipedia.org/wiki/Exif) data. Photos sometimes even include GPS coordinates in the file metadata. Windows has a built-in metadata remover, but it has a [very limited number](https://digitalconfidence.com/Remove-Properties-and-Personal-Information-a-Misleading-Feature.html) of supported file types and metadata types, necessitating a reliable third-party tool.
```
redoomed1: The linked article by Digital Confidence about limitations in the metadata removal function of Windows's File Explorer also promotes the company's own software as a solution and links to their website's articles as "related white papers."

We prefer links to first-party documentation (in this case, documentation from Microsoft) if available, and encourage not linking to articles which promote a product created by the same company which hosts the article.

redoomed1 (changes requested):

docs/data-redaction.md
kuferu: I prefer to link to a documentation by Microsoft as well, but it does not exists. You got to explain to people who browse privacyguides.org why do they need a good 3rd-party metadata scrubber anyway, when Windows already have one built-in. The web is filled to the brim with articles, videos, blog posts, and discussions posts, containing bogus info about the built-in scrubber, saying it can remove all metadata from all kind of files. Just search for "remove properties and personal information". It's important to refute it. The article by Digital Confidence is not ideal, but it certainly better than nothing.

friadev:
I don’t think this is needed. The page isn’t specifically about Windows, and as stated I don’t want to link to an ad for some sketchy software. The entire page already has third party metadata removal tools, that’s what the intro is leading into.

docs/data-redaction.md
dngray: That document also mentions word documents. I think for that you're supposed to use office's own [document inspector](https://learn.microsoft.com/en-us/office/vba/library-reference/concepts/using-the-document-inspector) which that article mentions, though it also mentions no batch support.

Personally I'm not opposed to the Digital Confidence article being mentioned, because it is factual, we're not endorsing their software specifically and maybe could mention that.

dngray:
They've been around since like 2009, so I'm not precisely sure where that claim comes from, but I do agree in regard to it not being an advertisement for that particular solution.

docs/data-redaction.md
kuferu: You can also use the foss MAT2 for Word files. (it's listed in PG) LibreOffice doesn't have a tool akin to the Document Inspector.

Anyway, half a year has passed, and my proposed change hasn't been implemented still.

I've previously mentioned bogus info on the web about the Win built-in scrubber. Here are some examples google is referring to:

https://www.kaspersky.com/blog/how-to-remove-metadata/52913/
"You can also remove PDF metadata using File Explorer in the same way as for photos and videos."

https://www.popsci.com/diy/how-to-remove-metadata/
"Scrubbing files clean is easy on Microsoft’s operating system: You can use the same method across all types."

https://www.digitalcitizen.life/remove-metadata-file/
"This file can be anything"... " a PDF file, etc."... "clones the selected file, and the copy has no metadata associated with it."

These are all wrong! There are only few file formats supported, and PDF is not supported at all! PG will be doing a great service to its visitors by refuting this.

dngray: This is why I'm not a fan of these "catch all" tools they generally don't work thoroughly. When I have wanted to remove metadata from PDFs, I've specifically used Adobe Acrobat's "Remove Hidden Information" feature or Word's "Document inspector".

I'm pretty sure MAT actually does this by using poppler to convert the pages to images which means you lose all selectable text, often not what you want unless you use the `--lightweight` mode, how well that works is another question. What I like about acrobat is it shows you a before and after view of what is going to be removed and what exists to remove.

friadev (approved):

friadev:
@redoomed1 the link is removed now, could you take a look and see if you like the new version?

docs/data-redaction.md
```diff
@@ -5,11 +5,17 @@ icon: material/tag-remove
 description: Use these tools to remove metadata like GPS location and other identifying information from photos and files you share.
 cover: data-redaction.webp
 ---
+
 <small>Protects against the following threat(s):</small>
 
 - [:material-account-search: Public Exposure](basics/common-threats.md#limiting-public-information){ .pg-green }
 
-When sharing files, be sure to remove associated metadata. Image files commonly include [Exif](https://en.wikipedia.org/wiki/Exif) data. Photos sometimes even include GPS coordinates in the file metadata.
+When sharing files, be sure to remove associated metadata. Most common file types (including documents, images, and videos) include metadata. Image files, for example, commonly include Exif data. Photos sometimes even include GPS coordinates in the file metadata. Windows has a built-in metadata remover, but unfortunately it cannot remove many types of data such as:
```
redoomed1:
```suggestion
When sharing files, be sure to remove associated metadata. Most common file types (including documents, images, and videos) include metadata. Image files, for example, commonly include Exif data. Photos sometimes even include GPS coordinates in the file metadata.

Windows has a built-in metadata remover, but unfortunately it cannot remove many types of data such as:
```

docs/data-redaction.md
```diff
@@ -5,11 +5,17 @@ icon: material/tag-remove
...
+
+- Documents: Comments, author names, tracked changes, hidden worksheets, slide notes, custom XML, and document revision histories.
+- Images: Camera serial numbers, XMP/IPTC data, C2PA metadata, and image thumbnails (which can dangerously expose original details from cropped or erased areas).
+- Audio & Video: XMP metadata, C2PA metadata, and certain ID3v2 tag fields.
+- Embedded Files: Hidden metadata nested inside images that are embedded within your documents.
```
redoomed1:
```suggestion
- Embedded Files: Hidden metadata nested inside images that are embedded within documents.
```

redoomed1 (approved):
LGTM

PR #3112, opened 2025-09-08
update!: Add Revolut

GorujoCY:

- Added Revolut in Card masking section of Financial Services

Relevant discussion: https://discuss.privacyguides.net/t/revolut-multinational-card-masking-financial-service/30879


GorujoCY:
Missing discussion, working on making it.

jonaharagon (changes requested):
I will have to look for the previous comments to remember exactly why, but IIRC we _specifically_ excluded Revolut for a reason when making this page, so this certainly needs more discussion.

GorujoCY:
discussion is added, apologies for the inconvenience

https://discuss.privacyguides.net/t/revolut-international-card-masking-financial-service/30879

GorujoCY:
I see some concerning statements in the thread, that would not be true anymore let's get some FAQ **obviously based on my experience**:

Q: Can you use a fake billing address?
A: Yes, personally I was able to trick HIDIVE once into thinking im in the US, the transaction went through. Watch out for some exceptions like Apple, you may likely not be able to trick Apple into thinking you're in the US with billing, however all other billing information except country to my understanding can be faked

Q: Masked Cards are pointless as it's in your name anyways
A: Untrue, for providers that there was no necessary reason to do so I have never given them my full name, if only abbreviations like `G`, Transactions still went through. I still do not see solid evidence that the card can reveal your full name, outside of my experience that it isn't

Q: Revolut shares data with Facebook/Meta
A: Was true back in 2020 ([when that article was written](https://medium.com/swlh/revolut-is-sharing-your-data-with-facebook-heres-what-you-need-to-know-cc511c7ab639)) but today checking both using https://reports.exodus-privacy.eu.org/en/reports/com.revolut.revolut/latest/ and [AppManager](https://github.com/MuntashirAkon/AppManager) on my installation of Revolut, this no longer exists. **However it is true that forms of google analytics do exist in the app, I have not seen signs of AppFlyer doe**. Of course I am open for others to confirm or deny **with evidence for the reason of denial**  [**Keep in mind this refers to the android application itself not what’s done outside of it**]

GorujoCY:
wanted to share a user's post that I have the same sentiment with on top of my statement to ensure fairness here

https://discuss.privacyguides.net/t/revolut-international-card-masking-financial-service/30879/16

ph00lt0:
OP left community

PR #3156, opened 2025-10-17
Proton is based in Plan-les-Ouates, not Geneva (blog)

ghost:

+ Updates the Private European Alternatives page to correctly state Proton's location.


ph00lt0:
Geneve is also the name of the Canton Plan-les-Ouates is part of. Proton themselves also write on their website to be based in Geneva. I see no reason to change this.

PR #3166, opened 2025-11-12
style: encrypted DNS prevents manipulation not spoofing

ignoramous:
also: discuss.privacyguides.net/t/secure-encryption-and-online-anonymity-are-now-at-risk-in-switzerland/26181/28

-


docs/basics/vpn-overview.md
```diff
@@ -59,7 +59,7 @@ You also should not trust a VPN to secure your connection to an unencrypted, HTT
...
 
-Another common reason encrypted DNS is recommended is that it prevents DNS spoofing. However, your browser should already be checking for [TLS certificates](https://en.wikipedia.org/wiki/Transport_Layer_Security#Digital_certificates) with **HTTPS** and warn you about it. If you are not using **HTTPS**, then an adversary can still just modify anything other than your DNS queries and the end result will be little different.
+Another common reason encrypted DNS is recommended is that it prevents DNS manipulation attacks. However, your browser should already be checking for [TLS certificates](https://en.wikipedia.org/wiki/Transport_Layer_Security#Digital_certificates) with **HTTPS** and warn you about it. If you are not using **HTTPS**, then an adversary can still just modify anything other than your DNS queries and the end result will be little different.
```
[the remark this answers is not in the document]
dngray: Yeah we're not calling it snooping sorry copilot but you wrong.

friadev (approved):

ignoramous:
This is what "DNS Spoofing" means (which isn't what encrypted DNS will help in anyway against):

>  DNS spoofing, also referred to as DNS cache poisoning, is a form of computer security [hacking](https://en.wikipedia.org/wiki/Security_hacker) in which corrupt [Domain Name System](https://en.wikipedia.org/wiki/Domain_Name_System) data is introduced into the [DNS resolver](https://en.wikipedia.org/wiki/DNS_resolver)'s [cache](https://en.wikipedia.org/wiki/DNS_cache), causing the [name server](https://en.wikipedia.org/wiki/Name_server) to return an incorrect result record, e.g. an [IP address](https://en.wikipedia.org/wiki/IP_address). This results in [traffic being diverted](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) to any computer that the attacker chooses. Put simply, a hacker makes the device think it is connecting to the chosen website, when in reality, it is redirected to a different website by altering the [IP address](https://en.wikipedia.org/wiki/IP_address) associated with the [domain name](https://en.wikipedia.org/wiki/Domain_name) in the DNS server.

From: https://en.wikipedia.org/wiki/DNS_spoofing

docs/basics/vpn-overview.md
ph00lt0, line 62: I would agree the term used isnt fully right but i would propose to use the term DNS hijacking.

ph00lt0 (changes requested):

docs/basics/vpn-overview.md
dngray: I'm inclined to agree as spoofing in general is a jargon term, that doesn't mean a whole lot unless you know what that generally means. I think hijacking is probably a better term. The site is written in such a way we avoid having to link to wikipedia to explain certain jargon terms when they can be explained in a different way and on the site itself.

ignoramous: "DNS hijacking" is taking over the stub resolver entirely.

> DNS hijacking, DNS poisoning, or DNS redirection is the practice of subverting the resolution of [Domain Name System](https://en.wikipedia.org/wiki/Domain_Name_System) (DNS) queries.[[1]](https://en.wikipedia.org/wiki/DNS_hijacking#cite_note-1) This can be achieved by malware that overrides a computer's [TCP/IP](https://en.wikipedia.org/wiki/Internet_protocol_suite) configuration to point at a rogue DNS server under the control of an attacker, or through modifying the behaviour of a trusted DNS server so that it does not comply with [internet standards](https://en.wikipedia.org/wiki/Internet_Standard).

https://en.wikipedia.org/wiki/DNS_hijacking

DNS Manipulation is a subset of DNS hijacking and imo precisely outlines what encrypted DNS transports prevent (these transports cannot prevent DNS hijacking entirely).

https://en.wikipedia.org/wiki/DNS_hijacking#Manipulation_by_ISPs

ph00lt0: I stand corrected here.

ph00lt0 (approved):

PR #3139, opened 2025-09-26
update: Clarify Windows Home instructions

raghavkaul:
Clarify Windows Home instructions. On Windows Home, following the instructions will enable Device Encryption, which is a less-featured version of Bitlocker that does not support all of the security settings (see [here](https://debugging.works/blog/windows-bitlocker/) for more details).

One notable missing feature, for example, is the ability to use password + pin (or, really, most protectors that are not `rp` or `tpm`).

PR #3170, opened 2025-11-18
update: Bitlocker cmdline workaround no longer works in Home

dngray:

- Closes https://github.com/privacyguides/privacyguides.org/pull/3139


docs/encryption.md
```diff
@@ -95,61 +96,28 @@ TrueCrypt has been [audited a number of times](https://en.wikipedia.org/wiki/Tru
...
 
-**BitLocker** is the full volume encryption solution bundled with Microsoft Windows that uses the Trusted Platform Module ([TPM](https://learn.microsoft.com/windows/security/information-protection/tpm/how-windows-uses-the-tpm)) for hardware-based security.
+**BitLocker** is the full volume encryption solution bundled with Microsoft Windows that uses the Trusted Platform Module ([TPM](https://learn.microsoft.com/windows/security/information-protection/tpm/how-windows-uses-the-tpm)) for hardware-based security. We recommend that you use Bitlocker with the [TPM+PIN](https://learn.microsoft.com/en-us/windows/security/operating-system-security/data-protection/bitlocker/faq#what-is-the-difference-between-a-tpm-owner-password--recovery-password--recovery-key--pin--enhanced-pin--and-startup-key) option and not just your regular password as [extraction is impossible](https://blog.elcomsoft.com/2021/01/understanding-bitlocker-tpm-protection) when you use an extra a pre-boot protector like the PIN. The PIN is rate limited and the TPM will panic and lock access to the encryption key either permanently or for a period of time if someone attempts to brute force access.
```
friadev:
```suggestion
**BitLocker** is the full volume encryption solution bundled with Microsoft Windows that uses the Trusted Platform Module ([TPM](https://learn.microsoft.com/windows/security/information-protection/tpm/how-windows-uses-the-tpm)) for hardware-based security. We recommend that you use BitLocker with the [TPM+PIN](https://learn.microsoft.com/en-us/windows/security/operating-system-security/data-protection/bitlocker/configure?tabs=os#require-additional-authentication-at-startup) option and not just your regular password as [extraction is impossible](https://blog.elcomsoft.com/2021/01/understanding-bitlocker-tpm-protection) when you use a pre-boot protector like the PIN. The PIN is rate limited and the TPM will panic and lock access to the encryption key either permanently or for a period of time if someone attempts to brute force access.
```

jordan-warne: The TPM+PIN page you've linked doesn't really explain how to use the TPM+PIN.

This one might be better?

https://learn.microsoft.com/en-us/windows/security/operating-system-security/data-protection/bitlocker/configure?tabs=os#require-additional-authentication-at-startup

docs/encryption.md
```diff
@@ -95,61 +96,28 @@ TrueCrypt has been [audited a number of times](https://en.wikipedia.org/wiki/Tru
...
-    ```
-
-<div class="admonition tip" markdown>
-<p class="admonition-title">Tip</p>
-
-Backup `BitLocker-Recovery-Key.txt` on your Desktop to a separate storage device. Loss of this recovery code may result in loss of data.
-
-</div>
+BitLocker is [officially supported](https://support.microsoft.com/en-us/windows/bitlocker-overview-44c0c61c-989d-4a69-8822-b95cd49b1bbf) on the Pro, Enterprise, and Education editions of Windows. If you have Home edition we recommend you [upgrade to Pro](https://support.microsoft.com/en-us/windows/upgrade-windows-home-to-windows-pro-ef34d520-e73f-3198-c525-d1a218cc2818), which can be achieved without reinstalling Windows or losing your files. An alternative solution could be to use VeraCrypt's [system encryption](https://veracrypt.io/en/System%20Encryption.html) feature. VeraCrypt does not use the system's TPM chip and all encryption keys are stored in memory, leaving them [vulnerable to extraction](https://blog.elcomsoft.com/2021/06/breaking-veracrypt-obtaining-and-extracting-on-the-fly-encryption-keys) while the device is online and mounted. We also recommend if you're using Bitlocker to make sure you
```
jordan-warne: If we are already recommending people upgrade to Pro or Enterprise versions, I don't think we should recommend VeraCrypt for this when the Windows built-in tools are there and provide superior security.

It seems like VeraCrypt's implementation is a bit buggy as well, especially since the upgrade from Windows 10 to Windows 11.

https://github.com/veracrypt/VeraCrypt/issues?q=Full%20disk%20encryption

jordan-warne: Windows 11 Home also automatically enables device encryption as long as the device is officially supported.

https://support.microsoft.com/en-us/windows/device-encryption-in-windows-cf7e2b6f-3e70-4882-9532-18633605b7df

docs/encryption.md
dngray: Might link to that actually, and simply say that you should click on Settings > Accounts > Sign-in options and then select either "PIN (Windows Hello) or "Security Key", you can do it with a local user account too, don't need to have signed in with a microsoft account.

jordan-warne: I think you might be misunderstanding, Windows Hello is different from the Bitlocker startup PIN.

dngray: Oh, yeah. The TPM+PIN is pre boot not post boot like login. I forgot about that

Image

Seems like it's more like LUKS password on Linux.

docs/encryption.md
dngray: I found on my desktop computer automatic device encryption was not working. It turned out because of that last one regarding PCR7 binding. It turns out there's a few other reasons besides the ones stated there that can prevent that for example if I go into.

1. Win + R
2. Enter "msinfo32"
3. Ctrl + Shift + Enter

The reason is "Hardware Security Test Interface failed and device is not Modern Standby, Un-allowed DMA capable bus/device(s) detected." Kind of odd this machine doesn't have anything like Thunderbolt. Turns out the reason for that can be a little more obscure https://learn.microsoft.com/en-us/windows-hardware/design/device-experiences/oem-bitlocker#un-allowed-dma-capable-busdevices-detected

friadev: > It seems like VeraCrypt's implementation is a bit buggy as well, especially since the upgrade from Windows 10 to Windows 11.

True but that's a PR for another day. Let's get this one sorted first.

PR #3172, opened 2025-11-19
update: fix inaccurate claim about mullvad browser having the same protections as arkenfox

friadev:

- remove statement about Mullvad Browser having the same protections as Arkenfox
- removed outdated sentence about being close to Mullvad Browser's release

### DIFF
--- a/docs/desktop-browsers.md
+++ b/docs/desktop-browsers.md
@@ -222,7 +222,7 @@ Max Protection enforces the use of DNS over HTTPS, and a security warning will s
 <div class="admonition tip" markdown>
 <p class="admonition-title">Use Mullvad Browser for advanced anti-fingerprinting</p>
 
-[Mullvad Browser](#mullvad-browser) provides more anti-fingerprinting protections out of the box than Firefox, and does not require the use of Mullvad's VPN to benefit from these protections. Coupled with a VPN, Mullvad Browser can thwart more advanced tracking scripts which Arkenfox cannot. Firefox still has the advantage of being much more flexible, and allowing per-site exceptions for websites which you need to stay logged in to.
+[Mullvad Browser](#mullvad-browser) provides stronger anti-fingerprinting protections out of the box than Firefox, and does not require the use of Mullvad's VPN to benefit from these protections. Coupled with a VPN, Mullvad Browser can thwart more advanced tracking scripts which Arkenfox cannot. Firefox still has the advantage of being much more flexible, and allowing per-site exceptions for websites which you need to stay logged in to.
 
 </div>
 


docs/desktop-browsers.md
jonaharagon, line 1: The original statement is accurate I believe, MB does provide the same protections. It just **also** provides more protections. If we clarify this statement, that fact should be clear in the change.

jonaharagon (changes requested):

docs/desktop-browsers.md
friadev: It uses RFP vs FPP in Arkenfox by default and you're meant to customize Arkenfox more. It's just way too complicated to say "the same protections" I think. This way it's also more concise and understandable. If you have some more specific things you can suggest changing maybe then we can start there.

docs/desktop-browsers.md
```diff
@@ -96,11 +96,9 @@ Modifications other than adjusting this setting would make your fingerprint uniq
 
 ### Anti-Fingerprinting
 
-**Without** using a [VPN](vpn.md), Mullvad Browser provides the same protections against [naive fingerprinting scripts](https://github.com/arkenfox/user.js/wiki/3.3-Overrides-%5BTo-RFP-or-Not%5D#-fingerprinting) as other private browsers like Firefox+[Arkenfox](#arkenfox-advanced) or [Brave](#brave). Mullvad Browser provides these protections out of the box, at the expense of some flexibility and convenience that other private browsers can provide.
```
jonaharagon, line 99: I'm concerned that by removing this line entirely, people who only read the Mullvad Browser entry won't realize it is a perfectly acceptable alternative to Firefox even without a VPN.

I agree the new statement later on in the Arkenfox section is more concise 👍

docs/desktop-browsers.md
```diff
@@ -222,7 +220,7 @@ Max Protection enforces the use of DNS over HTTPS, and a security warning will s
 <div class="admonition tip" markdown>
 <p class="admonition-title">Use Mullvad Browser for advanced anti-fingerprinting</p>
 
-[Mullvad Browser](#mullvad-browser) provides the same anti-fingerprinting protections as Arkenfox out of the box, and does not require the use of Mullvad's VPN to benefit from these protections. Coupled with a VPN, Mullvad Browser can thwart more advanced tracking scripts which Arkenfox cannot. Arkenfox still has the advantage of being much more flexible, and allowing per-site exceptions for websites which you need to stay logged in to.
+[Mullvad Browser](#mullvad-browser) provides anti-fingerprinting protections out of the box, and does not require the use of Mullvad's VPN to benefit from these protections. Coupled with a VPN, Mullvad Browser can thwart more advanced tracking scripts which Arkenfox cannot. Firefox still has the advantage of being much more flexible, and allowing per-site exceptions for websites which you need to stay logged in to.
```
xe-3: I think the first sentence would benefit from a bit of tweaking for clarity.

_"Mullvad Browser provides anti-fingerprinting protections out of the box"_ needs a relational adjective to contextualize it:

e.g. "Mullvad Browser provides **stronger** anti-fingerprinting protections than..." or "Mullvad Browser provides anti-fingerprinting protections **that go beyond** what Arkenfox enables out of the box.

I'd suggest this for two reasons.
1. This paragraph is in the Arkenfox section. It's not a paragraph _about_ MB, it's a paragraph about how MB _compares_ to AF.
2. As written right now, the first sentence can easily be read to imply that MB has anti-fp protection and AF / FF do not. Which is not accurate. Including a relational adjective ("stronger," "more robust" etc) that ambiguity is avoided.

I think you should also consider changing "advanced tracking scripts"to "advanced **fingerprinting** scripts" for clarity and accuracy.

friadev: Alright reworded it a bit.

PR #3183, opened 2025-12-25
docs/os: AVB does not prevent evil maid

ignoramous:
AVB may help deter rouge firmware changes and prevent tampering of the OS.

cc: @eylenburg / @SkewedZeppelin

### DIFF
--- a/docs/os/android-overview.md
+++ b/docs/os/android-overview.md
@@ -20,7 +20,7 @@ Key components of the Android security model include [verified boot](#verified-b
 
 ### Verified Boot
 
-[**Verified Boot**](https://source.android.com/security/verifiedboot) is an important part of the Android security model. It provides protection against [evil maid](https://en.wikipedia.org/wiki/Evil_maid_attack) attacks, malware persistence, and ensures security updates cannot be downgraded with [rollback protection](https://source.android.com/security/verifiedboot/verified-boot#rollback-protection).
+[**Verified Boot**](https://source.android.com/security/verifiedboot) is an important part of the Android security model. It protects from boot path changes, deters firmware tampering, prevents malware persistence, and ensures security updates cannot be downgraded with [rollback protection](https://source.android.com/security/verifiedboot/verified-boot#rollback-protection).
 
 Android 10 and above has moved away from full-disk encryption to more flexible [file-based encryption](https://source.android.com/security/encryption/file-based). Your data is encrypted using unique encryption keys, and the operating system files are left unencrypted.
 


SkewedZeppelin:
but it does? The encryption key is partially derived through the state of the verified boot chain.

an attacker could attempt to modify the ciphertext of the drive since that isn't authenticated but that likely isn't implied here

SkewedZeppelin:
thinking about it more, I agree a bit more.

the heads approach for example shows you a totp key you can manually verify to confim state before entering any passwords

it would be neat if eg. GrapheneOS allowed you to use its Auditor before first unlock

cc @thestinger

thestinger:
Don't agree with the statement that it doesn't protect against evil maid attacks, especially when considering the existence of Auditor and secondary profiles with separate encryption keys.

> the heads approach for example shows you a totp key you can manually verify to confim state before entering any passwords

It doesn't prevent proxying this to another machine and also doesn't really work since it fails to implement complete verified boot. TOTP is also very cryptographically insecure due to the very small output.

> it would be neat if eg. GrapheneOS allowed you to use its Auditor before first unlock

This is a planned feature but we haven't gotten around to implementing it yet. It could be done via a quick tile or lockscreen shortcut for starting Auditor in the Auditee mode.

ignoramous:
Evil maid attacks, at least in Mediatec SoC, have been shown to bypass Verified Boot: https://blog.quarkslab.com/android-data-encryption-in-depth.html

> We implemented our PoC on the Samsung A22 device (more precisely on the A226B and the A225F). These devices are using two vulnerable SoCs from Mediatek: MT6769V and MT6833V, which can be exploited using [MTKClient](https://github.com/bkerler/mtkclient).
>
> This tool interacts with the Download Mode (similar to the EDL mode on Qualcomm SoCs), which exposes a USB interface that is originally used to perform support operations on it (such as flashing the firmware). `To trigger the bug, physical access is required,` and on some devices (such as the A225F) two pins must be shorted on the device PCB to enter Download Mode.
>
> Once the device is booted in the right mode, the tool exploits the boot ROM vulnerability, then modifies BL2 (called preloader in the Mediatek boot schema) to disable the next secure boot check, and finally loads it on the device and boots on it.

At best, AVB may deter all but determined evil maids? Can't really prevent, I don't think?

Per https://www.cl.cam.ac.uk/teaching/1314/SecurityII/2014-security2-4-skorobogatov.pdf (2014), secure chips are eventually defeated in *years*, if given physical access.

And the Fuschia docs on [*Verified Execution*](https://fuchsia.dev/fuchsia-src/concepts/security/verified_execution) scope pretty broad:

> ...
>
> The goal of a defender in this security model is to recover by eliminating untrustworthy states (code and data) in which attackers could persist control across reboots.

friadev:
I think with the input of GOS I'm going to go ahead and close this.

ignoramous:
> I think with the input of GOS I'm going to go ahead and close this

Did you read Quarks Lab blog post, too? If so, what do you think of their "input" in successfully "evil maid" attacking *Verified Boot* & brute force decrypting the synthetic password used for FBE? /cc: @thestinger

thestinger:
> If so, what do you think of their "input" in successfully "evil maid" attacking Verified Boot

An exploit of a specific platform does not make security useless. Do you think anything which is ever exploited should be considered to not have security?

> decrypting the synthetic password used for FBE

That's not how encryption works and is not what happened.

ignoramous:
> An exploit of a specific platform does not make security useless

The security on offer by *Verified Boot*, imo, against "Evil Maids" (physical attacks), is a mitigation not "prevention" (and only against a specific type of "Evil Maids" not all):

The page goes, "[Android Verified Boot] provides protection against [evil maid](https://en.wikipedia.org/wiki/Evil_maid_attack) attacks" and links to the Wikipedia article on "Evil Maid" which inturn goes, "An evil maid attack can also be done by replacing the victim's device with an identical device." ... is all misleading the reader. 

>  and is not what happened

After bypassing AVB, Quarkslab patched the whole world and ended up with a PoC brute forcing user password, no? Believe, their exploit led to AOSP changing the order of the keys with which the synthetic password is now encrypted and decrypted.

thestinger:
> The security on offer by Verified Boot, imo, against "Evil Maids" (physical attacks), is a mitigation not "prevention" (and only against a specific type of "Evil Maids" not all):

That doesn't change that it's an important part of protecting against it and that unlike the described mechanism with incomplete verified boot / attestation and an insecure TOTP code actually does what it's meant to do properly when properly implemented.

> After bypassing AVB, Quarkslab patched the whole world and ended up with a PoC brute forcing user password, no? Believe, their exploit led to AOSP changing the order of the keys with which the synthetic password is now encrypted and decrypted.

No, not really. The device they targeted did not use the modern disk encryption implementation with Weaver. The legacy code path for devices not implementing the standard recommended security features doesn't particularly interest us and we don't consider those devices to have working disk encryption without the user setting a very strong passphrase.

ignoramous:
The changes I've proposed here is in the context of PG talking about AVB in general and linking to Wikipedia's definition of Evil Maid.

---

> That doesn't change that it's an important part of protecting against it 

I don't see how AVB can prevent all forms of "Evil Maids" (the definition the page links to is pretty broad). May be a subset, like the modification of bootloader and OS components (to install rootkits, for instance). 

> The device they targeted did not use the modern disk encryption implementation with Weaver.

Don't think Weaver is requirement in CDDs for AVB and/or FBE?  

> not implementing the standard recommended security features doesn't particularly interest us

You're in the right, of course. That said, the page is generally speaking about AVB in the context of *all* Android devices and not just GrapheneOS on Pixels.

thestinger:
> Don't think Weaver is requirement in CDDs for AVB and/or FBE?

It's not required but it's a complete joke for devices to be failing to be provide a core security feature first shipped by the Pixel in 2017. That said, most Android devices are a complete joke for security compared to iPhones and Pixels as a whole.

> You're in the right, of course. That said, the page is generally speaking about AVB in the context of all Android devices and not just GrapheneOS on Pixels.

It's hard to write about Android as a whole when non-Pixel devices are largely far behind on updates and skipping a bunch of security features. Pixels are far from perfect and hardly hardened devices but that just means everything else is a joke...

ignoramous:
> Pixels are far from perfect and hardly hardened devices but that just means everything else is a joke

Oh well ... https://discuss.grapheneos.org/d/12848-claims-made-by-forensics-companies-their-capabilities-and-how-grapheneos-fares (:

I guess you agree that the page needs to be careful about how it speaks about AVB mitigating a subset of Evil Maid attacks (as discussed on Wikipedia which the page links to), and for all Androids, in general? Because otherwise, @friadev just said above, "with the input of GOS I'm going to go ahead and close this"...

PR #3186, opened 2026-01-20
update: Remove mention of Machine-ID in Linux Overview

Tiagoquix:
- Machine ID uniqueness may be desired. See: https://github.com/secureblue/secureblue/issues/1121 and https://gitlab.tails.boum.org/tails/tails/-/issues/7100.
  - TL;DR of both links is: having a static machine-id makes it easier to identify that a machine is, for example, running the Tails OS.
  - It may also cause breakage on systemd-creds encryption: https://github.com/secureblue/secureblue/issues/1121#issuecomment-3054194314 (not confirmed, though)

---------------
I declare that I have no conflict of interest by sending this PR.

### DIFF
--- a/docs/os/linux-overview.md
+++ b/docs/os/linux-overview.md
@@ -159,7 +159,6 @@ There are other system identifiers which you may wish to be careful about. You s
 
 - **Hostnames:** Your system's hostname is shared with the networks you connect to. You should avoid including identifying terms like your name or operating system in your hostname, instead sticking to generic terms or random strings.
 - **Usernames:** Similarly, your username is used in a variety of ways across your system. Consider using generic terms like "user" rather than your actual name.
-- **Machine ID:** During installation, a unique machine ID is generated and stored on your device. Consider [setting it to a generic ID](https://madaidans-insecurities.github.io/guides/linux-hardening.html#machine-id).
 
 ### System Counting
 


friadev (approved):

docs/os/linux-overview.md
```diff
@@ -159,7 +159,6 @@ There are other system identifiers which you may wish to be careful about. You s
 
 - **Hostnames:** Your system's hostname is shared with the networks you connect to. You should avoid including identifying terms like your name or operating system in your hostname, instead sticking to generic terms or random strings.
 - **Usernames:** Similarly, your username is used in a variety of ways across your system. Consider using generic terms like "user" rather than your actual name.
-- **Machine ID:** During installation, a unique machine ID is generated and stored on your device. Consider [setting it to a generic ID](https://madaidans-insecurities.github.io/guides/linux-hardening.html#machine-id).
```
jonaharagon, line 162: Given we say:

> There are other system identifiers which you may wish to be careful about. You should give this some thought to see if it applies to your threat model

...it is unclear to me whether this line should be removed entirely. I'm inclined to agree that we probably do not want most people to adjust this setting, I have run into problems with machines with duplicate machine IDs on the same network myself.

However, it seems clear based on the issues linked in the OP as well as other resources we used to write this guide originally that machine ID _could_ have a _potential_ impact on one's privacy.

I would probably recommend keeping this line, removing the recommendation & link, and maybe adding information about when you might want to be concerned about this if it applies at all.

jonaharagon (changes requested):

docs/os/linux-overview.md
Tiagoquix: @jonaharagon Ultimately, there is no benefit in having a constant Machine-ID. If an evil actor can see your `/etc/machine-id`, he can also probably see `/etc/fstab`, which would also be another way to identify machine uniqueness, meaning the change done on `/etc/machine-id` would have no effect.

And, as you said, having the same ID on the local network can cause problems.

If we keep it and do not recommend the change, people will search on the internet about how to change it, so it would result in more or less the same outcome. Removing it completely is the way to go to avoid people changing it.

jonaharagon (approved):
Prior discussion: https://discuss.privacyguides.net/t/madaidens-security-machine-id-question-linux-hardening/16775

PR #3193, opened 2026-02-05
update: Stop using "zero-knowledge" and "zero-access" buzzwords

jonaharagon:

- Remove references to "zero-access"
- Remove references to "zero-knowledge" except when referring to "zero-knowledge proofs"
- Remove references to E2EE where it is not fully accurate

### DIFF
--- a/blog/posts/activists-guide-securing-your-smartphone.md
+++ b/blog/posts/activists-guide-securing-your-smartphone.md
@@ -221,7 +221,7 @@ On Android:
 2. Navigate to **Location**
 3. Switch **Use location** to **off**
 
-If you use an Android phone, you should also check your Google account settings to [ensure location history is disabled](https://support.google.com/accounts/answer/3118687). Google is frequently tapped by law enforcement to provide location data, because they don't protect your personal information with strong, zero-knowledge encryption.
+If you use an Android phone, you should also check your Google account settings to [ensure location history is disabled](https://support.google.com/accounts/answer/3118687). Google is frequently tapped by law enforcement to provide location data, because they don't protect your personal information with strong encryption.
 
 ### Check Your Keyboard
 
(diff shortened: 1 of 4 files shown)


jordan-warne (approved):
This is a good change, zero-knowledge and zero-access buzzwords are like the new military grade encryption.

redoomed1 (approved):

friadev:
I think we should say end-to-end encryption then instead of zero-knowledge which is not a [buzzword](https://support.apple.com/en-us/102651). Services will sometimes claim to be encrypted but they're actually just using TLS transport encryption, so the distinction is important.

[Client-side encryption](https://en.wikipedia.org/wiki/Client-side_encryption) might also be an option.

Obviously this doesn't apply to any mail services since it's impossible to have proper E2EE for all mail.

blog/posts/activists-guide-securing-your-smartphone.md
```diff
@@ -265,7 +265,7 @@ You should be prepared to have your phone taken or lost during a protest. You ca
 
 If you have an iPhone, you can make a local backup to a macOS computer or a Windows computer with iTunes. You can also back up to iCloud, but these backups are only secure if you enable [Advanced Data Protection](https://www.privacyguides.org/en/os/ios-overview/#icloud) on your iCloud account. We strongly encourage [enabling Advanced Data Protection](https://support.apple.com/en-us/108756) for all iCloud users in any case, as it protects not only device backups but most iCloud account data as well.
 
-The backup situation on Android is not nearly as robust unfortunately, but you can back up photos and files with a variety of services. If you use an online backup service we recommend choosing one with strong, zero-knowledge encryption so that the service provider is unable to access your data.
+The backup situation on Android is not nearly as robust unfortunately, but you can back up photos and files with a variety of services. If you use an online backup service we recommend choosing one which uses strong encryption with keys you control, so that the service provider is unable to access your data.
```
friadev:
```suggestion
The backup situation on Android is not nearly as robust unfortunately, but you can back up photos and files with a variety of services. If you use an online backup service we recommend choosing one which uses strong client-side encryption, so that the service provider is unable to access your data.
```

friadev (changes requested):

blog/posts/activists-guide-securing-your-smartphone.md
jonaharagon: We should **not** use E2EE to refer to situations where there is only one "end" (the client), we should only use it to refer to situations where there is **both** a sender and recipient.

We could say "we recommend choosing one which uses strong client-side encryption," but this has the same meaning as the current text, and client-side encryption is more jargon-y & less descriptive.

friadev: It's more descriptive since "keys you control" could mean you're messaging someone with asymmetric encryption like PGP. Many clients also don't give you direct access to the keys anyway so saying you control them doesn't make much sense. Client-side encryption gets across the important thing: that the local client running on your machine is doing the encryption *before* it's sent off.

freddy-m (approved):

friadev (approved):

dngray (approved):

PR #3203, opened 2026-03-10
update!: Remove ProtonVPN

jonaharagon:

- Remove ProtonVPN (https://discuss.privacyguides.net/t/remove-protonvpn/33980/312)


jordan-warne (approved):

hook9 (approved):
Looks good!
It doesn't meet our criteria, if we want to reconsider that criteria to recommend services for only specific platforms it should be done separately.

dngray (changes requested):
I think we should avoid doing this if, the reason is that [Network Extensions](https://developer.apple.com/documentation/networkextension) has an issue because it looks like Apple does not want you using pf in the future:

https://developer.apple.com/documentation/technotes/tn3165-packet-filter-is-not-api#Moving-off-packet-filter

meaning this will be an issue to all VPN providers as their apps will need to be updated to work on newer versions of mac platforms.

ph00lt0 (changes requested):
I prefer to hold this until we have done proper analysis of the root cause of this issue in the way proton uses apple's libraries where it seems that other client solutions use a legacy solution that might not be durable in the end either.

jordan-warne:
Whether there is an issue with NetworkExtensions on macOS and whether these apps are using a legacy method to create a firewall doesn't change that the ProtonVPN app on macOS leaks users IP addresses when switching servers despite the kill switch being enabled. I believe this should be separate issue where we have to discuss adjusting the criteria or adding a warning.

dngray:
While this specific issue with Network Extensions (the Apple-preferred method) mainly impacts Proton VPN, there are plenty of other ways a device can leak.

For example, we've seen cases affecting various VPN providers on macOS, [especially after system updates](https://mullvad.net/en/blog/2024/10/16/macos-sometimes-leaks-traffic-after-system-updates), and similar leaks have been found on other platforms too. A big one is how [DNS traffic can still leak outside](https://mullvad.net/en/blog/dns-traffic-can-leak-outside-the-vpn-tunnel-on-android) the tunnel on Android.

If you really care about preventing leaks, you shouldn't rely on the mediocre firewall rules these apps use. They usually fail because they have to allow unrestricted access for the VPN service itself while trying to block everything else on the device. For a real fix, you need a separate VPN router rather than relying on the device's own software.

ignoramous:
> A big one is how DNS traffic can still leak outside the tunnel on Android.

That one's just Mullvad ~~FUD~~ placing the blame elsewhere.

> Their post acknowledges that they fixed a bug in their app which resolved a major part of the issues. The issue while reconnecting looks a whole lot like a race condition and it's not yet clear if the issue is on the OS side or the app side. The other multicast issue looks like an OS bug, but this one needs further research to determine the cause. It's likely that the OS can prevent these leaks by working around how apps behave even if it's an app bug but that doesn't imply that the leak blocking toggle was meant to do that. It's meant to block access when the VPN is down, not if the VPN sets a partial configuration or does the setup in the wrong order.

https://github.com/GrapheneOS/os-issue-tracker/issues/3442#issuecomment-2093825046

friadev (changes requested):
I think we decided that we aren’t going to remove Proton over this:
https://discuss.privacyguides.net/t/remove-protonvpn/33980/359
So I think we should just close this PR

ignoramous:
> we decided

If this "we" includes the community, then no. Still open: https://discuss.privacyguides.net/t/how-should-we-handle-protons-misleading-marketing/36176/74

dngray:
>  So I think we should just close this PR

Done.

PR #3204, opened 2026-03-10
update!: Change VPN kill switch criteria

jonaharagon:
Kill switches are not available on some mobile platforms, particularly iOS, and have been removed from IVPN's clients on iOS as a result: https://www.ivpn.net/blog/removal-of-kill-switch-from-our-ios-app-due-to-apple-ip-leak-issue/

As we do not wish to hold providers responsible for platform limitations on platforms we do not recommend using in the first place, we should only hold this requirement as a firm rule for Linux and Android, but we should still emphasize we'd like to see them on all platforms when possible.

### DIFF
--- a/docs/vpn.md
+++ b/docs/vpn.md
@@ -298,19 +298,20 @@ It is important to note that using a VPN provider will not make you anonymous, b
 
 ### Technology
 
-We require all our recommended VPN providers to provide standard configuration files which can be used in a generic, open-source client. **If** a VPN provides their own custom client, we require a kill switch to block network data leaks when disconnected.
+We require our recommended providers to support modern technologies currently available to VPNs.
 
 **Minimum to Qualify:**
 
+- Must provide standard configuration files which can be used in a generic, open-source client such as the WireGuard apps.
 - Support for strong protocols such as WireGuard.
-- Kill switch built in to clients.
+- Functional kill switch built in to service-provided clients on our recommended [desktop](desktop.md) and [mobile](android/distributions.md) platforms. This kill switch should be able to block all internet traffic when the VPN connection drops unexpectedly.
 - Multi-hop support. Multi-hopping is important to keep data private in case of a single node compromise.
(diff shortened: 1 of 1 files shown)


dngray (approved):

jordan-warne (approved):

docs/vpn.md
```diff
@@ -298,19 +298,20 @@ It is important to note that using a VPN provider will not make you anonymous, b
 
 ### Technology
 
-We require all our recommended VPN providers to provide standard configuration files which can be used in a generic, open-source client. **If** a VPN provides their own custom client, we require a kill switch to block network data leaks when disconnected.
+We require our recommended providers to support modern technologies currently available to VPNs.
```
ph00lt0, line 301: this is extremely vague.

docs/vpn.md
```diff
@@ -298,19 +298,20 @@ It is important to note that using a VPN provider will not make you anonymous, b
...
 
 **Minimum to Qualify:**
 
+- Must provide standard configuration files which can be used in a generic, open-source client such as the WireGuard apps.
 - Support for strong protocols such as WireGuard.
-- Kill switch built in to clients.
+- Functional kill switch built in to service-provided clients on our recommended [desktop](desktop.md) and [mobile](android/distributions.md) platforms. This kill switch should be able to block all internet traffic when the VPN connection drops unexpectedly.
```
ph00lt0, line 307: will approve the change but let's come back to this one later.

ph00lt0 (approved):

docs/vpn.md
jonaharagon: It's intentionally vague because with these changes, this sentence is only serving as a lead up to the list below this sentence. That's why I added the first part of the sentence to the list, instead of stating it here.

hook9 (approved):
Do think they should provide kill switches on platforms even when a platform has system connections that won't go through VPNs, but looks fine to me.

docs/vpn.md
ph00lt0: alright then

PR #3212, opened 2026-03-17
update!: Add the Apple Health app to Menstrual Tracking apps

NylaTheWolf:

- Added Apple Health under Menstrual Tracking apps.
- Updated descriptions of Apple Fitness and Apple Health Records to fit with the new addition.

Apple Health Records and Apple Fitness are recommended apps in Health and Wellness. I think the Apple Health app in general is worth adding to the menstrual tracking list. I mentioned that it includes many other health and wellness features. As far as I can tell, it's gender-neutral. It's possible that I might've missed something in its information sections (information about menstrual cycles and the like) though.

Relevant discussion: https://discuss.privacyguides.net/t/add-apple-health-to-menstrual-trackers/36356

### DIFF
--- a/docs/health-and-wellness.md
+++ b/docs/health-and-wellness.md
@@ -62,6 +62,26 @@ Popular menstrual trackers like [Flo](https://techcrunch.com/2021/01/13/flo-gets
 
 </div>
 
+### Apple Health
+
+<div class="admonition recommendation" markdown>
+
+![Apple logo](assets/img/health-and-wellness/apple-health.webp#only-light){ align=right }![Apple logo](assets/img/health-and-wellness/apple-health-dark.webp#only-dark){ align=right }
+
+Apple Health is one of the default apps installed on iOS devices. It includes many health and wellness features (see [Health Records](#apple-health-records)), including menstrual cycle tracking. It also uses gender-neutral language. Apple Health always uses end-to-end encryption when syncing across multiple devices.
+
+[:octicons-home-16: Homepage](https://apple.com/health){ .md-button .md-button--primary }
+[:octicons-eye-16:](https://apple.com/legal/privacy/consumer-health-personal-data/en-ww){ .card-link title="Privacy Policy" }
+
+<details class="downloads" markdown>
+<summary>Downloads</summary>
+
+- [:simple-appstore: App Store](https://apps.apple.com/app/apple-health/id1242545199)
+
+</details>
+
+</div>
+
 ## Fitness Trackers
 
 These general purpose apps can do everything from counting steps and tracking sleep to measuring your heartbeat.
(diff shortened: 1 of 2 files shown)


docs/health-and-wellness.md
```diff
@@ -62,6 +62,28 @@ Popular menstrual trackers like [Flo](https://techcrunch.com/2021/01/13/flo-gets
 
 </div>
 
+### Apple Health
+
+<div class="admonition recommendation" markdown>
+
+![Apple logo](assets/img/health-and-wellness/apple-health.webp#only-light){ align=right }![Apple logo](assets/img/health-and-wellness/apple-health-dark.webp#only-dark){ align=right }
+
+Apple Health is one of the default apps installed on iOS devices. It includes many health and wellness features (see [Health Records](#apple-health-records)), including menstrual cycle tracking. It also uses gender-neutral language. Apple Health always uses end-to-end encryption when syncing across multiple devices. Additionally, almost all measured data is processed on your device.
```
friadev:
```suggestion
Apple Health is one of the default apps installed on iOS devices. It includes many health and wellness features (see [Health Records](#apple-health-records)), including menstrual cycle tracking. It also uses gender-neutral language. Apple Health always uses end-to-end encryption when syncing across multiple devices.
```
Not sure why this says “almost” all data, I’m pretty sure it’s all E2EE? You just put in the data manually so I’m not even sure if “measured” is the right word here.

docs/health-and-wellness.md
```diff
@@ -72,7 +94,7 @@ These general purpose apps can do everything from counting steps and tracking sl
 
 ![Apple Fitness logo](assets/img/health-and-wellness/apple-fitness.webp){ align=right }
 
-**Apple Fitness** is the default fitness app for iOS. Apple Fitness always uses end-to-end encryption when syncing across multiple devices. Additionally, almost all measured data is processed on your device.
+**Apple Fitness** is the default fitness app for iOS. It shares the security and privacy features of [Apple Health](#apple-health).
```
friadev:
```suggestion
**Apple Fitness** is the default fitness app for iOS. Apple Fitness always uses end-to-end encryption when syncing across multiple devices. Additionally, almost all measured data is processed on your device.
```
I think we should just keep this as it was. It’s actually measuring data with sensors so it’s a bit different than how Apple Health works.

friadev (changes requested):

docs/health-and-wellness.md
NylaTheWolf: This is talking about security and privacy policies/features specifically.

friadev: It's a totally different app though, I think just leave it how it was. If it really needs to be changed we can do a separate PR.

docs/health-and-wellness.md
NylaTheWolf: I was repeating what was there before with Apple Fitness, haha.

docs/health-and-wellness.md
NylaTheWolf: Fair enough!

friadev (approved):

dngray (approved):

dngray:
I assume we looked for SVG images where they existed?

Other than that looks good. Good criteria.

dngray:
Merged in https://github.com/privacyguides/privacyguides.org/commit/945cf8c254aea2437213226c6a6b406884f9b5ae not sure why this one didn't auto close when merging, normally that happens.

PR #3228, opened 2026-05-14
update: Update Mobile Brave Browser Settings

friadev:

- Remove "Clear data on exit" since it does the same thing as setting [Auto Shred](https://brave.com/privacy-updates/37-shred-button-android/) to clear on app close
- Remove social media blocking settings since they might break sites, and blocking specific sites being embedded doesn't stop other sites from being embedded
- Fix inconsistent tabs by removing these two settings (Android settings showing up for both iOS and Android)
- Fix mislabeled V8 setting
- FIx Leo AI setting name and add iOS version of the setting

relevant discussion: https://discuss.privacyguides.net/t/brave-mobile-settings-page-doesnt-reflect-new-android-update/37433


ph00lt0 (approved):
LGTM

docs/mobile-browsers.md
```diff
@@ -150,24 +150,16 @@ Shields' options can be downgraded on a per-site basis as needed, but by default
 
     1. This option disables JavaScript, which will break a lot of sites. To unbreak them, you can set exceptions on a per-site basis by tapping on the Shield icon in the address bar and unchecking this setting under *Advanced controls*.
 
-##### Clear browsing data (Android only)
-
-- [x] Select **Clear data on exit**
```
jonaharagon, line 143: Isn't auto shred a per-site setting instead of this whole-app setting though?

docs/mobile-browsers.md
```diff
@@ -89,11 +89,11 @@ Tor Browser is the only way to truly browse the internet anonymously. When you u
 
 === "Android"
 
-    These options can be found in :material-menu: → **Settings** → **Brave Shields & privacy**.
+    These options can be found in :fontawesome-solid-ellipsis-v: → **Settings** → **Brave Shields & privacy**.
```
jonaharagon:
```suggestion
    These options can be found in :material-dots-vertical: → **Settings** → **Brave Shields & privacy**.
```

docs/mobile-browsers.md
```diff
@@ -182,23 +174,33 @@ Shields' options can be downgraded on a per-site basis as needed, but by default
...
 
-1. This option is not present in Brave's iOS app.
+    - [ ] Uncheck **Show In Quick Search Engine Bar**
 
 #### Search engines
 
-These options can be found in :material-menu:/:fontawesome-solid-ellipsis: → **Settings** → **Search engines**.
+=== "Android"
+
+    These options can be found in :fontawesome-solid-ellipsis-v: → **Settings** → **Search engines**.
```
jonaharagon:
```suggestion
    These options can be found in :material-dots-vertical: → **Settings** → **Search engines**.
```

docs/mobile-browsers.md
```diff
@@ -182,23 +174,33 @@ Shields' options can be downgraded on a per-site basis as needed, but by default
...
+
+    - [ ] Uncheck **Show search suggestions**
+
+=== "iOS"
+
+    These options can be found in :fontawesome-solid-ellipsis: → **All Settings** → **Search engines**.
```
jonaharagon:
```suggestion
    These options can be found in :material-dots-vertical: → **All Settings** → **Search engines**.
```

docs/mobile-browsers.md
```diff
@@ -150,24 +150,16 @@ Shields' options can be downgraded on a per-site basis as needed, but by default
...
     1. This option disables JavaScript, which will break a lot of sites. To unbreak them, you can set exceptions on a per-site basis by tapping on the Shield icon in the address bar and unchecking this setting under *Advanced controls*.
 
-##### Clear browsing data (Android only)
-
-- [x] Select **Clear data on exit**
-
-##### Social Media Blocking (Android only)
-
-- [ ] Uncheck all social media components
-
```
jonaharagon, line 148: Should we note limitations instead of removing this setting?

jonaharagon (changes requested):

docs/mobile-browsers.md
friadev: This setting is the global default for Brave Shields and then you can override it for individual sites if you want to. It's the same on iOS.

docs/mobile-browsers.md
friadev: I use the font awesome ones to be in line with the iOS menu, we should change both to the material dots ones to be consistent.

docs/mobile-browsers.md
friadev: It kinda looks bad to have this one setting that doesn't change with the tabs and all it's going to do is break websites. For example I embed tweets all the time in my articles. I don't really think it provides any real privacy benefit.

If someone really wants to block twitter they should enable the block social media setting in shields.

docs/mobile-browsers.md
ph00lt0: You are kinda both right. It is a setting per site. But we should like @friadev suggested recommend to change the default. It essentially then possible to whitelist websites you want to remind you (perhaps that requires some additional guidance). Setting a default to forget me/shred is a good idea.

docs/mobile-browsers.md
ph00lt0: I mean i guess it could limit twitter from seeing what websites you visit. Those embeds are tracking pixels in the end.

friadev: I’m just saying I think a proper method would be blocking social media in shields and not just specifically embeds of like two websites for some reason.

jonaharagon (approved):

PR #3231, opened 2026-05-18
feat!: Hugo website redesign

jonaharagon:
https://privacyguides-preview.web.garage.jonaharagon.net/3231/en/


friadev (changes requested):
Image

I think we should put the logos inside the title cards the same way we have it on the old website:

Image

friadev (changes requested):
Image

It's confusing when you click a section like Hardware you get taken to the first entry, it should just expand to show the entries without taking you to a different page.

jonaharagon:
@friadev https://www.privacyguides.org/en/meta/pr-comments/#commenting lol

friadev (approved):

ph00lt0 (changes requested):
The text is relatively small. Hard to read for some i suppose. It at least makes it too difficult to navigate the hamburger menu. 

For the hamburger menu i think the buttons to expand the sections should also be having more padding to avoid missclicks. 

Still missing translations until decision to remove them.

ph00lt0:
Search seems broken. 

Screenshot_20260521-151722

jonaharagon:
Search only appears broken in this preview (probably because it is running out of a subdirectory)

> The text is relatively small. Hard to read for some i suppose.

Only in the hamburger menu?

--- 

Also https://github.com/privacyguides/privacyguides.org/pull/3231#issuecomment-4491346364

ph00lt0:
> Search only appears broken in this preview (probably because it is running out of a subdirectory)
> 
> > The text is relatively small. Hard to read for some i suppose.
> 
> Only in the hamburger menu?
> 
> Also [#3231 (comment)](https://github.com/privacyguides/privacyguides.org/pull/3231#issuecomment-4491346364)

Ah okay. Hope it will work on previews in future.

Well all texts are quite small. I think for the me u that is even more problematic as it is also hard to click the right thing.

PR #3233, opened 2026-05-19
update: Add source for claim about Mullvad extension fingerprinting

friadev:

- Added a link to Mullvad's [site](https://mullvad.net/en/help/tag/mullvad-browser#112) for the claim that removing the Mullvad extension from Mullvad Browser won't affect the fingerprint


jonaharagon (changes requested):
Mullvad's site is less informative than ours. There is a forum post somewhere from rui which both has more detail and can be used as an official source.

dngray (approved):

jonaharagon (approved):
I actually can't find this forum post with a search unfortunately. Weird, maybe I am misremembering 🤔 

Without it, I'm fine with this then.

PR #3235, opened 2026-05-24
update: Remove mention of removed Firefox PPA setting

tastenbier:
The linked article states that privacy-preserving attribution is no longer included in Firefox:

> Privacy-preserving attribution (PPA) was an experimental feature in Firefox version 128 which was never activated and was later removed.

https://support.mozilla.org/en-US/kb/privacy-preserving-attribution

### DIFF
--- a/docs/desktop-browsers.md
+++ b/docs/desktop-browsers.md
@@ -197,12 +197,6 @@ Additionally, the Mozilla Accounts service collects [some technical data](https:
 1. Open your [profile settings on accounts.firefox.com](https://accounts.firefox.com/settings#data-collection)
 2. Uncheck **Data Collection and Use** > **Help improve Firefox Accounts**
 
-##### Website Advertising Preferences
-
-- [ ] Uncheck **Allow websites to perform privacy-preserving ad measurement**
-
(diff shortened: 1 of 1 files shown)


friadev (approved):

jordan-warne (approved):
LGTM

friadev:
@tastenbier could you go ahead and sign your commit? I can’t merge it until it’s signed:
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

tastenbier:
Ah, thanks for the heads up. I went and set up a GPG key. Unfortunately GitHub seems to be bugged.

I signed my commit and can verify it locally just fine:

```
$ git verify-commit HEAD

[...]
gpg: Good signature [...]
```

I added the gpg public key to my GitHub account, but GitHub still keeps saying my commit is 'Unverified'.

I know that the key is correct, because if I add an email to my commit, Github all of a sudden says my commit is verified, even-though the commit mail has nothing to do with the gpg signature itself, so adding one should change nothing - yet it does. See my last two commits on this test branch for example: https://github.com/tastenbier/privacyguides.org/commits/gpg-test/. Both commits are signed with the same key, but only one shows up as verified.

Seems like I can't move this forward without being forced to add an email to my commit, which I would prefer not to do.

So what now? Do you want to amend my commit and sign it yourself?

friadev:
>Seems like I can't move this forward without being forced to add an email to my commit, which I would prefer not to do.
>
>So what now? Do you want to amend my commit and sign it yourself?

GitHub should provide you with a `noreply` [email address](https://docs.github.com/en/account-and-profile/reference/email-addresses-reference#your-noreply-email-address) you can use if you set your preferences to **Keep my email address private** in your email preferences.

tastenbier:
Yep, I know, that's what I tested on my [test branch](https://github.com/tastenbier/privacyguides.org/commits/gpg-test/). But I prefer to not add any address to my commits - not even the generic noreply one.

I signed my commit and uploaded the public key to GitHub, yet they don't verify the commit and seem to insist on me adding a mail address. So I don't see any other way to move this forward, other than you signing the commit yourself or closing this PR (or is there?). Choose which ever option you prefer or let me know if there is another way.

PR #3236, opened 2026-05-24
update: Cleanup

tastenbier:
Made some small cleanups to the markdown.

I found these inconsistencies via [`rumdl`](https://github.com/rvben/rumdl), so some of them might not have shown up previously via markdownlint (but some did).

---

On that note:
Currently, you seem to only use markdownlint as an optional helper for editors that have markdownlint installed. Would you see any value in making markdownlint part of your pipeline?

Obviously, you couldn't just enable it right away. With the current configuration, markdownlint reports dozens of unfixed 'errors', so you would drown in reports. Some errors might also be false-positives, because many checks depend on preference.

But with some careful tweaking and fixing, you could build a harness that will pass all checks and will then prevent future errors (e.g. dead links) from creeping in.

If you are open to this, I would be happy to help. Just wanted to check with you first.


friadev (approved):

jonaharagon (approved):

jonaharagon:
> Would you see any value in making markdownlint part of your pipeline?

Yes, but not until #3231 is merged.

PR #3239, opened 2026-05-28
update!: Remove Shelter

friadev:

- Remove Shelter, since Private Space is a standard feature in Android now. Shelter hasn't been updated in years as well.

relevant discussion: https://discuss.privacyguides.net/t/remove-shelter/21658


Refocus537:
I firmly disagree. Private space is a completely different feature and should not be compared with the use-case enabled by Shelter.

From https://support.google.com/android/answer/15341885:
>When private space is locked:
>The apps in the private space are:
>- Completely stopped. These apps can’t perform foreground or background activities, like showing notifications
>    - When the device is locked, apps can't access sensor data or perform any functions. For example, medical apps can’t track health information when the device is locked. **You should not use private space for apps that need to provide you with functionality in the background when your private space is locked.**

However you can configure private space to lock automatically at every reboot and unlock manually every time after a reboot. That practically means private space will stay unlocked during times when background functionality is needed.

This appears to be a workaround rather than the intended use of the feature.

jordan-warne:
There are two alternatives that I believe fulfil the use case of Shelter, user profiles and private spaces.

User profiles can run in the background and offer better separation between data than a work profile that shelter provides. If the user doesn't require background usage of their apps they can use a private space in their main profile. Shelter was useful during a time where user profiles and private spaces didn't exist however it's usefulness has decreased significantly since.

Shelter also hasn't been updated since 2023, and hasn't had a commit since [2024](https://github.com/PeterCxy/Shelter/commits/master/).

We recommend that users utilise the latest android version for maximum security and privacy therefore Shelter should be removed as it only really provides value for users running older unsupported Android versions.

jordan-warne:
Mentions of CalyxOS should be removed as it's not recommended by Privacy Guides any more.

blog/posts/grapheneos-or-calyxos.md
```diff
@@ -67,7 +67,7 @@ On the other hand, GrapheneOS officially recommends [Sandboxed Google Play](http
...
 
-CalyxOS doesn't feature any improvements to user profiles over AOSP, and instead includes a device controller app so that the [work profile](https://www.privacyguides.org/android/overview#work-profile) can be used without needing to download a third party app such as [Shelter](https://www.privacyguides.org/android/#shelter). However, work profiles are not nearly as flexible (as you're limited to only one) and don't provide the same amount of isolation and security.
+CalyxOS doesn't feature any improvements to user profiles over AOSP, and instead includes a device controller app so that the [work profile](https://www.privacyguides.org/android/overview#work-profile) can be used without needing to download a third party app. However, work profiles are not nearly as flexible (as you're limited to only one) and don't provide the same amount of isolation and security.
```
jordan-warne: This should be removed.

friadev:
> Mentions of CalyxOS should be removed as it's not recommended by Privacy Guides any more.

Since that was a blog post, I removed the link so it didn't break but I didn't want to edit it too much. Maybe I could just remove the link and add the mention of shelter back even.

Refocus537:
Just recently another developer released a modernized Shelter fork. Available on [F-Droid](https://f-droid.org/packages/io.github.kennethchoinfosec.haven).

https://github.com/Kenneth-Cho-InfoSec/Haven
>Haven is a Free and Open-Source Android Work Profile isolation app. It is a modernized fork in the Shelter / Island lineage, **focused on maintaining the practical app-isolation workflow while bringing the codebase up to date with current Android security expectations, modern build tooling, and a Material Design 3 Expressive UI.**

jonaharagon:
Can't personally approve this change as IMO it is _completely_ against the community consensus at https://discuss.privacyguides.net/t/remove-shelter/21658

1. Work profiles can be scheduled

2. GrapheneOS [acknowledges](https://grapheneos.social/@GrapheneOS/114146765620652881) work profile use-cases:

   > The remaining use case of a work profile for now is that you can't have more than 1 Private Space so you can continue using a work profile as a 2nd poor man's Private Space.

   > We look forward to simply being able to tell people not to use work profiles for personal use but we can't yet because it's useful having 2 nested profiles rather than 1.

3. GrapheneOS would prefer to replace work profiles and secondary user profiles with multiple private spaces:

   > In the future, support for having multiple Private Spaces would largely replace 1 person using multiple secondary users.

4. Work profiles [use the same framework](https://grapheneos.social/@GrapheneOS/114146802284865723) [[2](https://grapheneos.social/@GrapheneOS/114146831065702881)] as private spaces and secondary profiles, just with customizable isolation, which is a feature for many, not an antifeature.

5. I am citing GrapheneOS frequently, but it is important to remember that GrapheneOS users represent a minority of our readers, and our Android recommendations are general purpose.

Until private spaces have a _reasonable_ quantity limit we should not remove work profile tools. It is important to continue recommending Shelter unless a better alternative which achieves the same functionality (secondary user profiles do not) is found, because it serves a purpose many people demand, and Island and Insular are worse.

hakavlad:
I use Android 11. My wife uses Android 13. Don't remove Shelter.

friadev:
> Can't personally approve this change as IMO it is _completely_ against the community consensus at https://discuss.privacyguides.net/t/remove-shelter/21658
> 
> 1. Work profiles can be scheduled
> 2. GrapheneOS [acknowledges](https://grapheneos.social/@GrapheneOS/114146765620652881) work profile use-cases:
>    > The remaining use case of a work profile for now is that you can't have more than 1 Private Space so you can continue using a work profile as a 2nd poor man's Private Space.
>    
>    
>    > We look forward to simply being able to tell people not to use work profiles for personal use but we can't yet because it's useful having 2 nested profiles rather than 1.
> 3. GrapheneOS would prefer to replace work profiles and secondary user profiles with multiple private spaces:
>    > In the future, support for having multiple Private Spaces would largely replace 1 person using multiple secondary users.
> 4. Work profiles [use the same framework](https://grapheneos.social/@GrapheneOS/114146802284865723) [[2](https://grapheneos.social/@GrapheneOS/114146831065702881)] as private spaces and secondary profiles, just with customizable isolation, which is a feature for many, not an antifeature.
> 5. I am citing GrapheneOS frequently, but it is important to remember that GrapheneOS users represent a minority of our readers, and our Android recommendations are general purpose.
> 
> Until private spaces have a _reasonable_ quantity limit we should not remove work profile tools. It is important to continue recommending Shelter unless a better alternative which achieves the same functionality (secondary user profiles do not) is found, because it serves a purpose many people demand, and Island and Insular are worse.

Well then we need to find a work profile app that's still supported. Maybe we could look at Haven like is suggested higher up. Something that's at least keeping up with the newer target SDK versions.

>Existing apps must target Android 14 (API level 34) or higher to remain available to new users on devices running Android OS higher than your app's target API level. Apps that target Android 13 (API level 33) or lower, including Android 12 (API level 31) or lower for Wear OS and Android TV, will only be available on devices running Android OS that are the same or lower than your app's target API level.

https://developer.android.com/google/play/requirements/target-sdk

jonaharagon:
I am hiding mine and @friadev's comments above because discussions should actually be continued on the forum, where I've copied my comment to: https://discuss.privacyguides.net/t/remove-shelter/21658/35

For the record, Fria's only response in the collapsed comment was that it is not maintained. I think Fria and @jordan-warne here are operating under an incorrect assumption that the GitHub repo linked above is the current source code, when it is not. F-Droid pulls from the correct repository. I don't know why the GitHub repo is frozen, but whatever.

Anyways, besides that we can talk more on the forum, thanks.

jonaharagon:
> https://github.com/Kenneth-Cho-InfoSec/Haven

@Refocus537 **strong oppose** to recommending Haven, for reasons I'll explain on the forum sometime this week haha

PR #3241, opened 2026-05-29
update: Remove mention of Firefox PPA setting

friadev:
The linked article states that privacy-preserving attribution is no longer included in Firefox:

>Privacy-preserving attribution (PPA) was an experimental feature in Firefox version 128 which was never activated and was later removed.

https://support.mozilla.org/en-US/kb/privacy-preserving-attribution


jonaharagon:
Duplicate of #3235

PR #3243, opened 2026-06-01
feat: Add rumdl config

dngray:
This config I believe mostly does what our old markdownlint config did.


jonaharagon (changes requested):
lgtm. just want to also make sure we add this to vscode recommended extensions and devcontainer extensions so it works with github.dev & codepaces.

PR #3246, opened 2026-06-06
update: add "Anonymous sign-in" authentication method

amrabumady:
This PR adds a new subsection under `## Authentication methods` describing the
anonymous sign-in pattern — where the client locally derives a site-specific
pseudonymous identifier (typically via HKDF over a user-controlled seed +
origin), so the service never receives PII and cross-site correlation is
structurally impossible.

The pattern is currently underrepresented in the Privacy Guides documentation
even though it has direct privacy-relevance: it is the only authentication
method in the page that does not require trusting the service with any
recoverable identifier.

Structure mirrors the existing subsections (Email and password, OAuth, Phone,
Username and password): one-sentence intro, "main advantages" / "but there are
disadvantages" lists, a closing paragraph on appropriate use, and one reference
to an open-source implementation for readers who want to dig further.

I've deliberately kept the implementation reference to a single open-source
project rather than promoting any hosted service. The substantive value of the
section is the pattern explanation; if maintainers feel the GitHub link is too
much, dropping it leaves the section intact.

Happy to revise wording to match house style — I tried to follow the tone of
the surrounding subsections but I'm new to the repo's conventions.

### DIFF
--- a/docs/basics/account-creation.md
+++ b/docs/basics/account-creation.md
@@ -83,3 +83,25 @@ In many cases you will need to provide a number that you can receive SMS or call
 ### Username and password
 
 Some services allow you to register without using an email address and only require you to set a username and password. These services may provide increased anonymity when combined with a VPN or Tor. Keep in mind that for these accounts there will most likely be **no way to recover your account** in the event you forget your username or password.
+
+### Anonymous sign-in
+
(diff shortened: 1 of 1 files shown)


I-I-IT:
It is pseudonymous, not anonymous.

dngray:
This seems like a developer library, so not really something a user can use, to enhance privacy. If it had some examples of large notable services using it then that might be something worth talking about, but I don't see that.

Also, please do not send AI slop to privacy guides, it won't be added.

PR #3256, opened 2026-07-21
update: Describe benefits of ControlD's "redirect" feature

twizzlebizzle:

- Updated controlD section of the dns provider site to mention the redirect feature.  
I've moved to this personally with all logging disabled.  It's only a feature on a paid plan.  
It allows an automatic proxy of your ip address that's transparent to the website.  eg.  I route all of Reddit and imgur services through a redirect to canada.  This means I do not get the UK based restriction (NSFW).  
  
 You might say well just use a VPN?  I personally see this as 2 seperate systems.  
  
The VPN is well covered so I won't discuss that.  Where the controld system is different is that it can operate without a VPN.  So you can apply it network wide without actually requiring a VPN on any device it silently intercepts and proxys your HTTP traffic through one of their servers in another country.  So reddit sees it as oh somebody from Canada is connecting.  
  
This may or may not have privacy benefits, however I see the benefit of not being pushed/coerced into needing to age verify and it works silently without accidentally turning it off, can easily apply to kids accounts, parents accounts, old peoples accounts at the network level.

If you are making changes that you have a conflict of interest with, you MUST
disclose this as well (this does not disqualify your PR by any means): NA apart from I use the service and find it a great benefit.


docs/dns.md
```diff
@@ -77,6 +82,8 @@ In addition to their paid plans, they offer a number of preconfigured DNS resolv
 
 </div>
```
dngray, line 84: > feature that routes DNS queries through transparent proxy servers across 69+ countries. This lets you mask your source IP and appear as if you're in a different region without using a VPN

I have a real problem with the way that is worded, because it makes it sound like it's a privacy feature and substitute for a VPN - which it is not. Especially when you consider text on their site:

> Control D's Traffic Redirection allows you to route DNS queries through predefined proxy locations across various countries, enabling you to mask your IP address and appear as if you're in a different region. This is ideal for organizations needing to manage their regional presence or comply with location-based requirements.

The purpose of this feature is not to provide anonymity. They are talking about things like "Is this being stored in Europe? Yes, we need to comply with GDPR". Not age verification.

> A practical privacy benefit: you can bypass age verification requirements on websites by redirecting traffic through a country where such checks aren't mandated,

This does not work the way you think it does as that kind of compliance checking happens at an application level, not network layer.

For many years I've sent my DNS queries from LAN > Unbound > VPN, regardless of whether or not I am connecting directly to a website or not via PBR rules on my router. The website still knows what country I am in when I connect to it.

You're still connecting to the site so it still knows your IP address, regardless of where the DNS lookup resolves to. Compliance checking works on that basis not which DC the user hits.

PR #3265, opened 2026-08-30
update: Recommend against Omarchy

### LINKED ISSUE #3213: Add Omarchy to the list of Arch-based distributions recommended against
### Affected page

https://www.privacyguides.org/en/os/linux-overview/#arch-based-distributions

### Description

As there is now a hot new Arch-based distro around, I feel like it is appropriate to consider updating the recommendations for Arch-based distros based on the existing recommendations against.

The rationale for proposing this as a correction is simple: Omarchy combines both outdated packages (for twice as long as Manjaro!) [1] [2] [3], and use of blindly built AUR packages [4]. It used to rely on Chaotic AUR, but migrated to its own package repository only building select packages automatically (every 6 hours, automated) [5].

On top of the distribution having both properties that lead to the recommendation against of Manjaro and Garuda, Omarchy being geared towards "general usability", its defaults include privacy endangering services as first-class citizens such as Google services (Maps, Photos...) [6] [7] as well as calling Google servers in distro-provided helper scripts [8].

Even worse also ships problematic configuration ootb. While most problems aren't directly impacting privacy and security, it makes signature optional and enables `TrustAll` for its pacman package repository [9], which the ArchWiki clearly states numerous times should not be done [10] [11] [12] (not that any privacy and/or security conscious person would need the heads up).

For all the reasons above it is pretty clear that Omarchy is not geared towards privacy, and its vision of usability and convenience goes fundamentally against the interests of privacy, warranting a mention as a distribution to avoid if privacy is a concern.

Addressing the elephant in the room: Omarchy is not a distribution free of controversy, for various reasons. Whether it is a good or bad distribution is not the topic; all the above arguments are factually verifiable and highlight design choices that are not compatible with privacy to an extent that justifies its mention as a heads up for people who are sensitive to their privacy.

### Sources

[1]\: https://github.com/basecamp/omarchy/discussions/4628#discussioncomment-15880346
[2]\: https://github.com/basecamp/omarchy/issues/4577#issuecomment-3883970957 (outdated packages "not a bug")
[3]\: https://learn.omacom.io/2/the-omarchy-manual/68/updates#:~:text=running%20one%20month%20behind%20the%20latest
[4]\: https://github.com/omacom-io/omarchy-pkgs
[5]\: https://github.com/basecamp/omarchy/pull/1348
[6]\: https://github.com/basecamp/omarchy/blob/b4cb5e287bd6b35add754e9f35ba2f92514d62a4/install/packaging/webapps.sh#L3-L15
[7]\: https://github.com/basecamp/omarchy/blob/b4cb5e287bd6b35add754e9f35ba2f92514d62a4/config/hypr/bindings.conf#L18-L27
[8]\: https://github.com/basecamp/omarchy/blob/b4cb5e287bd6b35add754e9f35ba2f92514d62a4/bin/omarchy-webapp-install#L16
[9]\: https://github.com/basecamp/omarchy/blob/b4cb5e287bd6b35add754e9f35ba2f92514d62a4/default/pacman/pacman-stable.conf#L29
[10]\: https://wiki.archlinux.org/title/Pacman/Package_signing#Disabling_signature_checking
[11]\: https://wiki.archlinux.org/title/Pacman/Package_signing#Configuring_pacman:~:text=The%20SigLevel,for%20debugging%20purposes
[12]\: https://wiki.archlinux.org/title/Pacman/Package_signing#Disabling_signature_checking:~:text=of%20trust%2C%20and-,should%20not%20be%20used,-%3A

### Before submitting

dngray:

- Closes https://github.com/privacyguides/privacyguides.org/issues/3213

### DIFF
--- a/docs/os/linux-overview.md
+++ b/docs/os/linux-overview.md
@@ -77,6 +77,7 @@ Additionally, we recommend **against** these two Arch derivatives specifically:
 
 - **Manjaro**: This distribution holds packages back for 2 weeks to make sure that their own changes don’t break, not to make sure that upstream is stable. When AUR packages are used, they are often built against the latest [libraries](https://en.wikipedia.org/wiki/Library_(computing)) from Arch’s repositories.
 - **Garuda**: They use [Chaotic-AUR](https://aur.chaotic.cx) which automatically and blindly compiles packages from the AUR. There is no verification process to make sure that the AUR packages don’t suffer from supply chain attacks.
+- **Omarchy**: This distribution combines out-of-date packages and blindly uses the AUR, which we consider extremely risky due to supply chain attacks. It enables `TrustAll` for the pacman repository, which is strongly advised against by Arch Linux. Most recently, it was found to be [exposing the Docker socket](https://0xcc.io/posts/omarchy-root-creds/), thereby allowing any program to escalate privileges to root without a password.
 
 ### Linux-libre kernel and “Libre” distributions
 


friadev (approved):

docs/os/linux-overview.md
ph00lt0, line 76: Seems there are more now

ph00lt0:
How does this meet our no anti recommendations?

ph00lt0 (changes requested):

docs/os/linux-overview.md
jordan-warne: We already recommend against Garuda and Manjaro, what's the issue with adding another distribution?

friadev: Yeah I'm generally against having anti-recommendations, because we could honestly fill the site up with them. But since we already do for a few it seemed ok. We could just remove those entirely though if you think that would be better.

ph00lt0: > We already recommend against Garuda and Manjaro, what's the issue with adding another distribution?

My comment here was more about the word "two" when we then recommend against "three".

ph00lt0: > Yeah I'm generally against having anti-recommendations, because we could honestly fill the site up with them. But since we already do for a few it seemed ok. We could just remove those entirely though if you think that would be better.

In favour of removing it entirely. We have multiple times had this discussion in the team which always ended in we should not do anti recommendations as the list could be endless.

friadev: > My comment here was more about the word "two" when we then recommend against "three".

Yeah this is kinda my argument against anti-recommendations, when there's already two you can't argue against another one and another one and there's always an infinite supply of terrible linux distros.

jordan-warne: It looks like this might be legacy stuff from PTIO anyway, so perhaps we just remove all of them to align with our no anti-recommendation rule. Generally this feels like information that should be accessible to people on the forum, no the site itself.

PR #3269, opened 2026-08-31
update: Remove Specific Arch-Based Anti-Recommendations

### LINKED ISSUE #3213: Add Omarchy to the list of Arch-based distributions recommended against

friadev:

- Removed specific anti-recommendations from the Linux Overview page, as there are an almost infinite number of bad Linux distros and we don't want to clutter the page with them

closes #3213

### DIFF
--- a/docs/os/linux-overview.md
+++ b/docs/os/linux-overview.md
@@ -67,16 +67,13 @@ Arch and Arch-based distributions are not recommended for those new to Linux (re
 
 For a secure system, you are also expected to have sufficient Linux knowledge to properly set up security for their system such as adopting a [mandatory access control](#mandatory-access-control) system, setting up [kernel module](https://en.wikipedia.org/wiki/Loadable_kernel_module#Security) blacklists, hardening boot parameters, manipulating [sysctl](https://en.wikipedia.org/wiki/Sysctl) parameters, and knowing what components they need such as [Polkit](https://en.wikipedia.org/wiki/Polkit).
 
-Anyone using the [Arch User Repository (AUR)](https://wiki.archlinux.org/title/Arch_User_Repository) **must** be comfortable auditing PKGBUILDs that they download from that service. AUR packages are community-produced content and are not vetted in any way, and therefore are vulnerable to software [:material-package-variant-closed-remove: Supply Chain Attacks](../basics/common-threats.md#attacks-against-certain-organizations){ .pg-viridian }, which has in fact happened [in the past](https://bleepingcomputer.com/news/security/malware-found-in-arch-linux-aur-package-repository).
+Arch-based derivatives designed to make Arch "easier to use" are not recommended, as they frequently encourage poor security practices while weakening the base system. These practices include bundling outdated packages in downstream repositories that are updated less frequently than Arch's official channels. Furthermore, these distributions usually fail to properly educate users on the risks associated with the Arch User Repository (AUR).
+
+Anyone using the [Arch User Repository (AUR)](https://wiki.archlinux.org/title/Arch_User_Repository) **must** be comfortable auditing PKGBUILDs (the recipes that build packages) that they download from that service. AUR packages are community-produced content and are not vetted in any way, and therefore are vulnerable to software [:material-package-variant-closed-remove: Supply Chain Attacks](../basics/common-threats.md#attacks-against-certain-organizations){ .pg-viridian }, which has in fact happened [in the past](https://bleepingcomputer.com/news/security/malware-found-in-arch-linux-aur-package-repository) and more recently when [1500 packages](https://discuss.privacyguides.net/t/around-1-500-aur-packages-compromised-with-rootkit-like-malware/38517) were compromised.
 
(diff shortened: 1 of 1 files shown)


jordan-warne (approved):
LGTM based on the discussion in: https://github.com/privacyguides/privacyguides.org/pull/3265

dngray:
Maybe this is a better approach, we should make sure that we say that the arch recommendation only applies to actual arch from archlinux,org not the other crappy derivatives out there with poorer security practices, supply chain risks (such as blindly using AUR), and ignoring all the other good archlinux practices.

jordan-warne (approved):
LGTM

cyyynthia:
b7d288a6cb84e86a9e1d5642fdf02c5a01fdb648 feels like it partially re-introduces the anti-recommendation wording; I think the addition to the last paragraph is fine as it clarifies the scope of the recommendation, whereas the added paragraph is an anti-recommendation (just a broader one than cherry-picking "bad derivatives").

It's also at odds with the section being "Arch-based distributions", and the mention of "Arch and Arch-based distributions" as first words; if the recommendation is moving towards "we recommend vanilla Arch Linux for advanced users, but you're on your own for other distros based on it" then the mention of Arch derivatives should be dropped altogether. That would be in line with the recommendations section, that simply mentions "Arch Linux".

Manjaro's practice of lagging behind doesn't seem to be explicitly addressed in the KB outside the mention in the Arch section which is being gutted, and I think is important to address. The "Release Cycle" section has a paragraph about "frozen distributions such as Debian", perhaps a new paragraph about "derivative distributions lagging behind its upstream such as Manjaro" would be appropriate?

dngray:
I think that later change better defines both scenarios of outdated lagging packages, and secondary repositories with packages kept back.

> anti-recommendation (just a broader one than cherry-picking "bad derivatives").

The main reason for that is that I still want to inform users what bad things to look out for in arch distributions which are not actually arch. The reason is because these options are specifically sold to users as "arch but easier" when in fact it's not actually arch anymore. That's why Arch people get particularly annoyed when people ask about those projects in their support channels.

The reason for not listing specific bad products, is twofold, partially it advertises them, and partially we don't want to keep a running list of all the bad projects out there. Imagine if we did that for something like the Email or VPN page :laughing: Also those products are not typically advertised to people as being "XYZ service", when it is in fact not.

jordan-warne (approved):
This seems like a better way to do this instead of having anti-recommendations.

dngray (approved):

